PR #8, opened 2014-07-21
Remove instance position

antonjefcoate:
@payman81 Careful on this, there's a dependency on a related PR on the JustEat internal library.


payman81:
@antonjefcoate This seems to be way out of date. Do you want to close it and make a fresh PR?

payman81:
This is way out of date @antonjefcoate - We need to have another fresh attempt and remove instance position from the codebase.

PR #46, opened 2014-12-29
Use the subscribe queue method in the SDK rather than implementing your ...

jaimalchohan:
...own see: https://github.com/aws/aws-sdk-net/blob/master/AWSSDK_DotNet35/Amazon.SimpleNotificationService/AmazonSimpleNotificationServiceClient.Extensions.cs#L78

### DIFF
--- a/JustSaying.AwsTools/QueueCreation/AmazonQueueCreator.cs
+++ b/JustSaying.AwsTools/QueueCreation/AmazonQueueCreator.cs
@@ -24,7 +24,7 @@ namespace JustSaying.AwsTools.QueueCreation
         {
             var queue = EnsureQueueExists(region, queueConfig);
             var eventTopic = EnsureTopicExists(region, serialisationRegister, queueConfig);
-            EnsureQueueIsSubscribedToTopic(eventTopic, queue);
+            EnsureQueueIsSubscribedToTopic(region, eventTopic, queue);
 
             return queue;
         }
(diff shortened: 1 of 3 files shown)


jaimalchohan:
@payman81 the "hard limit of 20 statements" is a limit to the number of Policy Statements (the security policy applied to the topic), not to the number of subscribers.  And I've verified that a Topic created by JustSaying can have over 20 subscribers.

The long bit ....

JustSaying already creates a policy on the Topic like so 

```
{
    "Version" : "2012-10-17",
    "Statement" : [
        {
            "Effect" : "Allow",
            "Principal" : {
                "AWS" : "*"
            },
            "Action"    : [
                "sns:Publish",
                "sns:RemovePermission",
                "sns:SetTopicAttributes",
                "sns:DeleteTopic",
                "sns:ListSubscriptionsByTopic",
                "sns:GetTopicAttributes",
                "sns:Receive",
                "sns:AddPermission",
                "sns:Subscribe"
            ],
            "Condition" : {
                "StringEquals" : {
                    "aws:SourceOwner" : "507204202721"
                }
            }
        }
    ]
}
```

This allows all resources in the same account to subscribe to the topic and the policy is applied to the topic when JustSaying creates it.  At no other time does JustSaying add another policy on the topic so it's impossible to the the 20 policy statements limit.

payman81:
@jaimalchohan I wondered if it creates a policy per subscriber as we used to do. But with a policy like above there is no issue.

PR #48, opened 2015-01-09
Region failover support.

brainmurphy:


payman81:
PR has been validated in production. Major update will be reflected in package version.

PR #51, opened 2015-01-16
Get rid of the generic serialiser 

jaimalchohan:
Generic interfaces are an anti-pattern IMO, but anyway, the generic on the serialiser is one of the root causes of a lot of forced implementation in JustSaying. 

When you have a generic class, anything that consumes that class and cares about the type will generally be forced into being generic itself, making code quite rigid as things that don't care will end up passing around Types via generics just for the sake of it.

This change is a start on breaking that dependency by wrapping the Type of class itself, allow dependant classes to lookup by the Type object rather than having to be generic themselves. 

It's a small prices to pay for the amount of refactoring that will be possible in coming PRs

### DIFF
--- a/JustSaying.AwsTools.UnitTests/Sqs/WhenPublishing.cs
+++ b/JustSaying.AwsTools.UnitTests/Sqs/WhenPublishing.cs
@@ -25,7 +25,7 @@ namespace JustSaying.AwsTools.UnitTests.Sqs
         {
             // ToDo: We need to clean up serialisation (away from Json.Net)
             //var serialiser = Substitute.For<IMessageSerialiser<GenericMessage>>();
-            //_serialisationRegister.GetSerialiser(typeof(GenericMessage)).Returns(serialiser);
+            //_serialisationRegister.GeTypeSerialiser(typeof(GenericMessage)).Returns(serialiser);
             _sqs.ListQueues(Arg.Any<ListQueuesRequest>()).Returns(new ListQueuesResponse{QueueUrls = new List<string>{Url}});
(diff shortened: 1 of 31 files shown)


JustSaying.AwsTools.UnitTests/Sqs/WhenPublishing.cs
```diff
@@ -25,7 +25,7 @@ protected override void Given()
         {
             // ToDo: We need to clean up serialisation (away from Json.Net)
             //var serialiser = Substitute.For<IMessageSerialiser<GenericMessage>>();
-            //_serialisationRegister.GetSerialiser(typeof(GenericMessage)).Returns(serialiser);
```
payman81: Can you clean up the unnecessary comments?

JustSaying.AwsTools/SqsPublisher.cs
```diff
@@ -24,7 +24,7 @@ public void Publish(Message message)
             _client.SendMessage(new SendMessageRequest
             {
                 MessageBody = GetMessageInContext(message),
-                //MessageBody = _serialisationRegister.GetSerialiser(message.GetType()).Serialise(message),
+                //MessageBody = _serialisationRegister.GeTypeSerialiser(message.GetType()).Serialise(message),
```
payman81: More unnecessary comments.

payman81:
@jaimalchohan The PR looks good. I ran your branch and all works as expected. There were a few comments that were there from before and thought you might want to remove them before I merge.

jaimalchohan:
@payman81 Done.

PR #52, opened 2015-01-20
Separate the AWS SNS Publisher implementation from IMessagePublisher

jaimalchohan:
- Squash SnsTopicByName and SnsTopicBase into 1 class and rename it to SnsPublisher
- Remove the IMessagePublisher from SnsPublisher and SqsPublisher and add a IPublisher interface
- Move IMessageSerializationFactory junk away from SnsPublisher


antonjefcoate:
Hey @jaimalchohan There's a bunch of stuff in here which doesn't quite speak for itself. Be goor to have 10 mins on it today.

PR #61, opened 2015-02-25
Implemented SQS message delivery delay

luke-goodrich-je:
Added support for specifying a message delay so I can use SQS as a distributed timer, for example checking the status of an order after 10 minutes.


antonjefcoate:
Hey @luke-goodrich-je I'll take a look for you...

antonjefcoate:
It looks great!

PR #63, opened 2015-02-26
Update NuSpec to reflect actual dependencies, the main project "JustSaying" builds its own nuspec

tharris29:
Since the last release we have a separate nuget package of JustSaying.Tools


brainmurphy:
Do we need the nuspec file at all? Building the package from the csproj might be good enough.

payman81:
@tharris29 Do you want to close this PR?

PR #65, opened 2015-03-18
Ensure that batch size is set when throttling is strategy is in place.

tharris29:
In the throttling strategy you can specify a "maximumBatchSize" however this doesn't seem to be warranted and batch size is always 10.

````c#
    public class SqsNotificationListener : INotificationSubscriber
    {
.....
        private const int MaxAmazonMessageCap = 10;
````

### DIFF
--- a/JustSaying.AwsTools/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/SqsNotificationListener.cs
@@ -104,7 +104,7 @@ namespace JustSaying.AwsTools
                 var sqsMessageResponse = _queue.Client.ReceiveMessage(new ReceiveMessageRequest
                 {
                     QueueUrl = _queue.Url,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: JustSaying.AwsTools/SqsNotificationListener.cs [hunk 1]
```diff
@@ -131,6 +131,12 @@ internal void ListenLoop()
             }
         }
 
+        private int GetMaxBatchSize()
+        {
+            var maxMessageBatchSize = _messageProcessingStrategy as IMessageMaxBatchSizeProcessingStrategy;
+            return maxMessageBatchSize != null ? maxMessageBatchSize.MaxBatchSize : MaxAmazonMessageCap;
```
payman81: What if batch size is greater than the MaxAmazonMessageCap?

PR #66, opened 2015-03-18
Added queue name to exception logging

luke-goodrich-je:
Recently we had an issue inside the message read loop but it was not easy to ascertain whether one queue or all queues were affected as the queue name was not part of the log entry. This change adds the queue name to the log entries.

### DIFF
--- a/JustSaying.AwsTools/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/SqsNotificationListener.cs
@@ -123,11 +123,11 @@ namespace JustSaying.AwsTools
             }
             catch (InvalidOperationException ex)
             {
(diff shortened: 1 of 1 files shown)


JustSaying.AwsTools/SqsNotificationListener.cs
```diff
@@ -123,11 +123,11 @@ internal void ListenLoop()
             }
             catch (InvalidOperationException ex)
             {
-                Log.Trace("Suspected no messaged in queue. Ex: {0}", ex);
+                Log.Trace("Suspected no messaged in queue {0}. Ex: {1}", _queue.QueueName, ex);
```
brainmurphy: Since we're touching this line of code, fixing the grammar would be great ;-)

PR #69, opened 2015-04-09
Issue 25 - Use HandlerResolver to resolve message handlers via an IoC container.

payman81:


JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
```diff
@@ -60,6 +60,12 @@
     <Reference Include="Shouldly">
       <HintPath>..\packages\Shouldly.1.1.1.1\lib\35\Shouldly.dll</HintPath>
     </Reference>
+    <Reference Include="StructureMap">
```
payman81: This dependency is just for integration test purposes. JustSaying library has no dependency on any IoC container.

antonjefcoate:
Is StructureMapHandlerResolver something we can offer as a bolt-on so that people don't have to write code?
Some bolt-on approach for other containers would be cool...

payman81:
Yes @antonjefcoate , StructureMapHandlerResolver will be pushed to a separate library and published to JustSaying.StructureMap Nuget. I think we can roll this out in a separate step after we ensure the HandlerResolver is stable.

PR #72, opened 2015-04-15
issue-25 Respect IHandlerResolver's responsibility to create handlers for every incoming message.

payman81:


JustSaying/JustSayingFluently.cs
```diff
@@ -231,16 +231,18 @@ public IFluentSubscription IntoQueue(string queuename)
...
+            if(!handlerResolver.ResolveHandlers<T>().Any())
             {
                 throw new HandlerNotRegisteredWithContainerException(string.Format("IHandler<{0}> is not regsistered in the container.", typeof(T).Name));
             }
-            foreach (var handler in handlers)
+            if (handlerResolver.ResolveHandlers<T>().Count() > 1)
             {
-                Bus.AddMessageHandler(handler);
-                Log.Info(string.Format("Added a message handler - Topic: {0}, QueueName: {1}, HandlerName: {2}", _subscriptionConfig.Topic, _subscriptionConfig.QueueName, handler.GetType().Name));
+                throw new NotSupportedException(string.Format("There are more than one registration for IHandler<{0}>. JustSaying currently does not support multiple registration for IHandler<T>.", typeof(T).Name));
```
payman81: Allowing multiple registration requires a lot of thinking and perhaps major redesign. This can be done later.

antonjefcoate: Just to clarify, does this mean multiple handlers of the same message can no longer be applied?
OR, is it just when using IOC style handler resolution that the limitation is imposed?

antonjefcoate:
Aside from that, all looks good...

JustSaying/JustSayingFluently.cs
payman81: Yes that restriction is when you use IoC only.

PR #74, opened 2015-04-22
ExactlyOnce should use the actual handler's name not the wrapping decora...

payman81:
...tor to form the unique key for the message.

### DIFF
--- a/JustSaying.Messaging/MessageHandling/ExactlyOnceHandler.cs
+++ b/JustSaying.Messaging/MessageHandling/ExactlyOnceHandler.cs
@@ -8,12 +8,14 @@ namespace JustSaying.Messaging.MessageHandling
         private readonly IHandler<T> _inner;
         private readonly IMessageLock _messageLock;
         private readonly int _timeOut;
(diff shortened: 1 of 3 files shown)


JustSaying.Messaging/MessageHandling/ExactlyOnceHandler.cs
```diff
@@ -21,7 +23,7 @@ public ExactlyOnceHandler(IHandler<T> inner, IMessageLock messageLock, int timeO
         
         public bool Handle(T message)
         {
-            var lockKey = string.Format("{0}-{1}-{2}", _inner.GetType().FullName.ToLower(), typeof(T).Name.ToLower(), message.UniqueKey());
```
payman81: _inner is not suitable for this because it is always FutureHanlder<T> at this point.

3686:
Discussed bumping up the version but I understand the ExactlyOnce feature is currently undocumented and therefore unsupported right now. This could cause a breaking change if we've changed the format of the locking table.

Issue #77, opened 2015-05-21
Question: the logic behind the listen() loop in JustSayingBus.cs:86


dariusdamalakas:
Hi,

This is more of a question not an issue. It's note exactly clear from docs whether we can have handlers for more than one message. I assume it is, so this is what it looks like:
```cs
CreateMeABus.InRegion(RegionEndpoint.EUWest1.SystemName)
             .WithSqsTopicSubscriber()
             .IntoQueue(queue)
             .WithMessageHandler<SubscriptionAdded>(new SubscriptionHandler())
             .WithMessageHandler<SubscriptionRemoved>(new SubscriptionHandler())
             .StartListening();
```

Now, looking at [JustSaying.Bus.cs:86](https://github.com/justeat/JustSaying/blob/master/JustSaying/JustSayingBus.cs), the Start() method would have 2 subcsribers, because there are 2 subscribers by topic. However, both topic subscribers are listening for the same queue, so is it the case that both listeners will poll the same queue at same time?

This all makes sense for multiple regions - needs to listen for each region separately as been noted in [issue 59](https://github.com/justeat/JustSaying/issues/59). But this case is just for single region single queue.

I might have completely misunderstood how configuration is supposed to be used for multiple handlers.

Thanks

dariusdamalakas:
I've looked into this further, and I can see that listener per topic paradigm is core to JustSaying. 

There is a [unit test](https://github.com/justeat/JustSaying/blob/master/JustSaying.UnitTests/JustSayingBus/WhenRegisteringMessageHandlers.cs) which explicitly test that multiple listeners will be started for each topic, regardless of whether topic listens into the same queue or not. Line 60 and 61 both start listeners.

dariusdamalakas:
More info. The reason I am chasing this is because having 2 listeners listening to same queue actually impacts message delivery in a negative way. I've written a sample consumer and producer app. Out of 200 messages published, 10 to 20 messages are not handled correctly and are sent to DLQ. This is strange, even with 2 listeners polling SQS queue that should still work correctly, there probably must be some threading issue.  If I change sample to just have 1 handler (and 1 listener), then things go back to normal and 200 messages are processed significantly more quicker, on my specific case, at least twice or so.    Can post the sample if it's needed.

payman81:
Multiple listener per queue has been supported from the start but admittedly so far I've stuck to using single listener per queue for reasons other than performance e.g. uncomplicated retry policy when one handler fails.

[...]

PR #78, opened 2015-05-22
Listener per queue per region

dariusdamalakas:
Updating JustSaying so that only one listener would be created per queue, instead of listener per topic.

As discussed in [issue77](https://github.com/justeat/JustSaying/issues/77).


payman81:
@dariusdamalakas I'll close this PR as it has some breaking changes at its current state. I'll follow this up with you in other threads and find a resolution to your specific scenario.

PR #84, opened 2015-06-19
Ensuring that SQS publisher uses IMessageSerialiser to serialise messages

dariusdamalakas:
Note that JsonConvert.SerializeObject is still left as is, the final message pushed onto SQS queue must contain a Subject and Message fields. For SNS publishing, this is done automatically by AWS.


payman81:
PR looks good. Thanks.

PR #85, opened 2015-06-22
Allowing to override settings for NewtonsoftSerializer

dariusdamalakas:
Exposting `JsonSerializerSettings` as an optional constructor parameter, allows to pass in custom settings.


payman81:
PR looks good. Thanks.

PR #87, opened 2015-06-23
Topic name configurable through fluent API

SamHWhale:
We wanted to set the topic name per environment, so that we can use the same AWS account in all our environments.


payman81:
@SamHWhale Can you please update your PR before I can merge it? 
PR #89 that was merged earlier has caused conflicts with yours.

dariusdamalakas:
@SamHWhale, have a look at the other PR #96 , we are proposing a broader approach than just allowing to control topic name. We have the same requirement - we wanted to use same AWS acount for all of our environments. We wanted to control both the queu names and topic names. 

* Queue pattern - [env]-[instance]-[servicename]
* Topic patern - [env]-[instance]-[messagename]

The way to configure that would be to provide a custom `INamingStrategy` implementation, below would create topic and queue prefixed with environment.  

````c#
public class BusNamingConvention : IBusNamingConvention
{
    private string environment = "testEnvironment";
    public string GetTopicName(string topicName, string messageType)
    {
         return string.Format("{0}-{1}", environment, messageType);
    } 

    public string GetQueueName(string queueName, string messageType)
    {
         return string.Format("{0}-{1}", "testEvironment", queueName);
    }
}
````

Then use it like that:

````c#
     var busNamingConvetion = new BusNamingConvention("myEnvironment");
     CreateMeABus
                .InRegion(awsSettings.Regions)
                .WithSerialisationFactory(busConfigurator.ConfigureSerialisationFactory())
                .WithNamingStrategy(() => busNamingConvention);
````

Hope this will work for you.

SamHWhale:
Updated from JustEat/Master @dariusdamalakas, @payman81 decide if he wants to accept this. I'm happy to use either method. I didn't see the need to extend the queue name as there was a method to do this on the FluentApi already.

payman81:
@SamHWhale I've merged #96 which allows you to provide a naming strategy for topics and queues. Therefore, I'll close this PR.

PR #94, opened 2015-07-15
NLog upgrade 

dariusdamalakas:
Hi,

This is an NLog upgrade from 2.1.0 to 4.0.1
And also an upgrade for AWSSDK' from version '2.3.23.0' to '2.3.47.0'

Darius


brainmurphy:
@dariusdamalakas Looks like you've added an app.config to these projects, but not added the files themselves:

JustSaying.AwsTools.IntegrationTests
JustSaying.Messaging.UnitTests
JustSaying.IntegrationTests

Can you add those files? Looks good otherwise.

dariusdamalakas:
Ah, yes, I was wondering how they've got there, might be some nuget magic.  Updated.

PR #98, opened 2015-08-25
Expose Subscribers and Publishers via WhatDoIHave

tharris29:
Opening discussion on implementation

### DIFF
--- a/JustSaying.Messaging/Interrogation/IAmJustInterrogating.cs
+++ b/JustSaying.Messaging/Interrogation/IAmJustInterrogating.cs
@@ -1,0 +1,7 @@
+﻿namespace JustSaying.Messaging.Interrogation
+{
+    public interface IAmJustInterrogating
+    {
+        IInterrogationResponse WhatDoIHave();
+    }
+}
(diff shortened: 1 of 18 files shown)


JustSaying.Messaging/Interrogation/IAmJustInterrogating.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace JustSaying.Messaging.Interrogation
+{
+    public interface IAmJustInterrogating
```
petemounce: :-)

JustSaying.Messaging/Interrogation/InterrogationResponse.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class InterrogationResponse : IInterrogationResponse
+    {
+        public InterrogationResponse()
+        {
+            Subscribers = new List<ISubscriber>();
+            Publishers = new List<IPublisher>();
+        }
+
+        public ICollection<ISubscriber> Subscribers { get; }
```
petemounce: Should these be mutable?

tharris29: The contents of list is currently mutable. This is required to allow the response is built up over the time of the setup. The list itself is immutable I cannot think of a reason to make the list mutable.

payman81: @tharris29 I agree with @petemounce. Can you please update IInterrogationResponse to expose IEnumerable<T> instead of ICollection<T>?

payman81:
LGTM

PR #115, opened 2015-10-12
Update INamingStrategy.GetQueueName to take SqsReadConfiguration for …

payman81:
…easier extension.

### DIFF
--- a/JustSaying/DefaultNamingStrategy.cs
+++ b/JustSaying/DefaultNamingStrategy.cs
@@ -16,9 +16,9 @@ namespace JustSaying
             return messageType.ToLower();
         }
 
-        public string GetQueueName(string queueName, string messageType)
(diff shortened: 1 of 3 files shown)


payman81:
@brainmurphy Submitted this PR as discussed.

brainmurphy:
@payman81 Actually, I think that we wanted to make SqsReadConfiguration.InstancePosition public.

payman81:
I checked and it was public already https://github.com/justeat/JustSaying/blob/master/JustSaying.AwsTools/QueueCreation/SqsReadConfiguration.cs#L30 cc @brainmurphy

brainmurphy:
Oh, bugger, it's SubscriptionType I was thinking of! We need that to distinguish between whether we're subscribing to SQS directly or to SQS-subscribed-to-SNS.

PR #116, opened 2015-10-12
make SubscriptionType public in SqsReadConfiguration

payman81:


payman81:
@brainmurphy Here it is.

PR #117, opened 2015-10-13
Assembly version no longer in topic name

SamHWhale:
The topic name extension used type.FullName. This caused an issue in an Amazon policy as we increment our assembly versions.

### DIFF
--- a/JustSaying.AwsTools/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/SqsNotificationListener.cs
@@ -183,7 +183,7 @@ namespace JustSaying.AwsTools
                         var watch = new System.Diagnostics.Stopwatch();
                         watch.Start();
 
-                        handlingSucceeded = handle(typedMessage);
+                        handlingSucceeded &= handle(typedMessage);
 
                         watch.Stop();
                         Log.Trace("Handled message - MessageType: " + messageType);
--- a/JustSaying.Messaging/Extensions/TypeExtensions.cs
+++ b/JustSaying.Messaging/Extensions/TypeExtensions.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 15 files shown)


payman81:
@SamHWhale Did you have a failing test before making this change?

SamHWhale:
No was seeing an issue in AWSSDK, it was making a new policy every time and exceeded the maximum length. This fix ensures there is only one policy is created per topic. We version our assemblies on every build hence the issue. Do you require a test for this?

JustSaying.AwsTools/SqsNotificationListener.cs
```diff
@@ -183,7 +183,7 @@ public void ProcessMessageAction(Amazon.SQS.Model.Message message)
                         var watch = new System.Diagnostics.Stopwatch();
                         watch.Start();
 
-                        handlingSucceeded = handle(typedMessage);
+                        handlingSucceeded &= handle(typedMessage);
```
payman81: Hi @SamHWhale  - Can you please provide reasoning for this change?

JustSaying.Messaging/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace JustSaying.Messaging.Extensions
+{
+    public static class TypeExtensions
+    {
+        public static string ToKey(this Type type)
+        {
+            return string.Join("_", type.Flatten().Select(t => Regex.Replace(t.Name + "_" + t.Namespace, "\\W", "_"))).TruncateTo(100);
```
payman81: @SamHWhale - This is problematic for us because it also impacts non-generic message types. This is going to break our existing conventions for naming topics.

payman81:
I'll close this PR for now. @SamHWhale If you're happy to address the issues in the comments please feel free to reopen the PR.

PR #118, opened 2015-10-16
Update Tools to remove ini parsing; we're using a version of the SDK that supports profiles.

brainmurphy:


payman81:
Thanks @brainmurphy

PR #120, opened 2015-10-18
Increase the .Net version to 4.5.2 on all projects

AnthonySteele:
Bump up the .Net version to 4.5.2 on all projects. And fix package refs that needed it, and the `supportedRuntime` is declared in app.config files

AFAIK the intent is that we don't merge to master, but make this a V3.0.0 branch as it is a [backwards incompatible change](http://semver.org/) to anyone still using .Net 4.0.

Having said that, clients can be encouraged to target .net 4.5.2, [as this framework version is better supported](https://visualstudiomagazine.com/articles/2014/08/11/dot-net-support-deadline.aspx). You lose nothing and gain some by selecting  4.5.2 from the dropdown!


brainmurphy:
@AnthonySteele Change look OK to me. Let's call it JustSaying 2.2?

brainmurphy:
@AnthonySteele "Initial commit" - you're intending to do something else?

AnthonySteele:
Intending to follow this up with an immediate package update to JustBehave that could not be applied because of prior .Net version, and then in a longer term
- start thinking about async
- start thinking about the AWS SDK v3 package

payman81:
@AnthonySteele Can you please rebase this PR?

AnthonySteele:
@payman81 done.

PR #126, opened 2015-11-26
Allow message TimeStamp to be set when deserializing

stuart-lang:
Made TimeStamp property setter public so that Json.NET can restore it's state to whatever is in the json, rather than leaving it to the time of when it was deserialized.

### DIFF
--- a/JustSaying.Models/Message.cs
+++ b/JustSaying.Models/Message.cs
@@ -11,7 +11,7 @@ namespace JustSaying.Models
         }
 
         public Guid Id { get; set; }
-        public DateTime TimeStamp { get; private set; }
+        public DateTime TimeStamp { get; set; }
         public string RaisingComponent { get; set; }
         public string Version{ get; private set; }
         public string SourceIp { get; private set; }


payman81:
@stuart-lang Can you please add a unit test so your change won't get reverted later.

stuart-lang:
@payman81 The unit test is here:
https://github.com/justeat/JustSaying/blob/1f02191be04f9e392e82fd2c36280f7c0d8542c4/JustSaying.Messaging.UnitTests/Serialisation/Newtonsoft/WhenSerialisingAndDeserialising.cs#L36

I have the change ready to go, but I need this merged first because this is referenced by nuget package, so this needs publishing to nuget.org

payman81:
@stuart-lang I'll let you know once Nuget package for JustSaying.Model is updated.

payman81:
@stuart-lang JustSaying.Model version 2.0.0.269 has been published with your change.
http://www.nuget.org/packages/JustSaying.Models/2.0.0.269

PR #128, opened 2015-12-01
Improve SQS logging for multi-region scenarios (see SqsNotificationListener).

simonness:
Unfortunately although the SQS client is constructed with a region, it is not accessible after this, so have wrapped the client.


payman81:
Looks good to me.

PR #130, opened 2015-12-07
make SQS pull loop async

theburningmonk:


AnthonySteele:
This looks to me like a good, incremental change in the right direction. A lot of files are changed but the meat is in `SqsNotificationListener.Listen` ?

it is a smaller more incremental change than full async - the *handlers* are not async, just the listen loop, so it doesn't run into the issue of supporting both sync and async handlers-  I want my handlers to be async, but not breaking compatibility would be bad, and supporting both also complex.


things I don't understand are:
- The use of `CancellationTokenSource _cts = new CancellationTokenSource();` instead of `bool _listen = true;` Does it make a difference?
-  In the commit messages it says "so to not block thread for up to 20s for every subscriber in every region" - is that how it works now?

Caveats: 

- this PR is also mixing in C# v6 `$` string interpolation - do we want to do this as it is quite new and makes more demands on anyone building this, and do we want to do it in the same PR?

- it looks good, bu do the tests pass and what is the effect on performance? This is key.

theburningmonk:
regarding blocking for 20s for every subscriber in every region, here's the code I saw in `JustSayingBus`:
```
public void Start()
{
    lock (_syncRoot)
    {
        if (!Listening)
        {
            foreach (var regionSubscriber in _subscribersByRegionAndQueue)
            {
                foreach (var queueSubscriber in regionSubscriber.Value)
                {
                    queueSubscriber.Value.Listen();
                }
            }

            Listening = true;
        }
    }
}
```

and in the `Listen()` method:
```
Action run = () => { while (_listen) { ListenLoop(); } };
run.BeginInvoke(null, null);
```
which uses one ThreadPool thread each (and blocked during the SQS.ReceiveMessage call for 20s each):
```
var sqsMessageResponse = _queue.Client.ReceiveMessage(new ReceiveMessageRequest
{
    QueueUrl = _queue.Url,
    MaxNumberOfMessages = GetMaxBatchSize(),
    WaitTimeSeconds = 20
});
```

theburningmonk:
wrt `CancellationTokenSource`, right now, both implementations are wrong in that after cancellation is requested you'll still process SQS messages one last time (because cancellation is checked at the start).

With CTS,
a) it's more synonymous with the async paradigm, and
b) the `IsCancellationRequested` check is thread-safe (though in this case prob doesn't matter)

in a semantically correct implementation, ideally it'll pass the CancellationToken along and checked after the async wait on SQS receive returns before proceeding to process the messages. But that'll be a breaking change because it changes the way the listener currently works so I'm not sure if that's the right thing to do.

theburningmonk:
updated PR, please review

AnthonySteele:
Looks good as afar as I can see, but it will need a perf test!

theburningmonk:
:+1: sounds good

Issue #131, opened 2015-12-30
Introduce and allow to configure SNS/SQS client factory 


dariusdamalakas:
Need to be able to verify what SNS/SQS messages we are publishing.

Several ways to do that:
* Mock `IAmazonSimpleNotificationService`, (and SQS counterpart) and verify on `Publish` method
* Create custom `AmazonSimpleNotificationServiceConfig`, and set `ServiceURL` property. Then use this along with [WireMock](http://wiremock.org/)

In any case, factory for creating SNS/SQS clients should be exposed.

PR #133, opened 2015-12-31
Introducing IAwsClientFactory, and WithAwsClientFactory option

dariusdamalakas:
This PR addresses issue #131 . 

`JustSayingFluently`  has got a new option `WithAwsClientFactory`, which allows to pass in custom `IAwsClientFactory` implementations. This lets control credentials and even service URL by directly configuring SNS and SQS clients.

````c#
public class AwsClientFact : IAwsClientFactory
{
        private BasicAWSCredentials credentials;
        public AwsClientFact()
        {
            credentials = new BasicAWSCredentials("keyId", "secret");
        }

        public IAmazonSimpleNotificationService GetSnsClient(RegionEndpoint region)
        {
            AmazonSimpleNotificationServiceConfig config = new AmazonSimpleNotificationServiceConfig();
            config.ServiceURL = "http://localhost:8080/sns/";
            return new AmazonSimpleNotificationServiceClient(credentials, config);
        }

        public IAmazonSQS GetSqsClient(RegionEndpoint region)
        {
            var config = new AmazonSQSConfig();
            config.ServiceURL = "http://localhost:44334/sqs/";
            return new AmazonSQSClient(credentials, config);
        }
    }
````

Sorry for the longish PR, bulk of the changes are around due to the fact that `ISqsClient` got deprecated, and Region property was moved from `ISqsClient` to `SqsQueueBase`. Otherwise, no functional changes apart from introduction of sqs/sqs factory.


payman81:
This needs a rebase following the merge of your earlier PR.

payman81:
@dariusdamalakas Out of curiosity, what's your use case for overriding ServiceURL?

dariusdamalakas:
@payman81, sure, rebased.

With modified `ServiceURL` it is possible to direct JustSaying to a moceked SNS/SQS instances. Not sure whether this is practical, as that would require knowledge of SNS/SQS protocol, but certainly this is an option. 

One important feature this PR gives is ability to mock out both SNS and SQS clients, and assert on them. Here's an example from [PR 134](https://github.com/Intelliflo/JustSaying/blob/multipleQueueCreation/JustSaying.AwsTools.IntegrationTests/WhenSettingUpMultipleHandlers.cs)

dariusdamalakas:
@payman81 , i've added one more extra commit - this consolidates all AWS client creation throughout integration tests. Additionally, it is now possible to change the profile under which integration tests are being run by chaning `IntegrationTestConfig.AwsProfileName`. Currently it is still left as 'default', but I would consider changing this to something custom, such as 'JustSaying.IntegrationTests' or similar. 

I have a number of AWS profiles, and I need to switch default profile every time I want to run JustSaying integration tests.

Issue #86, opened 2015-06-23
Extra queue listing operations


dariusdamalakas:
Hi,

For each topic handler that we register, there is a call to ensure that topic and queues are created. I really love how you can just fire JustSaying, and it will auto create topics and queues for you. There is some inefficiency in that though. 

In [AmazonQueueCreator.cs:35](https://github.com/justeat/JustSaying/blob/fcb9e827c52cf93560bd69ebadfb4a76cbfd4d15/JustSaying.AwsTools/QueueCreation/AmazonQueueCreator.cs#L35) a `SqsQueueByName` is created. Inside constructor it calls `Exists()` to check whether that queue exists. Then again on the next line immediately afterwards, in call to `EnsureQueueAndErrorQueueExistAndAllAttributesAreUpdated` again the call is made to `Exists()` which again queries SQS and lists the queus.

So it's checking whether a queue exists 2 times in a row. Furthermore, it will perform those same checks against the same queue when the next topic handler will be registered. So we have here 2*M number of checks whether a single queue exists, where M is the number of topic handlers for given queue. It is possible to update this to just do a single call instead.

dariusdamalakas:
Same happens with topics. [SnsTopicByName.Exists()](https://github.com/justeat/JustSaying/blob/fcb9e827c52cf93560bd69ebadfb4a76cbfd4d15/JustSaying.AwsTools/SnsTopicByName.cs#L23) is called twice for each topic.

dariusdamalakas:
I have started working on a PR for this. There will be a lot of changes here, as I am going to refactor TopicBase and QueueBase classes. Unfortunately, the state they are in right now does not follow single responsibility principle, and does not lend easily to modification. 

One example that I have in mind here is `SnsTopicByName`. This class does multiple things at the moment. It implements interface 'IMessagePublisher', but it not only does publish messages, but also handles topic creation and subscription to queue. Guess this is also the reason of a dreadful comment in the code [ToDo: Grrr...why is this here even. GET OUT!](https://github.com/justeat/JustSaying/blob/master/JustSaying.AwsTools/SnsTopicBase.cs#L14).

Then there is 'IVerifyAmazonQueues' which should create queues and topics, and ensure these are subscribed, but what happens inside is that queue and topic creation gets passed onto queus and topics themselves.

[...]

PR #134, opened 2016-01-04
Multiple queue creation (issue #86)

dariusdamalakas:
This PR addresses issue #86. Queue and topic list and create operations are ran more often than needed. If a queue has been created, it is unnecessary to run another `ListQueue` request to check the existence of the queue.

The bulk of the changes relate to making the tests pass in fixture `WhenSettingUpMultipleHandlers`. 

This PR depends on #133, and will need to be rebased before merging.

### DIFF
--- a/JustSaying.AwsTools/QueueCreation/AmazonQueueCreator.cs
+++ b/JustSaying.AwsTools/QueueCreation/AmazonQueueCreator.cs
@@ -1,4 +1,3 @@
-using System;
 using Amazon;
 using JustSaying.Messaging.MessageSerialisation;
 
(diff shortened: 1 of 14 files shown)


JustSaying.AwsTools/QueueCreation/AmazonQueueCreator.cs
```diff
@@ -6,27 +5,49 @@ namespace JustSaying.AwsTools.QueueCreation
 {
     public class AmazonQueueCreator : IVerifyAmazonQueues
     {
+        private readonly IAwsClientFactoryProxy awsClientFactory;
+        private readonly IRegionResourceCache<SqsQueueByName> queueCache = new RegionResourceCache<SqsQueueByName>();
```
payman81: Private members should start with '_'. i.e. _awsClientFactory, _queueCache etc...

dariusdamalakas: @payman81 , thanks, will update this. I've seen underscores being used before, it's just we use different Resharper settings.  Will ping once done.

JustSaying.AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -0,0 +1,26 @@
...
+                return default(T);
+            var regionDict = this[region];
+
+            if (regionDict.ContainsKey(key) == false)
+                return default(T);
+            return regionDict[key];
+        }
+
+        public void AddToCache(string region, string key, T value)
+        {
```
payman81: Like most of the logic in registration of publishers and consumers, this is not thread safe. It's no issue for me as I'm only calling this in a single thread. Perhaps we should add a note in the readme about registration being non thread safe.

dariusdamalakas:
@payman81 , i have rebased this PR. Some of the comments have gone away, as they were part of #133.  This now contains just 1 commit.  As before, all the changes done where to make the `WhenSettingUpMultipleHandlers` test pass. The test checks that same queue and topic are only queried and created once, and not multiple times as before.

dariusdamalakas:
@payman81 thanks.

PR #141, opened 2016-01-15
Integration tests for AppVeyor

dariusdamalakas:
@brainmurphy , as discussed, this PR changes integration tests to use the new `IntegrationAwsClientFactory` instead of default AWS profile. 

Basically, this will look at 'ci' environment variable. If it is set (i.e. not empty), it will then look for ci-access_key and ci-secret_key values and will create an AWS basic profile credentials from that. 

You can use this to set-up integration tests to be run on AppVeyor.

### DIFF
--- a/JustSaying.TestingFramework/IntegrationAwsClientFactory.cs
+++ b/JustSaying.TestingFramework/IntegrationAwsClientFactory.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using Amazon;
+using Amazon.Runtime;
+using Amazon.SimpleNotificationService;
+using Amazon.SQS;
+using JustSaying.AwsTools;
+
+namespace JustSaying.TestingFramework
+{
+    /// <summary>
+    /// AwsCustomClient factory for running integration tests in continuous integration mode.
+    /// 
+    /// If environment variable "ci" is not defined, use default AWS profile name as specified by <code>IntegrationTestConfig.AwsProfileName</code>
+    /// 
+    /// Otherwise, construct AWS Profile using access and secret key by looking at "ci-access_key" and "ci-secret_key" environment variables.
+    /// </summary>
+    public class IntegrationAwsClientFactory : IAwsClientFactory
+    {
+        private readonly AWSCredentials _credentials;
+        private readonly string _ci = "ci";
+        private readonly string _ciAccesskey = "AWS_ACCESS_KEY_ID";
+        private readonly string _ciSecretkey = "AWS_SECRET_KEY";
+
+        public IntegrationAwsClientFactory()
+        {
+            var ci = System.Environment.GetEnvironmentVariable(_ci);
+            if (string.IsNullOrWhiteSpace(ci))
+                this._credentials = new StoredProfileAWSCredentials(IntegrationTestConfig.AwsProfileName);
+            else
+                _credentials = CredentialsFromEnvironment();
+        }
+
+        private AWSCredentials CredentialsFromEnvironment()
+        {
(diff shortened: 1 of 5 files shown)


petemounce:
Would you mind awfully using environment variables  AWS_ACCESS_KEY_ID and
AWS_SECRET_ACCESS_KEY? That will allow the AWS SDK to also pick up instance
profile credentials and potentially temporary credentials (via assume
role), since that is the naming convention AWS standardised on a little
while ago.

(The other sdks resolve credentials in a chain, stopping when they find
some, and so a change here would allow this to be used for more than just
testing)

--

dariusdamalakas:
@petemounce , interesting. The [JAVA SDK](http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/credentials.html) uses a different default credential provider chain that the [.Net SDK](http://docs.aws.amazon.com/AWSSdkDocsNET/latest/V2/DeveloperGuide/net-dg-config-creds.html#creds-assign), the difference is that Environment variables takes precedence in the Java SDK. 

I've updated PR with your change.

petemounce:
Yeah - the prefix and credential file are now standardised, but different SDKs order the resolution chain differently. Still, they're converging slowly...

Thanks for your contribution!

petemounce:
That green tick with checks passing is pretty nice, eh? ;)

dariusdamalakas:
@petemounce , yup, love the green tick.  

@brainmurphy, can you please set these environment vars on app veyor, and let's see what state we have at the moment with integration tests. I would expect at least a handful of them to fail.  We should then work to remove the flaky tests and get all integration tests pass consistently.

JustSaying.TestingFramework/IntegrationAwsClientFactory.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// If environment variable "ci" is not defined, use default AWS profile name as specified by <code>IntegrationTestConfig.AwsProfileName</code>
+    /// 
+    /// Otherwise, construct AWS Profile using access and secret key by looking at "ci-access_key" and "ci-secret_key" environment variables.
+    /// </summary>
+    public class IntegrationAwsClientFactory : IAwsClientFactory
+    {
+        private readonly AWSCredentials _credentials;
+        private readonly string _ci = "ci";
+        private readonly string _ciAccesskey = "AWS_ACCESS_KEY_ID";
+        private readonly string _ciSecretkey = "AWS_SECRET_KEY";
```
brainmurphy: Can we use AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY here? Consistency with the existing environment variables helps me to think less :wink:

dariusdamalakas: @brainmurphy, depends what is your source of truth :) 

[This C# SDK](http://docs.aws.amazon.com/AWSSdkDocsNET/latest/V2/DeveloperGuide/net-dg-config-creds.html#creds-assign) talks about credentials being stored in "`AWS_ACCESS_KEY_ID` and `AWS_SECRET_KEY` environment variables." (emphasis mine)

[AWS CLI](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-environment) uses 'AWS_ACCESS_KEY_ID' and 'AWS_SECRET_ACCESS_KEY'.

brainmurphy: @dariusdamalakas You're completely correct :grin: I'd prefer to be consistent with the C# world :+1:

PR #152, opened 2016-01-29
Review/refactor of SqsNotificationListener for clarity

AnthonySteele:
Review/refactor of SqsNotificationListener for clarity, as it is is a key class to understand justSaying
added `HandlerFunc` alias for `Func<Message, bool>` as it's a generic that is used a lot, inside other generics. 

e.g. instead of `Dictionary<Type, List<Func<Message, bool>>>`  we have ` Dictionary<Type, List<HandlerFunc>>`

reviewed other spacing and `{ }`

remove unused var `rawMessage`

extracted methods for `DeleteMessageFromQueue` and `CallMessageHandlers`

When try..catching deserialisation exceptions, put that around the `DeserializeMessage` call not the whole method.

And use the special exception type `MessageFormatNotSupportedException` introduced in 30/12/2015

Question: What should the "success" semantics be in `CallMessageHandlers` when there are multiple handlers for a message?
 Do **all** of them need to return true for success, or is it **any** of them? Existing code only uses only the return value of the last handler, which doesn't seem right.
How common are multiple handlers called for the same message? Could they be called in parallel if they're async ;)
What should happen when there are no handlers? Existing code regards this as success and does not log anything. I have kept this behaviour but it may not be best.

### DIFF
--- a/JustSaying.AwsTools/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/SqsNotificationListener.cs
@@ -232,7 +232,13 @@ namespace JustSaying.AwsTools
                 _onError(ex, message);
                 return;
             }
-       
+            catch (Exception ex)
+            {
+                Log.Error(ex, "Error deserialising message");
+                _onError(ex, message);
+                return;
+            }
+
             try
             {
                 var handlingSucceeded = true;
--- a/JustSaying.AwsTools/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/SqsNotificationListener.cs
@@ -7,12 +7,12 @@ using Amazon.SQS.Model;
 using JustSaying.Messaging.MessageProcessingStrategies;
 using JustSaying.Messaging.Monitoring;
 using NLog;
-using Newtonsoft.Json.Linq;
 using JustSaying.Messaging;
 using JustSaying.Messaging.Interrogation;
 using JustSaying.Messaging.MessageHandling;
 using JustSaying.Messaging.MessageSerialisation;
 using Message = JustSaying.Models.Message;
+using HandlerFunc = System.Func<JustSaying.Models.Message, bool>;
 
 namespace JustSaying.AwsTools
 {
@@ -22,7 +22,7 @@ namespace JustSaying.AwsTools
         private readonly IMessageSerialisationRegister _serialisationRegister;
         private readonly IMessageMonitor _messagingMonitor;
         private readonly Action<Exception, Amazon.SQS.Model.Message> _onError;
-        private readonly Dictionary<Type, List<Func<Message, bool>>> _handlers;
+        private readonly Dictionary<Type, List<HandlerFunc>> _handlers;
         private readonly IMessageLock _messageLock;
 
         private CancellationTokenSource _cts = new CancellationTokenSource();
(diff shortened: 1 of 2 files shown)


AnthonySteele:
Thoughts for later PRs. Parts of this library can't make up it's mind if it's Functional or OO. i.e. the interface   `IHandler<T> { bool Handle(T message); } ` is equivalent to `Func<T, bool>` but when converting one to the other and back you implicitly wrap and thus there's a performance and readability penalty. Pick one.

The use of `Action` to hand a blind piece of code out to the `IMessageProcessingStrategy` which is then routed back into the same class seems a bit convoluted and I think could be more straightforward.

We don't need a list of handlers for a message. If anyone even uses this, they could use a single handler along these simpelr and more flexible lines:

```csharp
    public class MultiHandler<T> : IHandler<T>
    {
        private readonly IHandler<T>[] _handlers; 

        public MultiHandler(params IHandler<T>[] handers)
        {
            _handlers = handers;
        }

        public bool Handle(T message)
        {
            var allHandlersSucceeded = true;
            foreach (var handler in _handlers)
            {
                var itemSuccess = handler.Handle(message);
                allHandlersSucceeded = allHandlersSucceeded && itemSuccess; // could aslo do an "or" version if the requirement is "any handler succeeded"
            }
            return allHandlersSucceeded;
        }
    }
```

JustSaying.AwsTools/SqsNotificationListener.cs
```diff
@@ -269,14 +260,53 @@ public void ProcessMessageAction(Amazon.SQS.Model.Message message)
...
+            }
+
+            var allHandlersSucceeded = true;
+            foreach (var handlerFunc in handlerFuncs)
+            {
+                var watch = new System.Diagnostics.Stopwatch();
+                watch.Start();
+
+                var thisHandlerSucceeded = handlerFunc(message);
+                allHandlersSucceeded = allHandlersSucceeded && thisHandlerSucceeded;
```
payman81: This fixes a subtle bug when there are multiple handlers for a message. A failure in one handler could've been ignored by success in another handler. Thanks.

AnthonySteele: Yes, it was just returning success from status of the the last handler.

payman81: The multiple handlers for a message scenario is not very common but it's certainly possible. 
Here is my thoughts on the questions you've raised.

You should not delete a message unless all handlers of that message return success otherwise the reliability of the library will be compromised. 

Execution of handlers could be parallelised but the decision whether to delete or keep the message can only be made after all handlers of a message have returned.

Simply deleting messages that have no handlers is ignoring a misconfiguration and is not correct. But I would address that separately and in isolation in case some users are relying on this behaviour.

payman81: Does removing support for multiple handlers per message simplify the move to async handlers?The only cost of dropping the support is that we push the responsibility to users and they should implement something similar to your MultiHandler<T>.

AnthonySteele: I agree that we shouldn't push this responsibility onto the users. But I think we can still pull it out of SqsNotificationListener, to make that class simpler and clearer (and handle it transparently to the user elsewhere in JustSaying). I might try a PR some time for that.

Users *can* still implement something like MultiHandler, if they really want, and we can't stop them nor should we try.

AnthonySteele: Agreed that " deleting messages that have no handlers" is questionable, but should be in a separate PR.

AnthonySteele: What is the right behaviour when coming across a message with no handler? 
Just ignore it or throw an exception. Might be a breaking change either way!

brainmurphy: It's worth noting that >1 handler-per-message type isn't supported by the IHandlerResolver path anyway: https://github.com/justeat/JustSaying/blob/master/JustSaying/JustSayingFluently.cs#L239

PR #154, opened 2016-02-04
HandlerMap class extracted from  the SqsNotificationListener

AnthonySteele:
Trying to break up the SqsNotificationListener
The HandlerMap turns out to be small part of it
Is it worth it? Should I add a couple of tests on the HandlerMap?

### DIFF
--- a/JustSaying.AwsTools/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/SqsNotificationListener.cs
@@ -12,7 +12,6 @@ using JustSaying.Messaging.Interrogation;
 using JustSaying.Messaging.MessageHandling;
 using JustSaying.Messaging.MessageSerialisation;
 using Message = JustSaying.Models.Message;
-using HandlerFunc = System.Func<JustSaying.Models.Message, bool>;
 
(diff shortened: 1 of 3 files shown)


JustSaying.AwsTools/SqsNotificationListener.cs
```diff
@@ -88,7 +80,7 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
             }
 
             Subscribers.Add(new Subscriber(typeof(T)));
```
AnthonySteele: Not entirely related, but why is this list of subscribers populated? Does any other code read it? If not, could we delete it without anything breaking?

brainmurphy: That list is used for the interrogation feature - to allow clients to show the current state of JustSaying. We use it at JUST EAT so that we can see, at run time, what events a component has subscribed to / what events it can raise.

brainmurphy:
@AnthonySteele I'll hold off merging this right now - let's see if we can enumerate the responsibilities lurking inside this class and how they can be extracted for maximum improvement...

PR #158, opened 2016-02-07
Extract the messageDispatcher class from SQSMessageListener 

AnthonySteele:
Extract the `MessageDispatcher` class from `SQSMessageListener` as discussed on Friday.

Any thoughts on naming?

I think that the right place to go to a new class is so that `messageProcessingStrategy` is between them. I find this neater than calling back to the same object. There is still an `Action` with no params or result needed for the threading model, but that may change later.

I would inline the method `SQSMessageListener.HandleMessage` but it is public and used in tests :/

### DIFF
--- a/JustSaying.AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying.AwsTools/MessageHandling/MessageDispatcher.cs
@@ -4,6 +4,7 @@ using Amazon.SQS.Model;
 using JustSaying.Messaging.MessageSerialisation;
 using JustSaying.Messaging.Monitoring;
 using Message = JustSaying.Models.Message;
+using SQSMessage = Amazon.SQS.Model.Message;
 
 namespace JustSaying.AwsTools.MessageHandling
 {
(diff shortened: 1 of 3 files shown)


JustSaying.AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -0,0 +1,125 @@
...
+            HandlerMap handlerMap)
+        {
+            _queue = queue;
+            _serialisationRegister = serialisationRegister;
+            _messagingMonitor = messagingMonitor;
+            _onError = onError;
+            _handlerMap = handlerMap;
+        }
+
+        public void ProcessMessage(SQSMessage message)
```
brainmurphy:
```csharp
public void DispatchMessage(SQSMessage message)
```
?

AnthonySteele: I will rename it, I have no strong feelings either way.

PR #161, opened 2016-02-08
Task end state

AnthonySteele:
Extract method to log task end state rather than use a big lambda. 

Previous ordering of cases was not correct: the faulted case would never be hit since IsFaulted is a *subset* of IsCompleted, it means "completed due to an unhandled exception" Check the docs: https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.iscompleted.aspx

We can log all "non-faulted" states the same way. The "faulted" state has it's own code - The "faulted" state is the same as IsFaulted  = true. Is the same as Exception != null
http://stackoverflow.com/a/32273273

I would like a unit test on these faulted tasks, as it's tricky! We're feeding async code into `Task.Run`, so the type is `Action<Task<Task>>`  - how would we know if we have the wrong task in the completion logging?

### DIFF
--- a/JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -101,19 +101,15 @@ namespace JustSaying.AwsTools.MessageHandling
             {
                 Log.Info(
                     "[Failed] Stopped Listening - {0}\n{1}",
-                    queueInfo,
-                    task.Exception);
-            }
-            else if (task.IsCompleted)
-            {
-                Log.Info(
-                    "[Completed] Stopped Listening - {0}",
-                    queueInfo);
+                     queueInfo,
+                     task.Exception);
             }
             else
             {
+                var endState = task.Status.ToString();
                 Log.Info(
-                    "[Canceled] Stopped Listening - {0}",
+                    "[{0}] Stopped Listening - {1}",
+                    endState,
                     queueInfo);
             }
         }
--- a/JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -88,32 +88,34 @@ namespace JustSaying.AwsTools.MessageHandling
                         await ListenLoop(_cts.Token);
                     }
                 })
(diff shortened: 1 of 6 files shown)


JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -88,34 +88,32 @@ public void Listen()
...
             Log.Info(
                 "Starting Listening - {0}", 
                 queueInfo);
         }
 
+        private static void LogTaskEndState(Task task, string queueInfo)
+        {
+            if (task.IsFaulted)
+            {
+                Log.Info(
```
AnthonySteele: Should this be a 'Warn' ? It's a change but it might be better.

JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -88,34 +88,32 @@ public void Listen()
...
+                    "[Faulted] Stopped Listening - {0}\n{1}",
+                     queueInfo,
+                     task.Exception);
+            }
+            else
+            {
+                var endState = task.Status.ToString();
+                Log.Info(
+                    "[{0}] Stopped Listening - {1}",
```
AnthonySteele: Is this particular log message format important? I would think that "Queue Foo Stopped in state Completed" would be better than "Completed - Stopped listening [Foo]"

AnthonySteele:
Also, the exception being logged is always an `AggregateException`. We might want to pull out a method to log those as text. And given that this might be needed again, is this better off as a new static class?

JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
brainmurphy: Agree.

AnthonySteele:
I have updated as discussed: extracted a method to log the details of an AggregateException

I'd like this new code to be in a different (static) class but we have agreed to defer this until we need it somewhere else as well.

What really needs to happen is to get a test around this, there is no coverage of the task faulted case. 
Which contributed to how it wasn't working before... And I haven't yet run that AggregateException logging code!

AnthonySteele:
I'm not totally happy with the test coverage of the error cases, but it's better than before. Ready to merge?

PR #163, opened 2016-02-10
Stronger test cases - when we test the results of calling a handler, check that the handler was actually called

AnthonySteele:
Making a PR to see if this passes in AppVeyor. 

These work locally, but might fail in appveyor, rendering other parts of the test invalid
i.e. if we're testing that no exception was logged, this is not a good test if the handler wasn't even called as expected.


AnthonySteele:
Merge?
It passed in appyeyor: https://ci.appveyor.com/project/justeattech/justsaying/build/3.0.302

PR #164, opened 2016-02-19
A bit more async

AnthonySteele:
Evolutionary change to Throttled message processing strategy rather than a big change: We can make `IMessageProcessingStrategy.BeforeGettingMoreMessages` awaitable so less busy-waiting when all the available slots for handlers are occupied and we have to wait for one to complete.

Important lines are in `Throttled`: ` public void BeforeGettingMoreMessages()` becomes `public async Task BeforeGettingMoreMessages()` and the synchronous wait at ` Task.WaitAny(activeTasksToWaitOn);` becomes `await Task.WhenAny(activeTasksToWaitOn);`

This gets most of the benefit for very little code change
And won't have to be changed when we go full async


AnthonySteele:
Great! Would like to see what the performance implications are. I suspect not much either way, but I'm not sure.

payman81:
Thanks. I don't expect much difference in performance. We'll find out this once we pilot it this week.

PR #165, opened 2016-02-20
Remove the MaximumThroughput message processing strategy

AnthonySteele:
As discussed on Friday, my understanding is that the "Maximum throughput" message processing strategy is not good for production use.
If you are using it, you can instead use a Throttled strategy with a large max.
As a result, I'm not quite sure what to put on line 45 of SqsNotificationListener - have to default to a Throttled strategy, but with what values?
What do you think?


payman81:
I agree that the unbounded MaximumThroughput should go. But can we have a slightly more flexible default. Something like this https://gist.github.com/payman81/21463be9c6f9319cf569.

AnthonySteele:
Right, I had forgotten that part, I will add it!

PR #170, opened 2016-02-29
call Unwrap on Task of task

AnthonySteele:
So `Task.Factory.StartNew(async () => await.... ` returns `Task<Task>` which is not what we want to inspect in the Task end state logging.

Unwrap as per  https://msdn.microsoft.com/en-us/library/dd780917.aspx


payman81:
Nice one. Thanks @AnthonySteele

AnthonySteele:
Thanks. One more thought: we have altered the "end of listener task" block of code a couple of times. 

How do we get a test around it so as to be 100%  sure of what it is doing?

Issue #171, opened 2016-02-29
Async JustBehave tests are not reliable


AnthonySteele:
In all the `JustSaying.AwsTools.UnitTests.MessageHandling.SqsNotificationListener.WhenMessageHandlingXXX` tests, I can change `Monitor.ReceivedWithAnyArgs().HandleException` to `Monitor.DidNotReceiveWithAnyArgs().HandleException` and vice versa and the test still passes. 

This means that they are not actually testing anything. It is likely that the `async` has gone wrong, something is declaring success too early, instead of awaiting actual pass or fail.

I have also had issues on a PR with tests that pass locally and fail on Appveyor, after emitting so much output that the Appveyor build results cannot be viewed.

The combination of JustBehave and `await Patiently.VerifyExpectationAsync` seems to be a bad one.

AnthonySteele:
Further investigation in `JustBehave`:
1) There is a `AsyncBehaviourTestBase` for when the `When` has signature ` protected override async Task When()`. We should use this?

The `[Then]` attribute derives from `NUnit.TestAttribute`, so should work when it's async.

We could also remove `JustBehave` from `JustSaying` entirely, but that's 63 tests to change. Do we want to embark on that?

AnthonySteele:
I thought about it a bit more, and this is how I see it:

Suppose we have an async '[Then]` method that tests that an exception was not thrown. How long would you wait to see if it happens? In the code as it stands, this would succeed the first time around the retry loop, and not wait any more. If the exception is thrown later it will not be noticed and the test will spuriously pass. The results will vary if a different `[Then]` method has already delayed things due to waiting for a pass on other conditions.

The way to make these tests work reliably would be to make the `[When]` method only exit when the state is ready to test, and have no asynchrony or delay in `[Then]` methods.

The collection of `[Then]` methods would represent a set of conditions that you want to test *at a point in time*, and it's the role of the `When` to get into this state, using async if need be.

petemounce:
So something in the base class waits for When's Task to complete before proceeding?

AnthonySteele:
Should o, if you inherit from [AsyncBehaviourTest](https://github.com/justeat/JustBehave/blob/master/src/JustBehave/AsyncBehaviourTest.cs)  (see also [AsyncBehaviourTestBase](https://github.com/justeat/JustBehave/blob/master/src/JustBehave/AsyncBehaviourTestBase.cs)

GraemeF:
[...]

PR #173, opened 2016-03-06
Look at synchrony in SqsNotificationListener tests

AnthonySteele:
This PR only changes unit tests. It looks at problems with the current  `WhenMessageHandling**` tests on `SqsNotificationListener`. 

### The current problem is:

The `[When]` in `BaseQueuePollingTest` fires off a test scenario and does not wait for it to complete. After this, the `[Then]` methods all loop and wait for the specified condition to be met.

But when this is e.g. `Monitor.DidNotReceiveWithAnyArgs().HandleException("Godot will come")`  how long should you wait to be sure that this will not happen?

What actually happens is that `Patiently.VerifyExpectationAsync` checks a function repeatedly and it will pass the first time, regardless of whether an exception is thrown later or not.

So the test does not verify what it claims to, and could start failing due to timing issues in other `[Then]` methods if they are run, and in what order. Or it could exhibit behaviour dependant on how the test runner works. We use 3 test runners: Resharper, NCrunch and AppVeyor. Differences have been observed - appveyor breaks completely on tests that pass locally.

### My strategy for fixing this is:

1) Use the `AsyncBehaviourTest` base class. The `[When]` method is now async and must not complete until the system under test has completed the test and has been stopped. This is the only async or "patient" part of the test.

2) We strongly prefer to not guess at this test being complete by using long delays or retry loops. Instead we send a "done" signal using inter-thread communication mechanisms, in this case a `TaskCompletionSource` when the work is done. Where we need to wait for one thread to get ahead of another, the wait should be short. The expected pass should be almost as fast as the system under test.

3) The `[Then]` methods inspect the terminal state of the system at a point in time, and do not wait. I do not recommend using the "`Patiently`" class to retry: the implementation is fine, but the idea is not useful here.

To an extent this is yak-shaving, but also we need to be confident in these tests before we can make other changes.

We still have a lot of noisy messages `Issue in message handling loop for queue , region eu-west-1` in the appveyor build log - this seems to be because in some tests, the message loop spins around quite  few times before it gets shut down, but it's no longer breaking things and the build passes in under 2 minutes.

There are more places that this pattern and helpers can be applied, but I'd like to get this merged first as it's a cohesive section.


AnthonySteele:
This addresses https://github.com/justeat/JustSaying/issues/171

JustSaying.AwsTools.UnitTests/MessageHandling/SqsNotificationListener/WhenExactlyOnceIsAppliedToHandler.cs
```diff
@@ -10,46 +10,47 @@ namespace JustSaying.AwsTools.UnitTests.MessageHandling.SqsNotificationListener
...
-    public class MyHandler : IHandler<GenericMessage>
-    {
-        private bool _handlerWasCalled;
-        public bool Handle(GenericMessage message)
+        [Test]
+        public void ProcessingIsPassedToTheHandler()
         {
-            _handlerWasCalled = true;
-            return true;
+            ((ExplicitExactlyOnceSignallingHandler)Handler).HandlerWasCalled();
```
brainmurphy: No assertion?

AnthonySteele: Whoops. I'll fix that!

PR #176, opened 2016-03-08
Fix typo

AnthonySteele:
Spotted this in the test output - "lose" not "loose" http://theoatmeal.com/comics/misspelling
and remove redundant string.Format


AnthonySteele:
Approved by @brainmurphy

PR #180, opened 2016-03-09
More detailed nlog output in appveyor

AnthonySteele:
Typical output in appveyor includes a lot of lines reading `Issue in message handling loop for queue , region eu-west-1`
This is the output of a nlog logging line that also includes an exception. An exception is expected in some tests.
Use a nlog format so that the repetitive logging shows more details
This allows use to much more easily see which test is spinning.

### DIFF
--- a/JustSaying.AwsTools.UnitTests/JustSaying.AwsTools.UnitTests.csproj
+++ b/JustSaying.AwsTools.UnitTests/JustSaying.AwsTools.UnitTests.csproj
@@ -110,6 +110,9 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
+    <None Include="NLog.config">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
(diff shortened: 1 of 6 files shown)


AnthonySteele:
Hm. Appveyor is still showing a raw output of just the message `Issue in message handling loop for queue , region eu-west-1` from the statement `Log.Error(ex, msg);` in `SqsNotificationListener` line 201.

This is intended to make debugging easier, but is itself hard to debug. I don't know how appveyor is finding files and so if the `nlog.config` file is valid or is picked up by appveyor.

petemounce:
You can load the nlog config programmatically if you need, or indeed
configure nlog itself programmatically.

--

AnthonySteele:
We might have to do that. It's not clear at present what it's doing - the output doesn't correspond to either the config file or the default layout mentioned in the online docs. It seems to be just `{'${message}'} ` so who set that?

PR #182, opened 2016-03-15
Log more structured data in tests

AnthonySteele:
Log more structured data in tests by configuring NLog in code
we use the `Logging` helper class to programmatically set the log format as suggested by GraemeF.
fields include: timestamp, log message, exception type and message

The `TestException` class is moved into `TestingFramework` so it can be used elsewhere.

You can see what this PR does by looking at the output of test in Appveyor:

The line logged changes from e.g. 
`Issue in message handling loop for queue , region eu-west-1`
to
`12:57:21.8846|Error|Issue in message handling loop for queue , region eu-west-1|TestException testing the failure on first call`

### DIFF
--- a/JustSaying.TestingFramework/packages.config
+++ b/JustSaying.TestingFramework/packages.config
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="AWSSDK" version="2.3.52.0" targetFramework="net452" />
+  <package id="NLog" version="4.2.0" targetFramework="net452" />
   <package id="NUnit" version="3.0.1" targetFramework="net452" />
 </packages>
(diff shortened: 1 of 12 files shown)


JustSaying.TestingFramework/packages.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="AWSSDK" version="2.3.52.0" targetFramework="net452" />
+  <package id="NLog" version="4.2.0" targetFramework="net452" />
```
petemounce: 4.2.3 is latest - any reason to not use that?

AnthonySteele: NLog V 4.2.0 is that the other projects in the solution use. I take the view that an inconsistent package version is worse than being consistently a minor release or 2 behind.
We should update to latest at some point, in a different PR?

petemounce: Ah, agree.

PR #190, opened 2016-03-31
Throttling refactoring

AnthonySteele:
This is the throttling simplification that Brain, Payman and I worked on in February.
i think it's ready for another look.
before merge, we will want to review, and to run it with some load for sanity & perf testing!


slang25:
@AnthonySteele I hopefully have some fixes to the tests, just getting a PR ready now. It's to do with:
```
messageProcessingStrategy.ProcessMessage(async () => await tcs.Task);
```
Because the method task an ```Action```, it's basically like awaiting inside of a void method. You need to switch to ye olde ```.Wait()``` TPL stylie.

The other failures are fixed with an ```await Task.Yield()```

PR #192, opened 2016-04-01
Throttling refactoring

AnthonySteele:
This is the throttling simplification that Brain, Payman and I worked on in February.
i think it's ready for another look.
before merge, we will want to review, and to run it with some load for sanity & perf testing!
Squashed the commits to aid redability


AnthonySteele:
Works fine under overnight load test

PR #200, opened 2016-04-15
Not patiently

AnthonySteele:
More removal of `Patiently` in Integration Tests

- building on previous PR that moved `Tasks.cs` to the TestingFramework project
- use `TaskCompletionSource` to signal when the "When" part of a test is done
- The "Then" part is always sync, never awaits, it inspects the final state of a stopped system.
- Removed some of the `Patiently` class as it was unused. But not yet all of it.

For the reasons behind this design, see https://github.com/justeat/JustSaying/pull/173


AnthonySteele:
This passes tests on appveyor, but the integration tests are not run there. 

They do run and pass locally.

PR #203, opened 2016-04-20
Prefer async handlers

AnthonySteele:
Ready for review and/or beta

Introduce [`IHandlerAsync`](https://github.com/AnthonySteele/JustSaying/edit/prefer-async-handlers/JustSaying.Messaging/MessageHandling/IHandlerAsync.cs?pr=/justeat/JustSaying/pull/203) as the async replacement for `IHandler`.  `IHandlerAsync` is used throughout internally, and `IHandler` is deprecated.
 [`BlockingHandler`](https://github.com/AnthonySteele/JustSaying/blob/prefer-async-handlers/JustSaying.Messaging/MessageHandling/BlockingHandler.cs) adapts legacy `IHander` to `IHandlerAsync` It is used as an adapter in the overload  `JustSayingFluently.WithMessageHandler(IHandler handler)`, though `WithMessageHandler(IHandlerAsync handler)` is preferred.

Going live with this code will  likely imply a new major version, and so we may want to roll in other breaking changes such as message envelopes before we do this.

Known issues:

Done:
- Performance characteristics: tested in Beta LegacyOrderView. it does not cause errors, performance is no worse, may be slightly better. This is probably due to there not being much actual awaiting in the code yet, just one call. 
- We have decided not to mark the interface `IHandler` as obsolete just yet. We can always add this in soon. Though in the medium/long term we want to push  consumers to `IHandlerAsync` instead of `IHandler`
- tests are green.
- while `IHandlerResolver` is just a contract, the sample implementation. `StructureMapHandlerResolver` in the tests, will also locate legacy `IHandler` instances and wrap them in  `AsyncingHandler`. And tests on this working in integration.
- When an `IHandler`. is wrapped with `BlockingHandler `, we will read the `ExactlyOnceAttribute` for the inner object.


GraemeF:
I would say marking `IHandler` obsolete is fine - you don't have to upgrade to a new major version if you don't want to deal with breaking changes.

AnthonySteele:
I agree with you, but now I'm seeing lots of `pragma warning disable 618` in the code and tests in order to work with obsolete (but still supported) `IHandler` things

JustSaying.Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -65,7 +65,7 @@ public int FreeTasks
 
         public void ProcessMessage(Func<Task> action)
         {
-            var task = new Task<Task>(async () => await action());
+            var task = new Task<Task>(action);
```
stuart-lang: Why ```new Task``` then ```Unwrap``` out of interest. Why not just invoke then ```ContinueWith```?

AnthonySteele: Hm, I thought it was for the Threadpool scheduling, but would it work without that? So long as there is still the `task.Start()` after?

brainmurphy:
Love it!

PR #205, opened 2016-05-30
rename "FreeTasks" to "AvailableWorkers"

AnthonySteele:
On [`IMessageProcessingStrategy`](https://github.com/justeat/JustSaying/blob/master/JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs), rename `FreeTasks` property to `AvailableMessageHandlers` and add `MaxConcurrentMessageHandlers`.  i.e from 

````csharp
    public interface IMessageProcessingStrategy
    {
        int FreeTasks { get; }

        Task AwaitAtLeastOneTaskToComplete();
        void ProcessMessage(Func<Task> action);
    }
`````
to

````csharp
    public interface IMessageProcessingStrategy
    {
        int MaxConcurrentMessageHandlers { get; }
        int AvailableMessageHandlers { get; }

        void ProcessMessage(Func<Task> action);
        Task AwaitAtLeastOneTaskToComplete();
    }
`````

Thoughts on naming here?

I feel that `AvailableMessageHandlers` is a better name as `FreeTasks` could be misread as an action "this method will free up all the tasks".

Also surfaced the `MaxConcurrentMessageHandlers` value as I want to log this at startup (to monitor what the `DefaultThrottledThroughput` is doing). You *can* use `FreeTasks` for that on a newly created `IMessageProcessingStrategy` on the assumption that, at startup, the current available count equals the max available count, but that's not great.

This is a minor breaking change (which probably affects mostly me) as `FreeTasks` was added quite recently.

And added comments to `IMessageProcessingStrategy`.

### DIFF
--- a/JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
+++ b/JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
@@ -13,7 +13,7 @@ namespace JustSaying.Messaging.MessageProcessingStrategies
         /// <summary>
         /// The number of worker tasks that are free to run messages handlers right now,
         /// i.e. MaxWorkers - (the number of currently running workers)
-        /// Always in the range: 0 >= AvailableWorkers >= MaxWorkers
+        /// Always in the range 0 to MaxWorkers
         /// </summary>
         int AvailableWorkers { get; }
 
--- a/JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
+++ b/JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
@@ -24,10 +24,10 @@ namespace JustSaying.Messaging.MessageProcessingStrategies
         void StartWorker(Func<Task> action);
 
         /// <summary>
-        /// after awaiting this, AvailableWorkers should be > 0,
-        /// i.e. you are in a position to add another one by calling ProcessMessage
+        /// After awaiting this, you should be in a position to start another worker
+        /// i.e. AvailableWorkers should be above 0
         /// </summary>
         /// <returns></returns>
-        Task AwaitAtLeastOneWorkerToComplete();
+        Task WaitForAvailableWorkers();
     }
 }
--- a/JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
+++ b/JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
@@ -18,10 +18,10 @@ namespace JustSaying.Messaging.MessageProcessingStrategies
         int AvailableWorkers { get; }
 
         /// <summary>
-        /// Start processing a message. 
+        /// Launch a worker to start processing a message.
         /// </summary>
         /// <param name="action"></param>
-        void ProcessMessage(Func<Task> action);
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by stuart-lang** [comment]
I see what you mean about ```FreeTasks```, you could go with ```TasksFree```, but I agree it's good to have consistent and clear naming. I think of Handlers as the actual classes and implementation of ```IHandler```, and the tasks as ```Worker```s, that would be a bigger code change, but I think that would be clearer and a little more consise, for example ```MaxWorkers``` instead of ```MaxConcurrentMessageHandlers```

**Comment by AnthonySteele** [comment]
That's a fair point - workers are threads that execute the handler code, 

so `MaxWorkers` and `AvailableWorkers` ?

**Comment by stuart-lang** [comment]
Yeah =) but again that's a bigger change, not sure how much of the external apis it would break

**Comment by AnthonySteele** [comment]
I don't think any external apis would be broken; and anyway this should go in just before a major version number increment.

**Comment by AnthonySteele** [comment]
Is now:

````csharp
    public interface IMessageProcessingStrategy
    {
        int MaxWorkers { get; }
        int AvailableWorkers { get; }

        void ProcessMessage(Func<Task> action);
        Task AwaitAtLeastOneWorkerToComplete();
    }
`````

File: JustSaying.Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs [hunk 1]
```diff
@@ -5,9 +5,30 @@ namespace JustSaying.Messaging.MessageProcessingStrategies
...
-        Task AwaitAtLeastOneTaskToComplete();
+        /// <summary>
+        /// The maximum number of worker tasks that will be used to run messages handlers at any one time
+        /// </summary>
+        int MaxWorkers { get; }
+
+        /// <summary>
+        /// The number of worker tasks that are free to run messages handlers right now,
+        /// i.e. MaxWorkers - (the number of currently running workers)
+        /// Always in the range: 0 >= AvailableWorkers >= MaxWorkers
```
stuart-lang: ```0 <= AvailableWorkers <= MaxWorkers```

AnthonySteele:
This may be a stretch, but how would you feel about renaming `ProcessMessage` to match the "Workers" theme, e.g. `StartWorker` ??

AnthonySteele:
Is now:

````csharp
    public interface IMessageProcessingStrategy
    {
        int MaxWorkers { get; }
        int AvailableWorkers { get; }

        void StartWorker(Func<Task> action);
        Task AwaitAtLeastOneWorkerToComplete();
    }
`````

stuart-lang:
I think that this is an improvement 👍
Not sure a method should start with Await, doesn't quite look right to me. Maybe ```WaitForAtLeastOneWorkerToComplete()``` or ```WaitForAvailableWorkers()```, splitting hairs now.

AnthonySteele:
I agree that a method name that starts "Await" is odd, and also I find the "AtLeastOneWorker" bit cumbersome, so I think `WaitForAvailableWorkers()` is the one that I like the most.

AnthonySteele:
Is now:

````csharp
    public interface IMessageProcessingStrategy
    {
        int MaxWorkers { get; }
        int AvailableWorkers { get; }

        void StartWorker(Func<Task> action);
        Task WaitForAvailableWorkers();
}
````

If we're happy with these names then lets close this one?

PR #207, opened 2016-06-06
Removed NLog dependancy

stuart-lang:
To make JustSaying a bit more friendly to newcomers I have I've pulled out NLog and put in [LibLog](https://github.com/damianh/LibLog), cause having a library force you to use a particular logging library ain't cool.
There should be no code changes required for consuming code that already has a NLog dependency loaded in their app.

One problem I have with this PR is that I have 2 versions on LibLog, one in JustSaying, and one in JustSaying.AwsTools, there is a tiny bit of reflection on the first use to determine if NLog or another supported logging library is loaded, by including 2 versions, we do this twice. The alternatives are to add only one to AwsTools, but this isn't really aws specific, so it would be wrong really. We could add another shared library, but then that's another dll that needs to ship.

Thoughts?


slang25:
Given that this is changing a dependency, it might be worth commenting on .Net Core support. LibLog doesn't immediately support core but it should be relatively easy to add because it already ships as a PCL. There is an issue on their repo I'll dig out later that discussed Core CLR support.

stuart-lang:
And yes, I am having a conversation with myself using 2 different accounts. Need more coffee ☕

AnthonySteele:
I agree that for general use, it isn't great to insist that clients *will* use NLog; and breaking this dependency is good. I don't know `LibLog` or how it does this, will have to read up on it.

payman81:
I'm not familiar with LibLog but the file with 2000 lines of code gave me a shock.
But I absolutely agree that we must break NLog dependency in JustSaying.
Just to get things started, why don't you extract all Nlog references out into its own package i.e. JustSaying.Nlog.

slang25:
@payman81 Yeah, because there are 2 versions of the cs files added, this PR would make me the largest contributor, which isn't right 😆 

I'm thinking that I'll try again, but with a separate project named JustSaying.Logging, with essentially a trimmed down, more fitted version of LibLog. I like it's discovery mechanism, it just checks that it can resolve a type from the NLog library (or whatever library), if it can then it will use NLog, no hard dependancy, and no code changes for consumers.

slang25:
Closing in preference of a lighter approach, a new PR is on the way soon

PR #209, opened 2016-06-07
Using AWS SDK Version 3

Liewe:
As discussed with my colleague (@edu115) we updated the AWS SDK to version 3. 
The only real change we did was in SqsQueueByName.cs: 

We would get an error in unittest WhenOptingOutOfErrorQueue.ErrorQueueShouldNotBeCreated: 
"Value {"maxReceiveCount":"5", "deadLetterTargetArn":""} for parameter RedrivePolicy is invalid. Reason: Invalid value for deadLetterTargetArn.". 
Since this only happens when opting out of an errorqueue and without an error queue you don't need a redrive policy, we moved the call to UpdateRedrivePolicy to within the errorqueue create/update condition block.


Liewe:
I'm going to re-do this from the company account

PR #210, opened 2016-06-07
Using AWS SDK Version 3

Liewe:
As discussed with my colleague (@edu115) we updated the AWS SDK to version 3.
The only real change we did was in SqsQueueByName.cs:

We would get an error in unittest WhenOptingOutOfErrorQueue.ErrorQueueShouldNotBeCreated:
"Value {"maxReceiveCount":"5", "deadLetterTargetArn":""} for parameter RedrivePolicy is invalid. Reason: Invalid value for deadLetterTargetArn.".
Since this only happens when opting out of an errorqueue and without an error queue you don't need a redrive policy, we moved the call to UpdateRedrivePolicy to within the errorqueue create/update condition block.

### DIFF
--- a/JustSaying.AwsTools.IntegrationTests/JustSaying.AwsTools.IntegrationTests.csproj
+++ b/JustSaying.AwsTools.IntegrationTests/JustSaying.AwsTools.IntegrationTests.csproj
@@ -36,8 +36,16 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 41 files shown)


AnthonySteele:
Hi, do you want to update this PR for current master? I looked at it and the conficts is easy to fix - just 1 line in `SqsQueueByName.cs`.

slang25:
@Liewe We'd like this to be merged to the develop branch, which will be where we'll be adding features to become v5.

Liewe:
I'll have a look in monday.

Liewe:
I upmerged master. @slang25 I created a new pullrequest targeting the develop branch:  https://github.com/justeat/JustSaying/pull/222

slang25:
Closing in favor of #222. Thanks @Liewe!

Issue #206, opened 2016-06-01
IHandlerResolver should support releasing handlers


codedemonuk:
IHandlerResolver has no method to allow the resolved handler to be released.  This can cause memory leaks in IoC containers such as Ninject, Windsor where you need to release components.

Would it be possible to add a release method to the IHandlerResolver and use that to allow handlers to be explicitly released by their IoC containers when people implement their own IHandlerResolvers.

payman81:
Why don't you implement IDisposable in your handlers and let your IoC container release them when they're out of scope?
Can you share your implementation of IHandlerResolver or send me a gist that shows the memory leak?
Implementation of IHandlerResolver isn't meant to hold reference to any IDisposable. It should simply pass the call to the IoC container. Here is an implementation for StructureMap:
````
public class StructureMapHandlerResolver : IHandlerResolver
{
    private readonly IContainer _container;

    public StructureMapHandlerResolver(IContainer container)
    {
        _container = container;
    }

    public IEnumerable<IHandler<T>> ResolveHandlers<T>()
    {
        return _container.GetAllInstances<IHandler<T>>();
    }
}
````

codedemonuk:
The issue is not a defect in JustSaying.

A number of IoC containers such as Castle Windsor keep track of all of the components that have been resolved (including transient objects) and the container will not release it's reference to these components until the container's `Release` method has been called.

This results in a memory leak because the CLR garbage collection will not collect these transient objects because Windsor will always have a reference to the object.

More details about this behaviour are in the [Castle Windsor documentation][1] 

Other frameworks like Quartz accommodate this behaviour in IoC frameworks by including a method on their interfaces to allow a resolved component to be released - [an example of how Quartz handles it][2].

[1]: https://github.com/castleproject/Windsor/blob/master/docs/release-policy.md
[2]: https://github.com/quartznet/quartznet/blob/b03aa020168a63ab062cb703804e845a07c2003f/src/Quartz/SPI/IJobFactory.cs

AnthonySteele:
Just for curiosity, have you observed a memory leak, and how severe is it?

stuart-lang:
I can see how this would be a problem for you. MEF works the same way with regards to holding onto ```IDisposable```, we "fixed" it the time by introducing an ```IClosable``` (someone wanted us to call it ```IDismosable``` lol). You container holding on to this reference will keep your whole object graph alive, not good.

[...]

PR #215, opened 2016-06-13
Handler instance per request from IHandlerResolver

AnthonySteele:
When using `IHandlerResolver`, resolve the handler per request, not once upfront
This needs tests (this behaviour was changed without tests breaking, so it's not well defined)  and docs and maybe breaking changes as the other overload of `WithMessageHandler` works differently.

But a quick, non-breaking start is going back to old behaviour (it was like this earlier this year: https://github.com/justeat/JustSaying/blob/3aee35f286c247ba2739f6ba57ec770a13dc7656/JustSaying/JustSayingFluently.cs#L247) as per this discussion https://github.com/justeat/JustSaying/issues/206#issuecomment-225610603


slang25:
@AnthonySteele This makes sense, shall I merge and we do a minor release?

AnthonySteele:
Yes, that's a good idea.

PR #216, opened 2016-06-18
Issue 211 - Prevent reaching max allowed size in SQS policy 

payman81:
Instead of adding a policy per topic just use a single policy with a wildcard.


stuart-lang:
@payman81, this looks good 👍. I'll wait for others to comment before merging

dariusdamalakas:
@stuart-lang i'm ok with this, this is an intermediary fix for allowing to control queue Policy

PR #218, opened 2016-07-02
DNM: Use "ConfigureAwait(false) everywhere

AnthonySteele:
Use "ConfigureAwait(false) everywhere
via a wrapper fn that is shorter and expresses intent
see:
http://blog.stephencleary.com/2012/02/async-and-await.html
http://stackoverflow.com/a/38118550/5599
http://stackoverflow.com/questions/13489065/best-practice-to-call-configureawait-for-all-server-side-code

What do you think?

### DIFF
--- a/JustSaying.AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying.AwsTools/MessageHandling/MessageDispatcher.cs
@@ -7,6 +7,7 @@ using JustSaying.Messaging.MessageSerialisation;
 using JustSaying.Messaging.Monitoring;
 using Message = JustSaying.Models.Message;
 using SQSMessage = Amazon.SQS.Model.Message;
+using JustSaying.Messaging;
 
 namespace JustSaying.AwsTools.MessageHandling
 {
@@ -63,7 +64,7 @@ namespace JustSaying.AwsTools.MessageHandling
 
                 if (typedMessage != null)
                 {
-                    handlingSucceeded = await CallMessageHandlers(typedMessage);
+                    handlingSucceeded = await CallMessageHandlers(typedMessage).OnAnyThread();
                 }
 
                 if (handlingSucceeded)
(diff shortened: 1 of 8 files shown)


slang25:
@AnthonySteele I have a local branch with a similar change 😄. You are right to do this, this is a library, it is for the consumer to determine if they want to continue on the same synchronization context. I personally would just use ```.ConfigureAwait(false)```, it's not really about threads, it happens to be a side affect of awaiting cpu bound (or thread bound) tasks (and even then, this is an implementation detail), but a lot of this is when awaiting IO bound async tasks (with no threads, unless you count IO completion port thread-pool threads). It's about not requiring preserving the synchronization context when awaiting, something like ```.ContinueOnAnySychronizationContext()``` would be semantically correct, but this is not shorter and you cannot get any static tooling or analysis help.
I am using this resharper plugin to make sure I don't miss any awaits:
https://github.com/aelij/ConfigureAwaitChecker/
^ we might be able to use the Roslyn analyser version of this as a nuget package, so that any future devs don't forget to add ```.ConfigureAwait(false)```. That would my preference.

slang25:
My comment above regarding IO/CPU bound tasks is not relevant I just realised. It's about where the resuming work happens. I guess you could saying what we are meaning can be phrased a number of ways:
```.OnAnySychronizationContext() == .UseTheContextMostConvienient() == .OnAnyThread()```
^ and it's that last version that we actually care about, because this is our precious resource. But we are breaking the abstraction (Tasks) by making statements about threads, and we lose out on static analysis by introducing our own mechanism for handling this.

AnthonySteele:
I'm still not thrilled by `.ConfigureAwait(false)` but at least it's well known. In the light of day the extension method may be overreach. 

Do you want to submit your PR instead?

AnthonySteele:
Closed in favour of https://github.com/justeat/JustSaying/pull/219

PR #232, opened 2016-07-24
DNM: Separate named queues from default naming

AnthonySteele:
Intended to make the difference clear when the subscriber is trying to control multiple queues or not
What do you think?
or a deeper look at `INamingStrategy` instead?

### DIFF
--- a/JustSaying/JustSayingFluently.cs
+++ b/JustSaying/JustSayingFluently.cs
@@ -191,7 +191,7 @@ namespace JustSaying
         {
             if (string.IsNullOrWhiteSpace(queueName))
             {
-                throw new ArgumentNullException(nameof(queueName));
+                throw new ArgumentException("You must supply a queue name or use IntoDefaultQueue");
             }
 
             _subscriptionConfig.BaseQueueName = queueName;
--- a/JustSaying/JustSayingFluently.cs
+++ b/JustSaying/JustSayingFluently.cs
@@ -187,9 +187,20 @@ namespace JustSaying
 
         #region Implementation of Queue Subscription
 
-        public IFluentSubscription IntoQueue(string queuename)
+        public IFluentSubscription IntoQueueNamed(string queueName)
         {
-            _subscriptionConfig.BaseQueueName = queuename;
+            if (string.IsNullOrWhiteSpace(queueName))
+            {
+                throw new ArgumentNullException(nameof(queueName));
+            }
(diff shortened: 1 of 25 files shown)


JustSaying/JustSayingFluently.cs
```diff
@@ -187,9 +187,20 @@ public ISubscriberIntoQueue WithSqsPointToPointSubscriber()
 
         #region Implementation of Queue Subscription
 
-        public IFluentSubscription IntoQueue(string queuename)
+        public IFluentSubscription IntoQueueNamed(string queueName)
         {
-            _subscriptionConfig.BaseQueueName = queuename;
+            if (string.IsNullOrWhiteSpace(queueName))
+            {
+                throw new ArgumentNullException(nameof(queueName));
```
stuart-lang: Maybe include in the exception message something to saying use ```.IntoDefaultQueue()``` if that was their intent. They might have found a compile error after upgrading and done a wholesale find & replace of all ```IntoQueue```-> ```IntoQueueNamed```

AnthonySteele: Done. Still not sure if this is a good idea or not...

stuart-lang: What the whole change? or throwing here vs making the call to ```IntoDefaultQueue``` on their behalf?

AnthonySteele: I mean the whole change. It makes it clearer in 1 case, but makes this case harder:

var queueNameOrNull = GetAQueueNameOrNull();

...

.WithSqsTopicSubscriber()		     
.IntoQueue(queueNameOrNull)
...

stuart-lang: Hmm, I see what you are saying. Why not have ```IntoDefaultQueue``` be an optional alternative for ```.IntoQueue(string.Empty)``` and keep ```IntoQueue``` behaving how it does now.
In our usage we do the scenario you talk about, i.e. a variable that is sometimes ```string.Empty```

AnthonySteele: The intent is to reduce risk of accidentally getting 2 handlers on the same queue to a topic instead of 2 handlers each on their own queue  to a topic (or vice versa). I think that we will have to find other ways to do that, but still might keep `IntoDefaultQueue()` as an extension as you suggest.

PR #236, opened 2016-07-30
Single handler to WithMessageHandler

AnthonySteele:
One handler at a time: You can only call `WithMessageHandler` once for a given queue and message type.
This is intended to make it harder to accidentally put multiple handlers on the same queue.
You use a new extension `WithMessageHandlers` if you still want to do this, and it's clear that you have 1 queue definition, and multiple handlers together on the end of it.

And as a consequence, the internals in `HandlerMap` and `MessageDispatcher` become simpler for the usual case, the support for multiple handlers moves out into `ListHandler`

What do you think?
Needs more tests?


slang25:
From my limited understanding, this looks a great improvement 👍. I'll sense check this with @payman81 tomorrow, and we could have commit 1000! 💃

payman81:
LGTM

PR #238, opened 2016-07-31
Make throttle processing strategy more robust

slang25:
Unfortunately this change isn't compatible with changing max workers at runtime. This is something that could be reintroduced, I think it would be a cleaner to have it pushed based, not a Func you pass in, otherwise it's going to get complicated.

### DIFF
--- a/JustSaying.Messaging/MessageProcessingStrategies/Throttled.cs
+++ b/JustSaying.Messaging/MessageProcessingStrategies/Throttled.cs
@@ -1,6 +1,6 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
+using System.Diagnostics;
+using System.Threading;
 using System.Threading.Tasks;
 using JustSaying.Messaging.Monitoring;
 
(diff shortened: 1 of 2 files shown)


AnthonySteele:
I have not seen any case where we actually change the `maxWorkers` at runtime.

AnthonySteele:
"A lightweight semaphore controls access to a pool of resources that is local to your application. When you instantiate a semaphore, you can specify the maximum number of threads that can enter the semaphore concurrently." - https://msdn.microsoft.com/en-us/library/system.threading.semaphoreslim(v=vs.110).aspx

This seems to be the right approach. You have piloted this code?

JustSaying.Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -1,109 +1,75 @@
...
+            MaxWorkers = maxWorkers;
+            _semaphore = new SemaphoreSlim(maxWorkers, maxWorkers);
         }
 
-        public int MaxWorkers
+        public void StartWorker(Func<Task> action)
         {
-            get { return _maxWorkersProducer(); }
+            var messageProcessingTask = new Task<Task>(() => ReleaseOnCompleted(action));
+            _semaphore.Wait();
```
AnthonySteele: Is it possible to put this `_semaphore.Wait()` inside `ReleaseOnCompleted`, i.e  just before line 32. This would also allow using `WaitAsync` .

slang25:
We have, in 2 components, with the throttle code hit very hard. We've had no issues for a week, whereas with the previous ```Throttle``` implementation we hit the problem scenario twice in 2 days as soon as we hit load.
I could spot a few issues with the code that calls the old ```Throttle``` implementation, one that meant we cannot actually control how much work is added, so we over add the task array. But the real issue seemed to be that the ```.ContinueWith(..., TaskContinuationOptions.ExecuteSynchronously``` task doesn't get called, and I'll put my hands up, I don't understand why.
This version will block if we try adding too many, and more importantly reliably releases the count.

JustSaying.Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -1,109 +1,75 @@
...
+                timeout: TimeSpan.FromMilliseconds(int.MaxValue),
+                executeOnlyOnce: true);
 
-        private void MarkTaskAsActive(Task t)
-        {
-            lock (_activeTasks)
-            {
-                _activeTasks.Add(t);
-            }
+            return tcs.Task;
```
AnthonySteele: Ack, this looks a bit of of complexity, but if it's the only way to wait for a completion without entering the semaphore? 
Would  `_semaphore.WaitAsync(); _sempahore.Release()` be even more of a hack?

JustSaying.Messaging/MessageProcessingStrategies/Throttled.cs
slang25: We could, but if we got here and ```_semaphore.Wait()``` blocks then we have entered a very edge case race condition, and I think we should block the loop momentarily which is what should have happened back when we wait for workers (but multiple threads will get released a the same time, this is the problem)

AnthonySteele: Good point, the `SqsNotificationListener` should have called `WaitForAvailableWorkers` just before this so it's an edge case to actually wait.

JustSaying.Messaging/MessageProcessingStrategies/Throttled.cs
slang25: Yeah, as you say the alternative is to wait (sync or async) then release. This is more semantically correct, but uglier.

AnthonySteele:
Ok, I think we should merge today, productionise next week and look at the smaller issues after.

PR #239, opened 2016-08-01
Fixes #189 - Expose SQS message's ReceiptHandle to allow changing message visibility

### LINKED ISSUE #189: Expose SQS message's ReceiptHandle
I have a use case where it would be beneficial to call [ChangeMessageVisibility](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ChangeMessageVisibility.html).

Seems like a big change to provide full support for this use case, an interim solution might be to expose an SQS message's ReceiptHandle on the Just Saying Message type so that the AWS SDK can at least be called directly?

ssemyonov:


AnthonySteele:
Do we want just this field, or do we want a fix with full access to the message metadata?

payman81:
@AnthonySteele We can add the fix now and phase it out when Message metadata aka Envelope is added later.

AnthonySteele:
If you think that's worthwhile, sure. I don't know if/when envelopes will be worked on, so if this is all we get in 5.0 it's worthwhile.

payman81:
@AnthonySteele Since this PR is not a breaking change it can be added to existing 4.x. We could still add message Metadata to version 5.0. right?

slang25:
Yeah, so if @ssemyonov creates a PR for this against the master branch we can get it merged and get a release out, we can cherry pick this later if we don't get around to message envelopes (which I think we will get done) in v5.0

AnthonySteele:
You are both correct, that's a good plan.

payman81:
@ssemyonov Can you please submit your PR against master?

ssemyonov:
Will submit PR to master

PR #242, opened 2016-08-02
#189 - Exposing queue URL as well not to implement queue discovery in the consuming code

ssemyonov:
Originally I was thinking it would be easier to figure out queue URL in the consuming code, however it is not aware of the AWS region and the default naming strategy.
This adds a bit more metadata to the ```Message``` class that should go to the envelope in v5.


slang25:
I've retriggered the build, can't see how your changes would fail the tests.

ssemyonov:
@slang25 I've update PR to fix the tests

ssemyonov:
The tests were failing because of ```UriFormatException``` but it looked like a timeout in the test report.

PR #249, opened 2016-08-13
DNM: stopwatch handler - metric name issue

AnthonySteele:
# Do not merge

The failing test demonstrates an issue: The `StopwatchHandler` assumes that it came get the name of the inner handler and send it in the metrics.  But the rest of the code assumes that we can safely wrap handlers at will, multiple times for different purposes. And we keep introducing more of them: recently there is `BlockingHandler` and `ListHandler`.

These assumptions don't play well together, so we see metrics in statds with names like: `stats.timers.myapp.handlers.orderaccepted.futurehandler1.mean` or `stats.timers.myapp.handlers.paid.stopwatchhandler1.upper_90`

The handler name here is just noise. Is there a good way to fix that, or do we have to remove it?

### DIFF
--- a/JustSaying.Messaging.UnitTests/Monitoring/StopwatchHandlerTests.cs
+++ b/JustSaying.Messaging.UnitTests/Monitoring/StopwatchHandlerTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using JustSaying.Messaging.MessageHandling;
 using JustSaying.Messaging.Monitoring;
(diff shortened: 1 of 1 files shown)


payman81:
I thought it would be valuable to provide metrics for handlers in a consistent way so I attempted to improve StopwatchHandler and I failed.
This https://goo.gl/zM5fhS would've worked before v4.x but now with so many different types of decorators including FutureHandler wrapping ````Func<IHandler<T>>```` it is increasingly difficult to solve this problem elegantly. 
Also, timing metrics for ````ListHandler<TMessage>```` wrapping IEnumerable<IHandler<>> is kind of pointless. It is more useful to time individual handler's execution and for that reason I gave up.

AnthonySteele:
These metrics are indeed valuable, and would be better if we can sort out the naming. I think `FutureHandler` has been there since before I looked at the V3 code, so these naming issues go back a bit.

PR #253, opened 2016-08-16
IHandlerResolver should resolve only one handler

AnthonySteele:
`IHandlerResolver` is expected to return exactly 1 handler, so put that on the contract.

i.e.
````csharp
public interface IHandlerResolver
 {		   
  IHandlerAsync<T> ResolveHandler<T>();
 }
````

Sorry about the merge commits, is it possible to flatten this when merging?


slang25:
I've been a bit slow on this one. This is clearly more sane given the current assertions we make in the code.

PR #257, opened 2016-08-20
Handler resolution context

AnthonySteele:
I would like Sebastien and others to have a look, as this related to a problem that Sebastien encountered.

When resolving a handler, provide a context so that different handlers can be resolved in different contexts if needed. The problem that we are addressing is that when we have two queues on the same topic, we want a different handler for each queue and this doesn't work well with the `IHandlerResolver`.

e.g.

````csharp
_fluentNotificationStack
    .WithSqsTopicSubscriber()
        .IntoQueue("QueueA")
        .WithMessageHandler<OrderPlaced>(_resolver)

    .WithSqsTopicSubscriber()
        .IntoQueue("QueueB")
        .WithMessageHandler<OrderPlaced>(_resolver)
````

Then we would like the resolver to resolve an instance of `OrderPlacedHandlerA` in the first case and an instance of `OrderPlacedHandlerB` in the second case. 

This change will give the handler resolver information that can be used to do this using IoC container mechanisms such as [named bindings](https://github.com/ninject/Ninject/wiki/Contextual-Binding#simple-constrained-resolution-named-bindings)

I'm not thrilled by the class name `HandlerResolutionContext` - it's long and the word "Context" is general enough to mean anything. But it is best to have a separate object for this task as it can be extended as been be without breaking anything else. Right now it looks like we only need the (short) queue name.

If the design is good, it probably needs a test demonstrating how to hook this into an IoC container and resolve different handlers?

### DIFF
--- a/JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
+++ b/JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
@@ -115,10 +115,12 @@
     <Compile Include="WhenRegisteringASqsSubscriber\WhenRegisteringLongNameMessageTypeTopicSubscriber.cs" />
     <Compile Include="WhenRegisteringHandlersViaResolver\GivenAPublisher.cs" />
     <Compile Include="WhenRegisteringHandlersViaResolver\MultipleHandlerRegistry.cs" />
+    <Compile Include="WhenRegisteringHandlersViaResolver\NamedHandlerResolverTests.cs" />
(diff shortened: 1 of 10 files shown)


slang25:
👍 @AnthonySteele this is what I had in mind while we were discussing this. @serialseb what are your thoughts?

AnthonySteele:
I am reasonably happy with this PR. Any thoughts on a better name than `HandlerResolutionContext`?

PR #259, opened 2016-08-23
DNM: Make message publishing async

AnthonySteele:
**Do not merge** we are rethinking this.

Publishing a message is awaitable
Review the retry code.
Also add more details when publishing fails altogether (it does sometimes happen).

What is the policy on `throw new Exception` ? Should I throw this, some other existing type, or declare a new exception type for the purpose?

### DIFF
--- a/JustSaying.AwsTools/MessageHandling/SnsTopicBase.cs
+++ b/JustSaying.AwsTools/MessageHandling/SnsTopicBase.cs
@@ -1,4 +1,5 @@
 using System.Linq;
+using System.Threading.Tasks;
 using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
 using Amazon.SQS;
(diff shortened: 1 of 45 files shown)


JustSaying.AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -44,19 +46,30 @@ public bool Subscribe(IAmazonSQS amazonSqsClient, SqsQueueBase queue)
             return false;
         }
 
-        public void Publish(Message message)
+        public async Task Publish(Message message)
```
stuart-lang: ```PublishAsync```? or provide both methods?

JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -19,13 +21,23 @@ public SqsPublisher(RegionEndpoint region, string queueName, IAmazonSQS client,
...
+            {
+                await _client.SendMessageAsync(request)
+                    .ConfigureAwait(false);
+            }
+            catch (Exception ex)
             {
-                MessageBody = GetMessageInContext(message),
-                QueueUrl = Url
-            });
+                throw new Exception($"Failed to publish message to SQS. QueueUrl: {request.QueueUrl} MessageBody: {request.MessageBody}", ex);
```
stuart-lang: This is good cause it shields the complexity from the consuming code (JustSaying is an abstraction), and wrapping keeps the information (https://msdn.microsoft.com/en-us/library/ms229049(v=vs.100).aspx)

It would be nice to have our own exception, even as generic as ```PublishException``` and have it in the method xml summary, so others devs will see it as they use it, if they see just ```Exception```, they may be hesitant to catch that as they may catch general exceptions local arround that call. That can come later though.

JustSaying.AwsTools/MessageHandling/SnsTopicBase.cs
AnthonySteele: Maybe we should provide both, in which case it would be `Publish and `PublishAsync`. But that is  a larger PR, so not today.

PR #260, opened 2016-08-23
An exception for when publishing fails

AnthonySteele:
Include the relevant message and destination data when throwing an exception that indicates that `Publish` failed
NB we retry publishing (3 times by default), so the exception seen outside of JustSaying will be from the last attempt. Previous exceptions will be discarded.

### DIFF
--- a/JustSaying/JustSayingBus.cs
+++ b/JustSaying/JustSayingBus.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using JustSaying.Extensions;
@@ -116,7 +117,10 @@ namespace JustSaying
             lock (_syncRoot)
             {
                 if (Listening)
+                {
                     return;
+                }
+
                 foreach (var regionSubscriber in _subscribersByRegionAndQueue)
                 {
                     foreach (var queueSubscriber in regionSubscriber.Value)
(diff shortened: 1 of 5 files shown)


JustSaying/JustSayingBus.cs
```diff
@@ -200,23 +205,18 @@ private void Publish(IMessagePublisher publisher, Message message, int attemptCo
...
                     Monitor.IssuePublishingMessage();
-
-                    var errorMessage = "Unable to publish message " + message.GetType().Name;
-                    Log.Error(ex, errorMessage);
+                    Log.Error(ex, $"Unable to publish message {message.GetType().Name} after {attemptCount} attempts");
                     throw;
                 }
 
-                Thread.Sleep(Config.PublishFailureBackoffMilliseconds * attemptCount); // ToDo: Increase back off each time (linear)
+                Thread.Sleep(Config.PublishFailureBackoffMilliseconds * attemptCount);
```
AnthonySteele: Maybe we should log the exception that we are discarding and that we are retrying here?

slang25: Yeah, no harm it in. It would be nice to be aware of transient connectivity issues, at the moment there's no knowing if it's working every-time or not.

AnthonySteele: I have logged the exception as a warning. This is not common but i think it's appropriate here when it will be retried.

PR #263, opened 2016-08-24
.Net core

AnthonySteele:
Code with Stuart Lang and Joseph Woodward

- Dependencies and metadata in project.json are not fine-tuned
- We use beta packages of AWS SDK and others for .Net core support
- For lots of ADS SDK methods we worked around only having the async version
- the tests are all .Net 4.6, not core. 

But it compiles. Lets get to a point where the test pass and get this merged, keeping other things out of scope: tests on .Net core, and more async.


slang25:
Looks great. I've tweak some of the project.jsons here: https://github.com/slang25/JustSaying/tree/AnthonySteele-netcore
And added a build script which runs the tests in AppVeyor, I'll work this evening on getting building the nuget package correctly. I want this to run locally the same as it does on AppVeyor, and to simplify and automate the release process.

AnthonySteele:
That's great! Should we switch the PR to your branch?
State of play for me now is that although the tests don't run on appveyor, I can get them to Pass On My Machine with e.g. 
````
cd JustSaying.UnitTest
dotnet test
````

And I have fixed the unit test failures (mostly around setting up different expectations in the mocks).
`JustSaying.AwsTools.IntegrationTests` also passes, but the test in`JustSaying.IntegrationTests` hang and I haven't had time to look into it. 

That should not prevent an Alpha package.

**Comment by AnthonySteele** [comment]
I asked appveyor about how to get this running. It is possible: "disable tests auto-discovery and use your own `dotnet test ...` command under test_script section." which I think is what your `build.cake` script basically does?

**But NB** changes to appveyor settings will apply to master and to other PRs, not just to this PR. It seems that we can't build both ways at the same time?

**Comment by slang25** [comment]
@AnthonySteele That is essentially what I'm doing yes. But I think it's a good thing anyway, we can run this from anywhere and have it work the same, and it becomes easy to automate deployment.

I would have thought AppVeyor would use the default branch for PRs or the target branch, cause AppVeyor has a setting for default branch (not to be confused with the GitHub default branch).
You can do branch specific stuff apparently:
https://www.appveyor.com/blog/2014/07/23/appveyor-yml-and-multiple-branches/
I think we will struggle to get a nupkg that contains all the dlls by the way, it looks like it's not the way in the new world, they want you to build a package per project, and packages do not contain dependencies. I am working on it though.

File: JustSaying.AwsTools/MessageHandling/SqsQueueByNameBase.cs [hunk 1]
```diff
@@ -74,7 +74,7 @@ public bool Create(SqsBasicConfiguration queueConfig, int attempt = 0)
                     Log.Info(
                         $"Waiting to create Queue due to AWS time restriction - Queue: {QueueName}, AttemptCount: {attempt + 1}");
                     await Task.Delay(60000);
-                    await CreateAsync(queueConfig, attempt: attempt++);
+                    await CreateAsync(queueConfig, attempt + 1);
```
slang25: nice spot

AnthonySteele:
I am a little concerned that with multiple long-lived branches we run the risk that they become increasingly  incompatible and harder to merge over time.  Are we going to try to get this onto the `develop` branch?

slang25:
Yes, fair point. Are you happy with it in it's current state? Great work btw

AnthonySteele:
It seems to be in a good state - propagating `async` through the internals, and providing sync wrappers are the edges seems to have worked well.

My only concern is that without a CI build, we could miss things that are breaking. But if that's only a short-term issue then we can go ahead.

slang25:
💥 Merged it. I'll follow up with a cake build PR and work on getting a beta release out in the next week or so.

martincostello:
Nice work guys!

AnthonySteele:
:boom: :+1:

pofallon:
This is great!  Any word on when this will be published to NuGet?  Thanks!

PR #266, opened 2016-09-14
Added 'Configuring the client' section.

JefClaes:


slang25:
Thanks @JefClaes 👍You can also leave this step out and it will use the awssdk default profile, but that's not documented either.

PR #267, opened 2016-09-15
ExactlyOnce timeout should default to max.

payman81:
Removing incorrect cast from double to int that is causing integer overflow.


slang25:
Good spot 👍 just a failing test to fix

PR #269, opened 2016-09-23
When moving messaging between queues, make sure the queue exists.

JefClaes:
In the current version, moving messages would throw an exceptions since the Url property of the queues is never initialized. By checking whether the queue exists, the Url property is initialized.


JefClaes:
Messed up.

PR #271, opened 2016-09-23
Ensure queue exists when moving messages.

JefClaes:
If we don't invoke this propety the QueueUrl property is never set.


slang25:
Thanks @JefClaes 😄

PR #273, opened 2016-10-02
Added timeout to receive message call

slang25:


slang25:
I probably should some tests around ```ListenLoop``` and inject the timeout CancellationToken.
I have "Allow edits from maintainers" enabled, so for trivial changes, go nuts!

AnthonySteele:
👍 I'm having a look now. If you go to our Kibana and look for "Issue in message handling loop for queue" over the last few days you should see some actual data. The exceptions seem to be `Amazon.Runtime.AmazonServiceException` and `	System.OperationCanceledException` so I don't think that we can count on the exception type.

slang25:
Good points and good changes.

I was hesitant of relying on the Exception type, I expected in this scenario that we would either get ```TaskCancelledException``` or ```OperationCancelledException``` depending on where inside ```ReceiveMessageAsync``` it's thrown from, ```TaskCacnelledException``` inherits from ```OperationCancelledException``` anyway so we could just catch that, but I wasn't confident and the docs give you no info around exceptions.

slang25:
Random observation, do we need to normalize line endings? or add a .gitattributes? Looks like your commits changed every line, had to do a ```?w=1``` to view the changes

AnthonySteele:
Sorry about that! Is it possible to revert that in another commit, and squash-merge after?

We have an `.editorconfig` file but no `.gitattributes` in this repo. Add one if you like.

PR #277, opened 2016-10-09
Fix memory leak introduced by Linked CTS

slang25:
Fixes leak that was introduced by not disposing the CTS returned from ```CreateLinkedTokenSource```. Thanks to those that spotted this leak so quickly.


AnthonySteele:
👍 Looks good to me.

PR #279, opened 2016-10-15
Revert fix to "ExactlyOnce"

AnthonySteele:
revert https://github.com/justeat/JustSaying/commit/15635c3e8961371498f1c84f94eec4af8c78bdf3
It's not great to revert a bug fix.
But
This turned out to not be a trivial change, it changes the behaviour of an important live system, which has been working fine for more than a year with existing code.
 
Unfortunately we also need changes to `SqsNotificationListener` which were applied *after* https://github.com/justeat/JustSaying/commit/15635c3e8961371498f1c84f94eec4af8c78bdf3, for reliability of connection to queues. And we cannot apply it.

So we are stuck. One way to get us moving again is to
 * back out of this change to `ExactlyOnce` 
 * build and package with all the other fixes. We can use this package
 * re-apply this change in a safe manner

### DIFF
--- a/JustSaying.AwsTools.UnitTests/MessageHandling/HandlerMetadataTest.cs
+++ b/JustSaying.AwsTools.UnitTests/MessageHandling/HandlerMetadataTest.cs
@@ -35,7 +35,7 @@ namespace JustSaying.AwsTools.UnitTests.MessageHandling
             Assert.That(reader.GetTimeOut(), Is.EqualTo(42));
         }
 
-        [Test]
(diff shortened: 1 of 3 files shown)


slang25:
@AnthonySteele Good summary. Unless there are any objections, I'll merge this later today. Maybe we change this in v5, and have the timeout take a ```TimeSpan``` which is a good idea anyway, and will break binary compatibility with our internal code that isn't synchronized with the new behavior.

PR #280, opened 2016-10-16
WIP cake build script DNMY

stuart-lang:
Still a WIP, just to get some early feedback.

I'd like to introduce cakebuild to this project, so that we are less coupled to AppVeyor, using cake we can run the exact same build script locally as we do on appveyor, and we can change our appveyor.yml to be "dumb" and just call the build script.

Current state of this is that it will clean, build, run tests (broken atm), and build the new mutli-framework nuget packages.

One big change with the nuget packages is that there is 1 package per project, we can work around this, but it is not pretty, and we'll be fighting the eco-system.

I'd like to tidy this PR up, and put in the release step, include GitHub release notes, automatic versioning using AppVeyor build numbers, and GitLink as a replacement for symbols for source stepping.


AnthonySteele:
I will have a closer look at this later, and I'm no expert on cake build. But I think that this is a good direction to take at present. Getting this branch and its test working on appyeyor and the same script running locally is what we need.

stuart-lang:
A great resource here, pretty much what I am intending on implementing:
http://www.michael-whelan.net/continuous-delivery-github-cake-gittools-appveyor/

PR #282, opened 2016-11-10
Update aws sdk to 2.3.55.2

AnthonySteele:
This is the latest release version and contains a fix to https://github.com/aws/aws-sdk-net/issues/152
Which is likely the case of our "hangs when reading from queues" issue
We have a workaround, but the internal fix is IMHO better.

The question is, do we still need the "linkedCancellationToken" measure that fixes it? 
If we view it as a fix to a very specific bug that we are sure is fixed then no.
If we view it as a general "belt and braces" reliability measure then yes.


slang25:
I think the fix is a good "belt and braces" as you say, we could test it and take it out if we are happy the issue is gone, I'm always in favor of simpler code.
I think we should merge this regardless, and then PR a rollback of the linked token, test it in QA (I'm happy to do this next week)

PR #286, opened 2016-12-02
Longer polling timeout

slang25:
@JonahAcquah has done some investigation, and it appears that we hit the 30 second timeout very frequently, from some quick analysis I would estimate 25% of polls during quiet periods exceed this duration resulting in errors being logs, this invalidates my original assertion that 30 seconds is more than adequate.
Included #285 @JonahAcquah I think this is the branch you wanted it in.

### DIFF
--- a/JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying.AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -166,7 +166,7 @@ namespace JustSaying.AwsTools.MessageHandling
                         WaitTimeSeconds = 20
                     };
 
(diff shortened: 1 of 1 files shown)


JonahAcquah:
Thanks @stuart-lang

AnthonySteele:
Right now it should go on both `master` and `develop` branches :/

JonahAcquah:
@stuart-lang, the argument names don't look right, I've updated them here https://github.com/JonahAcquah/JustSaying/commit/4a2f4227a2343b5f027c3715885145f95dceb449.

slang25:
Rolled out @JonahAcquah change for him to submit in a separate PR

slang25:
@AnthonySteele following our internal discussion, are you happy to merge this?

PR #297, opened 2017-01-23
Add logging abstraction

slang25:
This is using `Microsoft.Extensions.Logging.Abstractions`, we could change this for our own interface, however I think the `ILogger`, `ILoggerFactory` & `ILoggerProvider` pattern is really nice, [this post covers it quite well](https://wildermuth.com/2016/04/22/Implementing-an-ASP-NET-Core-RC1-Logging-Provider), and we would just be reinventing the wheel.

This introduces a breaking change to the fluent api, now:
```
CreateMeABus.InRegion(...).With...
```
becomes
```
CreateMeABus.WithLogging(someLoggerFactory).InRegion(...).With...
```

If not logging is required then the consuming code should do something like:
```
var emptyLoggerFactory = new LoggerFactory();
CreateMeABus.WithLogging(emptyLoggerFactory).InRegion(...).With...
```

There are many providers out there, [here is what we'd use for NLog](https://github.com/NLog/NLog.Extensions.Logging)


JustSaying/IHaveFulfilledLoggingRequirements.cs
```diff
@@ -0,0 +1,9 @@
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying
+{
+    public class JustSayingFleuntlyLogging
```
AnthonySteele: typo

JustSaying/CreateMeABus.cs
```diff
@@ -13,33 +12,9 @@ public static class CreateMeABus
         /// </summary>
         public static Func<IAwsClientFactory> DefaultClientFactory = () => new DefaultAwsClientFactory();
 
-        public static IMayWantOptionalSettings InRegion(string region) => InRegions(region);
+        public static JustSayingFleuntlyLogging WithLogging(ILoggerFactory loggerFactory) =>
```
AnthonySteele: Do a find replace "Fleuntly" to "Fluently"

JustSaying.Logging/LoggerExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.Logging
+{
+    public static class LoggerExtensions
```
AnthonySteele: Longer term we could inline these methods, as it's just a shim for existing methods to ILogger ?

AnthonySteele:
LGTM 👍 What do the expected on .Net core logging think?

JustSaying.Logging/LoggerExtensions.cs
slang25: Actually, I think I will inline this, cause I'm throwing out the rest of this project, this was just here to ease the migration. I'll do it as my last commit of this review to reduce the noise.

slang25:
Comments addressed 👍

JustSaying.Logging/LoggerExtensions.cs
AnthonySteele: 👍

JustSaying/project.json
```diff
@@ -37,10 +37,10 @@
     "AWSSDK.SimpleNotificationService": "3.3.0.6",
     "AWSSDK.SQS": "3.3.1.6",
     "JustSaying.AwsTools": "5.0.0",
+    "JustSaying.Logging": "5.0.0",
```
martincostello: Looks like the AppVeyor build fails trying to restore this reference.

JustSaying/JustSayingFluentlyExtensions.cs
```diff
@@ -1,11 +1,44 @@
...
+using JustSaying.Messaging.Monitoring;
 using JustSaying.Models;
 
 namespace JustSaying
 {
     public static class JustSayingFluentlyExtensions
     {
+        public static IMayWantOptionalSettings InRegion(this JustSayingFluentlyLogging logging, string region) => InRegions(logging, region);
+
+        public static IMayWantOptionalSettings InRegions(this JustSayingFluentlyLogging logging, params string[] regions)
```
martincostello: Is it worth adding an overload that also takes ```IEnumerable<string>``` for the regions?

JustSaying/project.json
martincostello: Looks like this is broken in master too actually...

slang25: Yeah, this is broken, but you are right the build is broken atm anyway. Good spot

JustSaying/JustSayingFluentlyExtensions.cs
slang25: We could do, I'm not sure it adds much though, I presume it's done like this to enable `...InRegions("uk","fr")` as well as `...InRegions(tenants.ToArray())`.
Yeah, let me do that unless anyone disagrees.

JustSaying/JustSayingFluentlyExtensions.cs
```diff
@@ -12,6 +14,8 @@ public static class JustSayingFluentlyExtensions
     {
         public static IMayWantOptionalSettings InRegion(this JustSayingFluentlyLogging logging, string region) => InRegions(logging, region);
 
+        public static IMayWantOptionalSettings InRegions(this JustSayingFluentlyLogging logging, IEnumerable<string> regions) => InRegions(logging, regions.ToArray());
```
martincostello: Does it work OK with overload resolution if the implementation is in ```IEnumerable<string>``` and the ```string[]``` overload calls it? Then we don't need the ```ToArray()``` call.

martincostello: I suppose that would just create a stack overflow without a cast, but a cast would be more lightweight than a copy?

martincostello:
I just noticed there's a section about logging in the README. Should we update it in this PR or do that separately?

martincostello:
The CI looks like it's hanging now having issues loading NUnit:

[```Could not load file or assembly 'nunit.framework' or one of its dependencies. The system cannot find the file specified.```](https://ci.appveyor.com/project/justeattech/justsaying/build/4.0.226#L493)

JustSaying/JustSayingFluentlyExtensions.cs
slang25: Yeah, so arrays will pick the `string[]` overload, things like `List<string>` will use the other overload, all works as expected. Unless you mean reversing the overloads, which makes more sense, we'd need to use `.AsEnumerable()` to not create an infinite recursive call.

martincostello: Yeah I meant reversing and using a cast/```as``` so a copy isn't needed:

```csharp
public static IMayWantOptionalSettings InRegion(this JustSayingFluentlyLogging logging, params string[] regions) => InRegions(logging, regions as IEnumerable<string>);

public static IMayWantOptionalSettings InRegions(this JustSayingFluentlyLogging logging, IEnumerable<string> regions)
{
    // Stuff
}
```

martincostello:
Something's still making it hang :/

slang25:
Yeah, it's AppVeyor selecting NUnit 3.5 automatically, we have to change it to use a script. If it works, I'll revert it and add an `appveyor.yml` to this branch.

slang25:
Wait, I'm being dumb. It is using an `appveyor.yml` but it might be using the one from master.

martincostello:
I thought it always used the one from the branch it builds.

appveyor.yml
```diff
@@ -1,41 +1,44 @@
...
-  on_build_failure: true
-  on_build_status_changed: true
-deploy:
-- provider: NuGet
-  api_key:
-    secure: 6MzbzEs4YdJKS67Gio5gEO8mNKmwfC4UHTCmECZ1KOutI6ndm4vAECazmVNB6an7
-  artifact: /.*nupkg/
-  on:
-    APPVEYOR_REPO_TAG: true
+version: 4.0.{build}
```
AnthonySteele: what's happening with the `^M` chars here?

slang25: Not sure, that wasn't intentional. I'll try to fix it

martincostello:
Looks good now apart from the ```^M```'s.

slang25:
Thanks. I'm just updating the README file to reflect the changes.

JustSaying.IntegrationTests/WhenRegisteringHandlersViaResolver/WhenRegisteringAHandlerViaContainerWithMissingRegistration.cs
```diff
@@ -1,34 +1,35 @@
...
+        protected override void Given()
+        {
+            RecordAnyExceptionsThrown();
+        }
+
+        protected override Task When()
+        {
+            var handlerResolver = new StructureMapHandlerResolver(new Container());
+
+            CreateMeABus.WithLogging(new LoggerFactory()).InRegion("eu-west-1")
```
AnthonySteele: Inconsistent formatting, use one "WithXXX" per line

AnthonySteele:
merge it!

PR #299, opened 2017-01-31
Increase polling timeout to 5 minutes

slang25:


slang25:
Approved by @AnthonySteele

fgiele:
I'm curious why this was increased to 5 minutes. 
We're experiencing 'stuck' receivemessage situations ourselves when using the AWS SDK directly. Have you found any root cause behind this or is the timeout a workaround?

slang25:
@fgiele We had run into a known issue with AWS SDK, which was fixed by AWS roughly a week after this fix was merged, basically when the network was interrupted (for us, when  our NAT boxes were restarted) then the SDK task would never complete.

We fixed this by adding a timeout, even though this in theory is fixed in the newer SDK, we like the safety for this fallback. We went for 5 minutes, cause it turns out we had a small percentage (but significate due to the volume we are doing) or polls that took longer than the 25 seconds we originally allowed for, in theory the polls shouldn't take longer than 20 seconds, they almost always finished in 19.9xxx seconds in practice.

PR #300, opened 2017-01-31
Create nuget packages

slang25:


AnthonySteele:
The `^M` line ending is back :(

slang25:
Weird, I'm using VSCode with the editorconfig plugin. I will make some changes. Could be a combination of the `.gitattributes` and `.editorconfig` or it may be the line endings in this repo aren't quite right at the moment. Will investigate later.

AnthonySteele:
It's possible that `.gitattributes` and `.editorconfig` are fighting.

slang25:
Not yet, still experimenting, having an open PR is the easiest way. Let me tag with WIP

JustSaying.AwsTools.IntegrationTests/project.json
```diff
@@ -1,57 +1,57 @@
...
-  },
-  "testRunner": "nunit",
-  "frameworks": {
-    "net46": {
-      "dependencies": {}
-    }
-  }
-}
+{
+  "version": "5.0.*-beta01",
```
martincostello: I'd suggest using a bigger zero-pad (```0001```) just in case we get in a scenario where we just use the auto-generated build numbers and then it gets into the hundreds/thousands.

AnthonySteele: I don't think that we're generating the beta numbers, juct manually counting. I really hope that there won't be 99 of them!

martincostello: Fair enough if it's always manual then. I use the AppVeyor build number so I can auto-publish/consume from the AppVeyor feed on push so each build gets a unique version.

slang25:
@martincostello Do you have any suggestions on how to easily set the `project.json` version number? I'm googling now, but not sure I can do what I want with `dotnet build --version-suffix ...`

martincostello:
@slang25 I do something like [this](https://github.com/martincostello/browserstack-automate/blob/master/Build.ps1#L23-L42) with one of my packages and then push that in with ```--version-suffix``` ([here](https://github.com/martincostello/browserstack-automate/blob/master/Build.ps1#L63) and [here](https://github.com/martincostello/browserstack-automate/blob/master/Build.ps1#L81)).

slang25:
Nice @martincostello 👍
@AnthonySteele what shall we do for this beta? Do we want the version numbers: 5.0.123-beta01
where 123 is the AppVeyor incrementing build number? If so then I think we'll need to do something to rewrite the `project.json` files

AnthonySteele:
I don't really mind. If it's easier, then go to generated 3-digit beta numbers, e,g ` 5.0.0-beta123`

stuart-lang:
This would be inconsistent with [previous versioning](https://www.nuget.org/packages/JustSaying/) and it doesn't change the assembly version info (I think), not sure if we care about this?

AnthonySteele:
What do you recommend?

martincostello:
Wouldn't using the assembly info patching feature of AppVeyor fix the major.minor.revision numbers?

slang25:
Ah, forgot that that would still work 😀

slang25:
Ok, I think I am happy with this now. @AnthonySteele @martincostello please review.

slang25:
Although, after doing some digging it looks like the general convention people are using is `5.0.0-beta1`, and we should change from incrementing the build number only on package releases. This is an easy change to make.

appveyor.yml
```diff
@@ -1,26 +1,44 @@
...
+      ConvertTo-Json $project -Depth 100 | Out-File -Encoding UTF8 $projectFile
+    }
+- ps: dotnet restore
+- ps: dotnet build JustSaying/project.json
 test_script:
   - dotnet test .\JustSaying.AwsTools.UnitTests\project.json
   - dotnet test .\JustSaying.Messaging.UnitTests\project.json
   - dotnet test .\JustSaying.UnitTests\project.json
+after_build:
+- dotnet pack "JustSaying" -o artifacts --no-build
```
martincostello: I suggest we use ```--include-symbols``` to make a symbols package with the PDBs in as well.

martincostello:
LGTM other than my suggestion to create symbols packages.

appveyor.yml
martincostello: My bad, it's doing it anyway.

PR #308, opened 2017-03-02
SQS Message Timers

### LINKED ISSUE #22: Implement delayed publishing to a topic
Sometimes we want to delay the the publishing of a notification to a topic for extended periods of time.  SNS doesn't support this out of the box.
# Example Scenarios when this could be used
1. When a user registers with my website, I send them a confirmation email with a confirmation link in the email. If the user doesn't respond in 7 days I want to notify the rest of the system that they haven't responded.so that a reminder email can be sent to the user.  I would do this by publishing a notification with a delay of 7 days (i.e. it would not be published to subscribers until exactly 7 days had passed).  If the user does click on the confirmation link within 7 days then I need to be able to delete the notification so it does not get published to subscribers.
2. On my shopping website, when a user adds items to their basket I need to reserve the stock for an hour.  After an hour I need to be able to release the stock. I would do this by publishing a notification with a delay of 1 hour.  If the user does not checkout within an hour then the notification is sent to subscribers so my reserved stock is released.  If the user completes their order within an hour then I need to be able to delete the notification so it is never published to subscribers.

vorotech:
Implementing SQS message timers.
http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-timers.html
Partially fixes #22 , because at the moment message can be delayed up to 15 mins when _DelaySecconds_ message property set (default is 0).

### DIFF
--- a/JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
+++ b/JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
@@ -24,7 +24,8 @@ namespace JustSaying.AwsTools.MessageHandling
             _client.SendMessage(new SendMessageRequest
             {
                 MessageBody = GetMessageInContext(message),
-                QueueUrl = Url
+                QueueUrl = Url,
+                DelaySeconds = message.DelaySeconds
             });
         }
 
--- a/JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
+++ b/JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
@@ -25,7 +25,7 @@ namespace JustSaying.AwsTools.MessageHandling
             {
                 MessageBody = GetMessageInContext(message),
                 QueueUrl = Url,
-                DelaySeconds = message.DelaySeconds
+                DelaySeconds = message.DelaySeconds ?? 0
             });
         }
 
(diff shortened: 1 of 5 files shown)


JustSaying.Models/Message.cs
```diff
@@ -19,6 +20,7 @@ protected Message()
         public string Conversation { get; set; }
         public string ReceiptHandle { get; set; }
         public string QueueUrl { get; set; }
+        public int DelaySeconds { get; set; }
```
stuart-lang: This is a breaking change for people on older version, maybe make it nullable?

vorotech: Makes sense, Will update PR

JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -24,7 +24,8 @@ public void Publish(Message message)
             _client.SendMessage(new SendMessageRequest
             {
                 MessageBody = GetMessageInContext(message),
-                QueueUrl = Url
+                QueueUrl = Url,
+                DelaySeconds = message.DelaySeconds.GetValueOrDefault()
```
AnthonySteele: `message.DelaySeconds.GetValueOrDefault()` is equivalent to `message.DelaySeconds ?? 0`

vorotech: For the sake of visibility I will change it 👍

vorotech:
Hey @stuart-lang, @AnthonySteele .
Are you happy to get these changes merged? Want to utilize the feature on the project.

PR #309, opened 2017-03-03
SQS Message Timers (#308)

### LINKED ISSUE #22: Implement delayed publishing to a topic

AnthonySteele:
* implementing SQS message timers (might fix #22)

* set delay seconds as optional for backward compatibility

### DIFF
--- a/JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
+++ b/JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
@@ -33,7 +33,8 @@ namespace JustSaying.AwsTools.MessageHandling
             var request = new SendMessageRequest
             {
                 MessageBody = GetMessageInContext(message),
-                QueueUrl = Url
+                QueueUrl = Url,
+                DelaySeconds = message.DelaySeconds ?? 0
             };
 
             try
--- a/JustSaying.TestingFramework/MessageStubs.cs
+++ b/JustSaying.TestingFramework/MessageStubs.cs
@@ -18,6 +18,14 @@ namespace JustSaying.TestingFramework
         public string Content { get; set; }
     }
 
+    public class DelayedMessage : Message
+    {
+        public DelayedMessage(int delaySeconds)
+        {
+            DelaySeconds = delaySeconds;
+        }
+    }
+
     public class MessageWithEnum : Message
     {
         public MessageWithEnum(Values enumVal)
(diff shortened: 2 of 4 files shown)


JustSaying.AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -33,7 +33,8 @@ public void Publish(Message message)
             var request = new SendMessageRequest
             {
                 MessageBody = GetMessageInContext(message),
-                QueueUrl = Url
+                QueueUrl = Url,
+                DelaySeconds = message.DelaySeconds ?? 0
```
jaimalchohan: Is there a Maximum threshold for DelaySeconds?

JustSaying.TestingFramework/MessageStubs.cs
```diff
@@ -18,6 +18,14 @@ public class AnotherGenericMessage : Message
         public string Content { get; set; }
     }
 
+    public class DelayedMessage : Message
+    {
+        public DelayedMessage(int delaySeconds)
+        {
+            DelaySeconds = delaySeconds;
```
jaimalchohan: Is there a Maximum threshold for DelaySeconds?

AnthonySteele: http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html

Valid values: 0 to 900. Maximum: 15 minutes.

PR #312, opened 2017-03-18
Only set DelaySeconds if provided

adammorr:
DelaySeconds on the SendMessageRequest was being set to 0 if no value was provided which would override any Delivery Delay set on the queue.

This change only sets the DelaySeconds if the message to be published has provided a value so that any queue delay is still honoured.


slang25:
Good spot =)

PR #315, opened 2017-04-03
Deprecate sync handlers

AnthonySteele:
The non-async `IHandler` interface is now deprecated in favour of `IHandlerAsync`

### DIFF
--- a/JustSaying.AwsTools.UnitTests/MessageHandling/HandlerMetadataTest.cs
+++ b/JustSaying.AwsTools.UnitTests/MessageHandling/HandlerMetadataTest.cs
@@ -65,7 +65,9 @@ namespace JustSaying.AwsTools.UnitTests.MessageHandling
         [Test]
         public void WrappedHandler_DoesHaveExactlyOnce()
         {
+#pragma warning disable 618
             var wrapped = new BlockingHandler<GenericMessage>(new OnceTestHandler());
+#pragma warning restore 618
 
             var reader = HandlerMetadata.ReadExactlyOnce(wrapped);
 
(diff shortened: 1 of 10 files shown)


martincostello:
Would it be better to migrate off the obsolete handler as well so we don't need all the pragmas which now hide its use?

martincostello:
Hide its _deprecated_ use I should say.

slang25:
@martincostello aren't those handlers also themselves deprecated, so we'd want to propagate the obsolete attribute, which this does is my understanding. The pragmas is for our benefit to say we know we are touching obsolete stuff and we are cool with that.

martincostello:
If that's the case it's cleaner to "duplicate" the ```[Obsolete]``` on the deprecated implementations. The compiler doesn't warn if you use obsolete things from obsolete things. Then you only need to put pragmas in things that test the obsolete things (or workaround that by making the tests obsolete, which is questionable).

stuart-lang:
If that's true then obsoleting the tests also makes sense.

martincostello:
I have mixed opinions on it in the test, yes they're kinda obsolete because they test obsolete things, but if you're still supporting them even though they're obsolete, then you still want to test it, meaning they're not really obsolete. Otherwise you'd just remove the tests completely and if the obsolete stuff got broken you'd not care.

stuart-lang:
What else could an obsolete test mean though? No one's going to think "better stop using this test now, it's obsolete!", it could only mean it's testing obsolete stuff?

martincostello:
I guess, to me, an obsolete test means "a test that isn't needed anymore".

AnthonySteele:
We still need to test `IHandler` and have code that support it - mainly `BlockingHandler` that adapts it to `IHandlerAsync`. Good call that `BlockingHandler` can also be marked as Obsolete.

PR #319, opened 2017-04-12
Put #if around sync Publish

AnthonySteele:
The not-async version of `Publish()` should be
- Not a dangerously blocking wrapper of the async version, but using the underlying sync call on the AWS SDK
- Only available if the underlying sync call on the AWS SDK is available. i.e.  on the full .Net framework, not on .Net core.

Better understanding of async / sync issues due to to [@stuart-lang](http://stackoverflow.com/a/43419278/5599) and [Stephen Cleary](http://stackoverflow.com/a/25096075/5599)

### DIFF
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
@@ -1,0 +1,139 @@
+﻿using System;
+using System.Threading.Tasks;
+using Amazon;
+using Amazon.SQS;
+using Amazon.SQS.Model;
+using JustBehave;
+using JustSaying.AwsTools.MessageHandling;
+using JustSaying.Messaging.MessageProcessingStrategies;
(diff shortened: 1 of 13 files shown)


slang25:
I've fixed up the tests here: https://github.com/slang25/JustSaying/tree/AnthonySteele-if-sync-publish

AnthonySteele:
Yeah, I was lazy about tests, sorry. If this is the right track then we'll need them.

Am I right in thinking that the test suite only runs in .Net full on appveyor right now? But that this might change and we might run the test suite in .Net core on travis like we do with a few other things.

If so, then we need to test both `.Publish` and `.PublishAsync` and we will also need a few `#if` blocks in the tests.

slang25:
Yeah, you are right about that, we may as well do that now and save a bunch of work later.

AnthonySteele:
On the subject of `#If` blocks: these are very easy to get wrong. 
e.g. https://github.com/justeat/JustEat.StatsD/pull/32
e.g. we use `#if NET451` in the code, but in the tests this would have to be `#if NET46` instead.

I did not add any `#If` to the tests as at present they only run on appveyor. It would be good to get them running on travis with .Net core and Linux some time soon in a seperate PR.

PR #321, opened 2017-04-13
Added message backoff strategy on handling failure

adammorr:
Start of introducing a backoff strategy if message processing fails tbc...

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -93,19 +93,7 @@ namespace JustSaying.AwsTools.MessageHandling
             {
                 if (!handlingSucceeded && _messageBackoffStrategy != null)
                 {
-                    if (message.Attributes.ContainsKey(MessageSystemAttributeName.ApproximateReceiveCount) && int.TryParse(message.Attributes[MessageSystemAttributeName.ApproximateReceiveCount], out int approximateReceiveCount))
-                    {
-                        var visibilityTimeoutSeconds = (int)_messageBackoffStrategy.GetVisibilityTimeout(typedMessage, approximateReceiveCount).TotalSeconds;
-
-                        try
-                        {
-                            await UpdateMessageVisibilityTimeout(visibilityTimeoutSeconds, message.ReceiptHandle);
-                        }
-                        catch (Exception ex)
-                        {
-                            _log.LogError(0, ex, $"Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds");
-                        }
-                    }
+                    await UpdateMessageVisibilityTimeout(message, message.ReceiptHandle, typedMessage);
                 }
             }
         }
@@ -140,17 +128,30 @@ namespace JustSaying.AwsTools.MessageHandling
 
             await _queue.Client.DeleteMessageAsync(deleteRequest);
         }
-
-        private Task UpdateMessageVisibilityTimeout(int visibilityTimeoutSeconds, string receiptHandle)
+        
+        private async Task UpdateMessageVisibilityTimeout(SQSMessage message, string receiptHandle, Message typedMessage)
         {
-            var visibilityRequest = new ChangeMessageVisibilityRequest
+            if (message.Attributes.TryGetValue(MessageSystemAttributeName.ApproximateReceiveCount, out string rawApproxReceiveCount) && int.TryParse(rawApproxReceiveCount, out int approxReceiveCount))
             {
-                QueueUrl = _queue.Url,
-                ReceiptHandle = receiptHandle,
-                VisibilityTimeout = visibilityTimeoutSeconds
-            };
+                var visibilityTimeoutSeconds = (int)_messageBackoffStrategy.GetVisibilityTimeout(typedMessage, approxReceiveCount).TotalSeconds;
 
-            return _queue.Client.ChangeMessageVisibilityAsync(visibilityRequest);
+                try
+                {
+                    var visibilityRequest = new ChangeMessageVisibilityRequest
+                    {
+                        QueueUrl = _queue.Url,
+                        ReceiptHandle = receiptHandle,
+                        VisibilityTimeout = visibilityTimeoutSeconds
+                    };
+
+                    await _queue.Client.ChangeMessageVisibilityAsync(visibilityRequest);
+                }
+                catch (Exception ex)
+                {
+                    _log.LogError(0, ex, $"Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds");
+                    _onError(ex, message);
+                }
+            }
         }
     }
 }
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Threading.Tasks;
+using Amazon.SQS;
 using Amazon.SQS.Model;
+using JustSaying.Messaging.MessageProcessingStrategies;
 using JustSaying.Messaging.MessageSerialisation;
 using JustSaying.Messaging.Monitoring;
 using Message = JustSaying.Models.Message;
@@ -16,6 +18,7 @@ namespace JustSaying.AwsTools.MessageHandling
         private readonly IMessageMonitor _messagingMonitor;
         private readonly Action<Exception, SQSMessage> _onError;
         private readonly HandlerMap _handlerMap;
+        private readonly IMessageBackoffStrategy _messageBackoffStrategy;
 
         private static ILogger _log;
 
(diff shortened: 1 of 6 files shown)


JustSaying/Messaging/MessageProcessingStrategies/IMessageBackoffStrategy.cs
```diff
@@ -0,0 +1,9 @@
+﻿using JustSaying.Models;
+
+namespace JustSaying.Messaging.MessageProcessingStrategies
+{
+    public interface IMessageBackoffStrategy
+    {
+        int GetVisibilityTimeout(Message message, int approximateReceiveCount);
```
slang25: This interface is near perfect. Minor detail, could it return a `TimeSpan` and we just take `TotalSeconds` later.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -196,7 +198,8 @@ public void StopListening()
             {
                 QueueUrl = _queue.Url,
                 MaxNumberOfMessages = numberOfMessagesToReadFromSqs,
-                WaitTimeSeconds = 20
+                WaitTimeSeconds = 20,
+                AttributeNames = new List<string> {MessageSystemAttributeName.ApproximateReceiveCount}
```
slang25:
```
AttributeNames = (_messageBackoffStrategy != null) ? new List<string> {MessageSystemAttributeName.ApproximateReceiveCount} : new List<string>()
```

adammorr:
Updated as per suggestions @slang25 or @stuart-lang

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -84,6 +89,25 @@ public class MessageDispatcher
 
                 _onError(ex, message);
             }
+            finally
+            {
+                if (!handlingSucceeded && _messageBackoffStrategy != null)
+                {
+                    if (message.Attributes.ContainsKey(MessageSystemAttributeName.ApproximateReceiveCount) && int.TryParse(message.Attributes[MessageSystemAttributeName.ApproximateReceiveCount], out int approximateReceiveCount))
```
martincostello: Maybe move this into it's own method? This method's getting pretty long now.

slang25: Also this line could be written slightly more concisely (nitpicking now:
```csharp
if (message.Attributes.TryGetValue(MessageSystemAttributeName.ApproximateReceiveCount, out string approxCount) && int.TryParse(approxCount, out int approximateReceiveCount))
```

slang25:
Very nice, other than tests, is there anything else needed?

adammorr:
Not that I can think of, I'll get some tests added tomorrow then we should be good to go

adammorr:
@slang25 @martincostello had a rethink that it may be useful to pass through the last exception (if there is one) into the GetVisibilityTimeout method so the consumer can extend the visibility based on that...if you agree, let me know and I'll update the contract 👍

adammorr:
@slang25 @martincostello went with what I suggested anyway and added some tests...ready to go

slang25:
Very nice Adam. I like the addition of using access to the exception. My only last thought is naming, should 'GetVisibilityTimeout' be named something a little more agnostic (I suck at naming)

adammorr:
Can be but like you I am terrible at naming. Seems only fair that Martin should decide?

adammorr:
@slang25 GetMessageVisibilityDelay() ?

slang25:
@adammorr `GetBackoffDuration(...)`?
As this is a non-trivial feature, I'd like @AnthonySteele to have a look at this too.

adammorr:
Changed the method name @slang25

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -116,5 +131,30 @@ public class MessageDispatcher
 
             await _queue.Client.DeleteMessageAsync(deleteRequest);
         }
+        
+        private async Task UpdateMessageVisibilityTimeout(SQSMessage message, string receiptHandle, Message typedMessage, Exception lastException)
+        {
+            if (message.Attributes.TryGetValue(MessageSystemAttributeName.ApproximateReceiveCount, out string rawApproxReceiveCount) && int.TryParse(rawApproxReceiveCount, out int approxReceiveCount))
```
AnthonySteele: This is a big ugly condition. can we extract method, e.g.  `int? ReadApproximateReceiveCount` ?

AnthonySteele:
By default the `IMessageBackoffStrategy`  is null and behaviour is as before, right?
Is it worth including a  default implementation? Or even changing the default behaviour?

slang25:
@AnthonySteele Yes, the behavior is the same as before, it doesn't even ask to include `ApproximateReceiveCount` in the request if there is no strategy provided.

I see that we have 3 options
 - Allow an implementation to be built and passed in, with no default (what we have now in this PR)
 - Set a default sensible implementation that can be overridden.
 - Provide 1 or more out of the box implementations that we think could be useful.

The current implicit default is everything uses the queue configured visibility timeout, people may not want this to suddenly change, although this is a major release, so it's an opportunity to do that.

I think given that the interface is so simple and self explanatory, it's best to leave it whomever is implementing it, as we cannot make sensible assumptions about the duration of the work being carried out in the handlers. Unless we want to include the configured `VisibilityTimeoutSeconds` from the queue config in order to build an default linear/exponential backoff? @adammorr - thoughts?

Additionally, maybe what is now needed is rather than a mammoth README that has grown organically, we have a sample/samples folder, where we show some of this stuff along with comments, that way we can get across how we intend this to be used, without encouraging people to use a default without thinking that will likely be unsuitable for them. Obviously not to be tackled for this PR.

slang25:
I'm feeling ill, and I'm hacking about with this, the results will make your eyes bleed.

With  C#7 you can make something like this:
```csharp
var (parsed, approxReceiveCount) = parseApproxReceiveCount(message.Attributes);
```
^ not bad right, but I did it like this (be warned, it's nasty):

```csharp
private delegate bool TryParse<T>(string str, out T value);

private async Task UpdateMessageVisibilityTimeout(SQSMessage message, string receiptHandle, Message typedMessage, Exception lastException)
{
    (bool, TValue) TryGetValue<TKey, TValue>(IDictionary<TKey, TValue> dictionary, TKey key) => (dictionary.TryGetValue(key, out TValue val), val);

    (bool, string) TryGetApproximateReceiveCount(IDictionary<string, string> dictionary) =>
        TryGetValue(dictionary, MessageSystemAttributeName.ApproximateReceiveCount);


    (bool, T) TryParse<T>(string input, TryParse<T> parseFunc) => (parseFunc(input, out T val), val);

    (bool, int) TryParseInt(string input) => TryParse<int>(input, int.TryParse);


    Func<TInput, (bool, TOutput)> TryCombine<TInput, TIntermediate, TOutput>(
        Func<TInput, (bool success, TIntermediate value)> try1,
        Func<TIntermediate, (bool success, TOutput value)> try2)
    {
        (bool, TOutput) ResultFunction(TInput input)
        {
            var (success1, value1) = try1(input);
            return !success1 ? (false, default(TOutput)) : try2(value1);
        }

        return ResultFunction;
    }

    var parseApproxReceiveCount =
        TryCombine<IDictionary<string, string>, string, int>(TryGetApproximateReceiveCount, TryParseInt);

    var (parsed, approxReceiveCount) = parseApproxReceiveCount(message.Attributes);
    if (!parsed) throw new Exception("oh teh noes!");

    var visibilityTimeoutSeconds = (int) _messageBackoffStrategy
        .GetBackoffDuration(typedMessage, approxReceiveCount, lastException)
        .TotalSeconds;

    try
    {
        var visibilityRequest = new ChangeMessageVisibilityRequest
        {
            QueueUrl = _queue.Url,
            ReceiptHandle = receiptHandle,
            VisibilityTimeout = visibilityTimeoutSeconds
        };

        await _queue.Client.ChangeMessageVisibilityAsync(visibilityRequest);
    }
    catch (Exception ex)
    {
        _log.LogError(0, ex,
            $"Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds");
        _onError(ex, message);
    }
}
```

AnthonySteele:
I'm not ready for code like that last one, and I might never be.

Local functions aren't my favourite language feature: the only compelling case for them that I know of is recursion, where the local function name has to be in scope inside itself. I have used a language for years that supported local functions, by view f them is purely from experience of poor use of them.  In the e.g. above, many of those local fns are actually generally useful, if only they were accessible.

A `parseXyz` method that returns a tuple of `(success, value)` seems fine though as an alternative to using nullable.

adammorr:
@AnthonySteele @slang25 unless you want me to change the conditional then I'm happy for this to go

AnthonySteele:
Extract the parse logic from the conditional.

AnthonySteele:
I'm happy with it. 👍

Issue #99, opened 2015-08-26
Correlate messages across a distributed system


petemounce:
I would like to have a way of knowing why a message was processed - where it originated from, which user at the application-layer caused that message to be published and consumed, and things like that. I tend to think of messages like deferred HTTP requests.

In the HTTP world, I'd be able to add headers to the request like `x-conversation` with some unique ID, that can then be harvested & passed through as work ripples around a distributed system. This is _really_ useful for tracing work and user journeys. I'm able to use the same mechanism to decide how to handle requests, for example if I'm A/B testing.

Having talked to @payman81, we think this is possible via something like
- JustSaying makes a message `Envelope` (analogous to an HTTP request body) a 1st-class citizen, and available to handlers when they process messages.
  - the envelope can basically expose the SQS metadata, which opens up a set of enhancements like being able to know dwell times of the underlying transport, how many attempts that the message has had by a given handler, and some other stuff
- To (try to) avoid a breaking change, we talked about introducing a base class for `Handler` with a virtual method that consumers can choose to override, or not, which has access to the `Envelope`, and can make decisions based on it.

dariusdamalakas:
We solved this by just adding a `dict<string, string>` as the `Headers` property on the message, and added middleware to put correlation ids on the outgoing messages. If correlation ID exists on incoming message, we pass that through to outgoing messages. In fact, we have the same going on with HTTP requests, so we can track API requests and bus messages and correlate these. 

We should consider this carefully, but this might be a useful feature.

petemounce:
@dariusdamalakas that sounds like what we want too. Do you have code to share?

dariusdamalakas:
It's not trivial, as we've built a bus middleware concept around the `IHandler` interface, similar to what Owin gives with its middlewares, in a similar fashion with `app.Use(justSayingBusMiddleware)`. I can try and extract some of it into a gist to give an idea how we went around with it. I'd be happy to hear your thoughts whether any of it could be made generic enough to suite everyone.

payman81:
@dariusdamalakas It would be great If you could send us a gist with your snippets. I think that's something people can benefit from.

dariusdamalakas:
[...]

PR #322, opened 2017-04-20
When a message is created from another message, store the parent id

DrValani:
We have an issue, where we are at least 3rd in the chain of messages which have missing OrderId's. Trying to trace the source of message, which has been created from other messages. 

Would be grateful for any feedback or an alternate solution.

### DIFF
--- a/JustSaying.Models/Message.cs
+++ b/JustSaying.Models/Message.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 
 namespace JustSaying.Models
 {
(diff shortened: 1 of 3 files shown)


JustSaying.Models/Message.cs
```diff
@@ -8,11 +8,20 @@ protected Message()
         {
             TimeStamp = DateTime.UtcNow;
             Id = Guid.NewGuid();
+            CorrelationId = Id;
```
AnthonySteele: Yep, please check with other groups before going down this route - there is likely already such an Id in many http requests, and I don't know if it is a guid.

slang25: We must also make best efforts to keep deserialization compatibility with previous versions, we might be alright here, but we should double check

MuayThai: There's also a related issue in that I don't think JustSaying has a notion of system/header fields when serializing/deserializing. If the Message base class has a property added with the same name as field  being used by an application's message class already, there might be compilation errors (if access modifiers or types differ) or overwritten data. Just adding a field to the base class therefore becomes a potentially breaking change unless we can be confident the name is not in use already.

Ideally it would be nice if JustSaying wrapped app-level message classes in an Envelope<T> instead of requiring inheritance of a base class -  not entirely sure how feasible/difficult it would be to retro fit something like that.

slang25: @MuayThai You are right, and we have this on our roadmap for v6 (unofficially).

See here #99 

I would see it that we add a new envelope wrapper that pulls these framework-y bits off of the base class, and remove the need for the base class altogether. We'd need side-by-side support for at least one version.

For now, ignoring the property clash edge case, it needs to not break existing messages be deserialized to the updated class.

brainmurphy:
@DrValani Gonna close this - branch is out of date & quite old. Have been thinking about message envelopes as a better way of accomplishing this kind of thing.

PR #327, opened 2017-05-25
Move the fluent setup to async [DNMY]

slang25:
This is an effort to get rid of the synchronous AWS APIs (with the exception of `Publish`).

In order to achieve this with fluent api, I've introduce a `.Build()` method that should be awaited, which will perform all of the startup actions asynchronously.


JustSaying.IntegrationTests/AwsTools/SqsQueueIntegrationTests.cs
```diff
@@ -17,12 +17,12 @@ protected override SqsQueueByName CreateSystemUnderTest()
         {
             QueueUniqueKey = "test" + DateTime.Now.Ticks;
             var queue = new SqsQueueByName(RegionEndpoint.EUWest1, QueueUniqueKey, CreateMeABus.DefaultClientFactory().GetSqsClient(RegionEndpoint.EUWest1), 1, new LoggerFactory());
-            queue.Exists();
+            queue.ExistsAsync().GetAwaiter().GetResult();
```
AnthonySteele: Is it possible to make the setup and teardown async as well?

slang25: I didn't do this cause the base class is in JustBehave, however I could always create a new base class in this project, I'll take a look 👍

AnthonySteele:
It's a lot of code to take in. But I agree with the general ideas though, with deleting the sync wrapper methods of `.Create` etc, and I don't mind the `.Build()` method at the end of the fluent chain.

slang25:
lol, thanks resharper:
```xml
<ItemGroup>
    <Reference Include="JustBehave">
      <HintPath>..\..\..\Users\stuart.lang\.nuget\packages\justbehave\1.0.0.12\lib\net452\JustBehave.dll</HintPath>
    </Reference>
    <Reference Include="Ploeh.AutoFixture">
      <HintPath>..\..\..\Users\stuart.lang\.nuget\packages\autofixture\3.50.2\lib\net40\Ploeh.AutoFixture.dll</HintPath>
    </Reference>
  </ItemGroup>
```

slang25:
@AnthonySteele Done

README.md
```diff
@@ -84,17 +86,17 @@ We currently support SQS subscriptions only, but keep checking back for other me
...
-                // Some logic here ...
-                // e.g. bll.NotifyRestaurantAboutOrder(message.OrderId);
-                return true;
-            }
-        }
-````
+```csharp
+public class OrderNotifier : IHandler<OrderAccepted>
+{
+    public bool Handle(OrderAccepted message)
```
AnthonySteele: I would prefer the first example to use the async version, as the sync `IHandler` is deprecated.

slang25: Good call 👍

AnthonySteele: Also, I know you didn't write this, but " BLL layer" stands for (I think, I'm not sure) "Business Logic Layer" layer?

slang25:
I'm in FPH tomorrow so might drop by to talk about this some more, I'm not particularly happy with the return type from `.Build()`, but I'll explain in full tomorrow.

jaimalchohan:
How are we doing on this? I'm very keen on getting a version 5 out as we're super close to needing to use it in a production setting

tonyharverson:
This is a year old, and not being explored any more.

PR #328, opened 2017-07-21
Use nunit test adapter

slang25:


martincostello:
Looks good. Can we get Travis to run the tests now too?

slang25:
We'd have to run them with mono for now until deps are migrated right now, not sure how many would pass.
I'll check the status of our test dependencies cause we might be close to getting the tests to run on core, we might need to migrate off of JustBehave, or rewrite to use something else (BDDfy?)

martincostello:
Ah fair enough, I thought all that stuff had been pulled out already.

slang25:
Unfortunately not. I did have a go and then gave up. It looks like JustBehave can be ported pretty soon, the AutoFixture port is done and just needs merging to master and deploying, once that is done I think we are good to go. That might be easier than replacing it with something like BDDfy, although if someone comes along and does this, I'll merge it ( @shaynevanasperen ).

I'll be happy when you can just pull this down on Windows, macOS or Linux and it all just works and you can happily dev away 😃

AnthonySteele:
Good to go?

slang25:
@AnthonySteele Yes please

PR #329, opened 2017-07-25
Handler resolver with message in context

Sergej-Popov:
I have not created a separate issue, let me know if you want one.

This is related to #257 
The problem: today HandlerResolver doesn't have the context of the incoming message it only has the context of Queue. The context of the message is handy for providing extras to the IoC container. For example, it would be useful to provide the Tenant context (or any other message metadata) to the IoC container so that other components like Repositories can use correct connections strings etc. At the moment this has to be achieved by passing properties from handler down the call chain all the way to Repositories. I know the example is JE specific but the concept is general.

For example, my NinjectResolver could look like this:

```c#
    public class NinjectHandlerResolver : IHandlerResolver
    {
        private readonly IResolutionRoot _resolutionRoot;

        public NinjectHandlerResolver(IResolutionRoot resolutionRoot)
        {
            _resolutionRoot = resolutionRoot;
        }

        IHandlerAsync<T> ResolveHandler<T>(HandlerResolutionContextWithMessage context)
        {
            return _resolutionRoot.GetAll<IHandlerAsync<T>>(new ConstructorArgument("tenantContext", new TenantContext(context.Message.Tenant), true));
        }
    }
```

This could be further simplified by not having a check [here](https://github.com/Sergej-Popov/JustSaying/blob/feature/HandlerResolverWithMessageInContext/JustSaying/JustSayingFluently.cs#L271)  But instead performing this check just in time. Then there would be no need for:
`HandlerResolutionContextWithMessage `, `IHandlerAndMetadataResolver` and `HandlerResolverAdapter`

Alternative suggestions welcome.


slang25:
@Sergej-Popov apologies for not looking at this sooner. This will need rebasing, as we've made changes to the resolver so that it happens per message, that likely overlaps with what's done here, or possibly simplifies it after some re-work in this PR.

I could see how this would be useful specifically for some JE scenarios.

I feel that we should keep JustSaying as simple as possible in terms of responsibility, this functionality could be very simply done inside of the `Handler`, you handler becomes your entry point, which could be a simple interceptor that fans out the `Message` to other handlers.

That said, I'm conflicted as I can see it would be useful. So it's a case that more is less, but more is also more 😆

Does anyone else have thoughts around this features?
@AnthonySteele @martincostello

Sergej-Popov:
@stuart-lang thanks for finding time for this =)

Originally this PR was to help with multitenancy by allowing for tenant context to be available at handler resolution time. 

> could be very simply done inside of the Handler

only if you have a handful of handlers yes. If like us you have hundreds then it becomes very unwieldy. We had to do some code generation to overcome this issue. We want to get rid of this code generation, hence this PR is still open.

If I remember correctly there were talks about adding a headers dictionary for storing meta-data like Tenant and idea of moving away from Tenant property (and some other properties) on the message itself. It would be optimal to have such headers available at resolution time rather than the message.

I would rather everyone come to an agreement on how that should look like before re-basing (or rather re-implementing) this. Personally, I would much rather see the headers + resolver supplying headers combination than what was in this PR. Let me know.

Sergej-Popov:
Closing as this is almost 1 year old. I don't think it is realistic that it will be merged. Might revisit the problem once other features like message envelope/meta data headers will be implemented as it might open alternative ways of solving this issue.

Issue #208, opened 2016-06-07
A queue already exists with the same name and a different value for attribute RedrivePolicy


dariusdamalakas:
I am not yet sure as to the origins of the error, potentially JustSaying is trying to create the queue twice, but I have not looked what is the root cause.

```
 justSaying
                .WithSqsTopicSubscriber()
                    .IntoQueue("queue1")
                        .ConfigureSubscriptionWith(x =>
                        {
                            x.VisibilityTimeoutSeconds = (int)TimeSpan.FromHours(12).TotalSeconds;
                            x.RetryCountBeforeSendingToErrorQueue = 1;
                        })
                .WithMessageHandler<Handler1>(handlerResolver)
                .WithSqsPointToPointSubscriber()
                    .IntoQueue("queue1")
                    .ConfigureSubscriptionWith(x =>
                    {
                        x.VisibilityTimeoutSeconds = (int)TimeSpan.FromHours(12).TotalSeconds;
                        x.RetryCountBeforeSendingToErrorQueue = 1;
                    })
                .WithMessageHandler<Handler2>(handlerResolver)
```

The above code fails with this error. I have not found a workaround for this, except for removing sqs point to point subscriber and switching to sns subscriber

```
2016-06-07 09:19:24,191 [7] INFO  JustSaying [(null)] - Created Queue: dev-dariouso-ringcentral2_error on Arn: arn:aws:sqs:eu-west-1:xxxxxx:dev-dariouso-ringcentral2_error
2016-06-07 09:19:24,268 [7] ERROR JustSaying [(null)] - Create Queue error: dev-dariouso-ringcentral2
Amazon.SQS.Model.QueueNameExistsException: A queue already exists with the same name and a different value for attribute RedrivePolicy ---> Amazon.Runtime.Internal.HttpErrorResponseException: The remote server returned an error: (400) Bad Request. ---> System.Net.WebException: The remote server returned an error: (400) Bad Request.
   at System.Net.HttpWebRequest.GetResponse()
   at Amazon.Runtime.Internal.HttpRequest.GetResponse() in d:\Jenkins\jobs\build-sdkandtools-release\workspace\sdk\src\AWSSDK_DotNet35\Amazon.Runtime\Pipeline\HttpHandler\HttpWebRequestFactory.cs:line 106
   --- End of inner exception stack trace ---
[...]

PR #332, opened 2017-09-18
Set Queue Attribute after create (port to v4 from master)

tomhaigh:
Copy and update the fix from master to v4 branch to fix issue #208

### DIFF
--- a/JustSaying.AwsTools.IntegrationTests/WhenSettingUpMultipleHandlers.cs
+++ b/JustSaying.AwsTools.IntegrationTests/WhenSettingUpMultipleHandlers.cs
@@ -2,6 +2,8 @@ using System;
 using System.Linq;
 using System.Threading.Tasks;
 using Amazon;
+using Amazon.SQS.Model;
+using Amazon.SQS.Util;
 using JustBehave;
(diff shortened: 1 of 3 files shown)


JustSaying.AwsTools.IntegrationTests/WhenSettingUpMultipleHandlers.cs
```diff
@@ -97,5 +99,18 @@ public void CreateQueueCalledOnce()
         {
             Assert.That(proxyAwsClientFactory.Counters["CreateQueue"][queueName].Count, Is.EqualTo(1));
         }
+
+        [Test]
+        public void SetQueueAttributesCalledOnceWithBasicAttributes()
+        {
+            var allSetQueueAttributesCalls = proxyAwsClientFactory.Counters["SetQueueAttributes"]
```
tomhaigh: This is ugly but not sure how to do it better. There are many calls to `SetQueueAttributes()` and I'm only interested in the one that happens immediately after the queue is created

tomhaigh:
also see recent update to #208

brainmurphy:
Looks OK to me 👍

PR #334, opened 2017-11-10
Add ApproximateReceiveCount to Message

zaccharles:
Including `ApproximateReceiveCount` on `Message` would allow handlers to perform different actions based on approximately how many times a message has been received. For example, checking a cache before falling back to calling an API for data.

### DIFF
--- a/JustSaying.Models/Message.cs
+++ b/JustSaying.Models/Message.cs
@@ -20,6 +20,7 @@ namespace JustSaying.Models
         public string ReceiptHandle { get; set; }
         public string QueueUrl { get; set; }
         public int? DelaySeconds { get; set; }
+        public int ApproximateReceiveCount { get; set; }
 
         //footprint in order to avoid the same message being processed multiple times.
         public virtual string UniqueKey() => Id.ToString();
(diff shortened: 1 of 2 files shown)


slang25:
Hi @zaccharles, sorry to leave this PR for so long with no comment, not cool.
I am very hesitant to accept additions to the message base class for the following reasons:

- Properties that are non-nullable are can not deserialize older messages (that could easily be fixed here)
- Due to horrible historic reasons, the `Messages` assembly lived in 2 packages, then in only `JustSaying`, now only in `JustSaying.Models`, any changes between v4 and v5 definitions might cause runtime exceptions if the "wrong" assembly wins during build. I hate this.
- I feel weird adding more attribute related stuff to the `Message` base class, it feels like mixing concerns.

We added the ability to have [variable backoff here](https://github.com/justeat/JustSaying/pull/321) so that might what you want, not sure if that covers your use case.

I'd like to get a better solution to this in v6, I think having a base class is not idea, and there needs to be a nice way to get at message attributes like this.

slang25:
We will be discussing a way to get access to the underlying SQS message receive response, that will eliminate the need to forever extend the `Message` base class.

PR #311, opened 2017-03-15
Improve performance (for cases when topic count is large)

MuayThai:
- Switch from using ListTopics to get loading the ARN (simply by calling Create() - even if they exist already)
- Fix broken integration tests
- Update brittle integration tests that broke due to being coding against implementation (eg. method call count assertions for AWS api calls)

PR #336, opened 2017-11-19
Cherrypicked "Improve performance (for cases when topic count is large) (#311)"

slang25:
This was a painful process, I will re-check this this evening. We'll need to beta this before we go live with v5


brainmurphy:
👍

PR #339, opened 2017-12-05
lighter dependencies for the models assembly

AnthonySteele:
`JustSaying.Models` is very small, contains just the `Message` base class. 
It doesn't use `Newtonsoft.json`, so should not specify a version for consumers etc


slang25:
👍 Not sure what happened here

PR #340, opened 2018-01-07
Add ability to handle exception on a per notification basis

JosephWoodward:
There are instances where it's acceptable if a notification fails to publish (say it times out for instance). This adds the ability for the consumer to dictate how exceptions are handled on a notification by notification basis as opposed to having to litter their code with try/catches in every instance the notification is published.

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
+++ b/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
@@ -2,6 +2,7 @@ using System;
 using System.Threading.Tasks;
 using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
+using JustSaying.AwsTools.QueueCreation;
 using JustSaying.Messaging;
 using JustSaying.Messaging.MessageSerialisation;
 using Microsoft.Extensions.Logging;
@@ -12,6 +13,7 @@ namespace JustSaying.AwsTools.MessageHandling
     public abstract class SnsTopicBase : IMessagePublisher
     {
         private readonly IMessageSerialisationRegister _serialisationRegister; // ToDo: Grrr...why is this here even. GET OUT!
+        public SnsWriteConfiguration SnsWriteConfiguration { private get; set; }
         public string Arn { get; protected set; }
         protected IAmazonSimpleNotificationService Client { get; set; }
         private readonly ILogger _eventLog;
@@ -53,9 +55,10 @@ namespace JustSaying.AwsTools.MessageHandling
             }
             catch (Exception ex)
             {
-                throw new PublishException(
-                    $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}",
-                    ex);
+                if (!ClientExceptionHandler(ex, request))
+                    throw new PublishException(
+                        $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}",
+                        ex);
             }
         }
 #endif
@@ -67,14 +70,25 @@ namespace JustSaying.AwsTools.MessageHandling
             try
             {
                 await Client.PublishAsync(request);
+
                 _eventLog.LogInformation($"Published message: '{request.Subject}' with content {request.Message}");
             }
             catch (Exception ex)
             {
-                throw new PublishException(
-                    $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}",
-                    ex);
+                if (!ClientExceptionHandler(ex, request))
+                    throw new PublishException(
+                        $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}",
+                        ex);
             }
+        }
+
+        private bool ClientExceptionHandler(Exception ex, PublishRequest request)
+        {
+            bool exceptionIsHandled = false;
+            if (SnsWriteConfiguration?.OnException != null)
+                exceptionIsHandled = SnsWriteConfiguration.OnException(ex, request);
+
+            return exceptionIsHandled;
         }
 
         private PublishRequest BuildPublishRequest(Message message)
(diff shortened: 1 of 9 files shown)


JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -12,6 +13,7 @@ namespace JustSaying.AwsTools.MessageHandling
     public abstract class SnsTopicBase : IMessagePublisher
     {
         private readonly IMessageSerialisationRegister _serialisationRegister; // ToDo: Grrr...why is this here even. GET OUT!
+        public SnsWriteConfiguration SnsWriteConfiguration { private get; set; }
```
martincostello: Public set, private get seems a bit weird to me. Why not just make them both public?

slang25: Was just about to make the same comment. I'd like this to be set only on construction, if that means adding an overload to the constructor and having a field then so be it.

JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace JustSaying.AwsTools.QueueCreation
+{
+    public class SnsWriteConfiguration
+    {
+        /// <summary>
+        /// Extension point allows custom error handling, including ability to specify whether exception has been explictly handled by consumer.
+        /// </summary>
+        /// <returns>Boolean to indicate whether the exception has already been handled by the consumer</returns>
+        public Func<Exception, PublishRequest, bool> OnException { get; set; }
```
slang25: I'm thinking this should be called `HandleException` because you have to indicate if it is handled or still requires propagation.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -67,16 +70,27 @@ public void Publish(Message message)
...
+                        $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}",
+                        ex);
             }
         }
 
+        private bool ClientExceptionHandler(Exception ex, PublishRequest request)
+        {
+            bool exceptionIsHandled = false;
+            if (SnsWriteConfiguration?.OnException != null)
+                exceptionIsHandled = SnsWriteConfiguration.OnException(ex, request);
```
slang25: Could use `?.Invoke...` for a single liner

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
slang25: tbf it doesn't help that this class is public when it should really be internal. I have a branch where I go nuts making things internal, which can wait for v6

JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
slang25: Also, is there a good reason to require `PublishRequest`, a lot of the extensibility tries to hide AWS specific types from the API surface, I would prefer this abstracted, of not included if not really needed, the exception contains things like endpoint etc...

slang25:
Suggestions: https://github.com/JosephWoodward/JustSaying/pull/1

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
JosephWoodward: No, that's a fair point @martincostello. I'll update it.

JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
JosephWoodward: I was considering `HandleException` but decided against it simply because you may not be handling the exception. The method is invoked when an exception occurs, whether you choose to handle it is down to the consumer. Happy either way though.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
JosephWoodward: 👍

JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
JosephWoodward: Didn't realise you were so keen on hiding the AWS API but it makes sense, main reason for including it is if the consumer wishes to log the exception the `PublishRequest` object provides more context around the request (bearing in mind the `PublishException` is not thrown if they choose to handle the exception themselves - [see here](https://github.com/justeat/JustSaying/pull/340/files#diff-f6c5b29e7f5c2decc9b2ec6ff6a449c3R79).

Happy either way on both points though!

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
JosephWoodward: Wasn't sure about constructor injection when I did it but in hindsight it's a good call. Update incoming.

JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
stuart-lang: Cool, I vote for:
```csharp
public Func<Exception, bool> HandleException { get; set; }
```

JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Amazon.SimpleNotificationService.Model;
+
+namespace JustSaying.AwsTools.QueueCreation
+{
+    public class SnsWriteConfiguration
+    {
+        /// <summary>
+        /// Extension point allows custom error handling, including ability to specify whether exception has been explictly handled by consumer.
```
tomhaigh: Minor, but I find this comment a bit hard to understand - because it talks about consumers, but I think it means the consumer as in the caller (which is the publisher). And we also have consumers that are listening to messages, but they are a different type of consumer.

JosephWoodward: That's a good point @tomhaigh, I will update.

JosephWoodward:
Alt Text

PR #341, opened 2018-01-14
v5 rc1

slang25:


slang25:
Just need to tell AppVeyor which what the build details are, because I don't want to maintain this in two places.
https://www.appveyor.com/docs/build-worker-api/#update-build-details

slang25:
Oh, and include symbols!

slang25:
Ok @AnthonySteele, I think this is ready to go. I don't think I can quickly get the appveyor build number to pick it up from the package, so we have to keep the `appveyor.yml` version number updated with the `.props` file. I might build something to do it later.

I've used embedded source files with JustBehave and it works great, so am turning them on here. I didn't embed the symbols into the dlls like I did in JustBehave, cause I'm on the fence about that.

Embedded source is more useful for internal project, and git-link is usually used for OSS projects, but I think source embedding is the future, and reckon everyone will adopt it at some point.

I've simplified the versioning, what I was doing with the AssemblyVersion was a bit funky and unneeded.
We now control the exact package version and when we tag the repo we get a nice clean package, any other commits will have a build123 suffix.

While we are testing this, I can put together some release notes and upgrade instructions.

I think we should get together with a few others after this and have a big brainstorm about v6, there are lots of directions we could go in and it's worth sharing our thoughts.

PR #344, opened 2018-02-12
Add IMessageLockAsync - asynchronous specification of IMessageLock

arturgasparian:


JustSaying/Messaging/MessageHandling/BlockingMessageLock.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                throw new ArgumentNullException(nameof(inner));
+            }
+
+            Inner = inner;
+        }
+
+        public IMessageLock Inner { get; }
+
+        public Task ReleaseLockAsync(string key) => Task.Run(() => Inner.ReleaseLock(key));
```
stuart-lang: Could this please be changed to:
```csharp
public Task ReleaseLockAsync(string key)
{
    Inner.ReleaseLock(key);
    return Task.Completed;
}
```

JustSaying/Messaging/MessageHandling/BlockingMessageLock.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.MessageHandling
+{
+    /// <summary>
+    /// Used to convert "IMessageLock " instances into IMessageLockAsync
+    /// So that the rest of the system only has to deal with IMessageLockAsync
+    /// </summary>
+    public class BlockingMessageLock : IMessageLockAsync
```
slang25: Can we change this from `public` to `internal`

JustSaying/Messaging/MessageHandling/BlockingMessageLock.cs
arturgasparian: We can't using CompletedTask for .NET 4.5.1

tomhaigh: Could use `Task.FromResult()` ?

tomhaigh:
Was wondering what the benefit is of adding async methods if they don't actually do anything asynchronously?

stuart-lang:
@tomhaigh We have an internal library with an implementation of `IMessageLock`, that uses dynamodb as a backing store, if we wish to migrate to AWSSDK v3, we need an async version.

This change is done in a light touch way and will wrap the existing sync version. I am on the fence between dropping support for `IMessageLock` (in favour of `IMessageLockAsync`) as we are in rc, at the same time, this could be a minor version bump and we get it out in 5.1

You thoughts on this are welcome.

PR #349, opened 2018-03-15
Add cancellation token overloads to publish/subscribe methods

JosephWoodward:

PR #350, opened 2018-03-19
Request Handling [WIP]

JosephWoodward:


JustSaying/AwsTools/MessageHandling/MessageResultLogger.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace JustSaying.AwsTools.MessageHandling
+{
+    /// <summary>
+    /// Handler allowing you to access the SQS or SNS message response.
+    /// </summary>
+    /// <param name="result">The result of the SNS notification or SQS enqueue.</param>
+    /// <returns></returns>
+    public delegate void MessageResultLogger(MessageResult result);
```
martincostello: Could you just use `Action<MessageResult>` instead of having to declare a custom delegate type?

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -68,17 +72,21 @@ public void Publish(Message message)
                         $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}",
                         ex);
             }
+            finally
+            {
+                _messageResultLogger?.Invoke(new MessageResult(response?.MessageId, response?.HttpStatusCode));
```
martincostello: Is it that useful to call the delegate if all the properties of the `MessageResult` are `null`?

JosephWoodward: I think so personally, call the delegate regardless of response for a couple of reasons:
- Principle of least surprise (As a consumer I'd expect the handler to be invoked regardless of response).
- Reduces complexity, far simpler API

JustSaying/AwsTools/MessageHandling/MessageResultLogger.cs
JosephWoodward: Yeah, this is going

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
martincostello: Would it be better to extend it to optionally also provide the exception? Otherwise getting an object with null properties just means "something went wrong" without telling you what. Then it's always called and always has something useful to tell the callee.

JosephWoodward: I think that's a good idea

JosephWoodward: Actually, on further thought I think we go with your initial comment as we already have an exception handler. I'll move this call out of the `finally` block and into the `try` one.

JustSaying.UnitTests/AwsTools/MessageHandling/Sns/TopicByName/WhenPublishingAsyncExceptionCanBeHandled.cs
```diff
@@ -21,11 +21,11 @@ public class WhenPublishingAsyncExceptionCanBeHandled : XAsyncBehaviourTest<SnsT
         private readonly IMessageSerialisationRegister _serialisationRegister = Substitute.For<IMessageSerialisationRegister>();
         private readonly IAmazonSimpleNotificationService _sns = Substitute.For<IAmazonSimpleNotificationService>();
         private const string TopicArn = "topicarn";
-        private readonly MessageResultLogger _messageResultLogger = r => {};
+        private readonly Action<MessageResponse> _responseLogger = r => {};
```
martincostello: Is it worth adding a static property somewhere that all the tests (or integrations) can just use as a "no-op" for this to make it easier?

```csharp
public static Action<MessageResponse> NoOp { get; } = (_) => { };
```

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -72,22 +77,23 @@ public void Publish(Message message)
...
             var request = BuildPublishRequest(message);
-            PublishResponse response = null;
 
             try
             {
-                response = await Client.PublishAsync(request).ConfigureAwait(false);
+                var response = await Client.PublishAsync(request).ConfigureAwait(false);
                 _eventLog.LogInformation($"Published message: '{request.Subject}' with content {request.Message}");
+
+                _messageResponseLogger?.Invoke(new MessageResponse
```
martincostello: Open question - is there any use case for this being async or there being an alternate async version so integrators have the _option_ of doing something async in their logger?

JosephWoodward: I think so. I've been inclined to change it from logger or something like handler or event to make it more open ended.

JustSaying.UnitTests/AwsTools/MessageHandling/Sns/TopicByName/WhenPublishingAsyncExceptionCanBeHandled.cs
```diff
@@ -23,7 +23,7 @@ public class WhenPublishingAsyncExceptionCanBeHandled : XAsyncBehaviourTest<SnsT
 
         protected override SnsTopicByName CreateSystemUnderTest()
         {
-            var topic = new SnsTopicByName("TopicName", _sns, _serialisationRegister, Substitute.For<ILoggerFactory>(), new SnsWriteConfiguration
+            var topic = new SnsTopicByName("TopicName", _sns, _serialisationRegister, _ => {}, Substitute.For<ILoggerFactory>(), new SnsWriteConfiguration
```
JosephWoodward: As Martin mentioned in his comment, happy to move this into a static `NoOp` property for clarity.

JustSaying/IPublishConfiguration.cs
```diff
@@ -1,11 +1,15 @@
...
+
+namespace JustSaying
+{
+    public interface IPublishConfiguration
+    {
+        int PublishFailureReAttempts { get; set; }
         int PublishFailureBackoffMilliseconds { get; set; }
-        IReadOnlyCollection<string> AdditionalSubscriberAccounts { get; set; }
-    }
+        Action<MessageResponse> ResponseLogger { get; set; }
```
JosephWoodward: Interested to know thoughts on Martin's question around either making this async or having an async version (considering JustSaying has a synchronous and asynchronous publish call).

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -14,16 +14,16 @@ public class SqsPublisher : SqsQueueByName, IMessagePublisher
     {
         private readonly IAmazonSQS _client;
         private readonly IMessageSerialisationRegister _serialisationRegister;
-        private readonly Action<MessageResponse> _messageResponseLogger;
+        private readonly MessageResponseLogger _messageResponseLogger;
 
         public SqsPublisher(RegionEndpoint region, string queueName, IAmazonSQS client,
             int retryCountBeforeSendingToErrorQueue, IMessageSerialisationRegister serialisationRegister,
-            Action<MessageResponse> messageResponseLogger, ILoggerFactory loggerFactory)
+            MessageResponseLogger MessageResponseLogger, ILoggerFactory loggerFactory)
```
martincostello: `messageResponseLogger`

JustSaying/MessageResponseLogger.cs
```diff
@@ -0,0 +1,18 @@
...
+using JustSaying.AwsTools.MessageHandling;
+
+namespace JustSaying
+{
+    public class MessageResponseLogger
+    {
+        public MessageResponseLogger()
+        {
+            ResponseLogger = _ => { };
+            ResponseLoggerAsync = _ => Task.FromResult(0);
```
martincostello: Can you use `Task.CompletedTask` here too, or are we targeting too low a TFM for it?

JosephWoodward: Too lower framework.

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
JosephWoodward: Good catch 👍

JustSaying/NoOpMessageResponseLogger.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Threading.Tasks;
+using JustSaying.AwsTools.MessageHandling;
+using JustSaying.Models;
+
+namespace JustSaying
+{
+    internal class NoOpMessageResponseLogger : IMessageResponseLogger
```
martincostello: This isn't _strictly_ no-op, as it has properties set to `null`, rather than delegates that just do nothing.

JosephWoodward: Forgot to rename that after switching them to nulls!

JosephWoodward: There we go 👍

JosephWoodward: Renamed to `DefaultMessageResponseLogger`.

JustSaying/JustSaying.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netstandard1.6;net451</TargetFrameworks>
+    <LangVersion>7.1</LangVersion>
```
martincostello: Could just define these in `Directory.Build.props` instead?

martincostello:
I guess this will conflict with the `CancellationToken` changes in #349.

JosephWoodward:
@martincostello Yes it will, hopefully get the cancellation token changes in first then I'll rebase this on top.

JustSaying/JustSaying.csproj
JosephWoodward: 👍

JustSaying/IMessageResponseLogger.cs
```diff
@@ -0,0 +1,20 @@
...
+using JustSaying.Models;
+
+namespace JustSaying
+{
+    public interface IMessageResponseLogger
+    {
+        /// <summary>
+        /// Capture response of message after it has been sent.
+        /// </summary>
+        Action<MessageResponse, Message> ResponseLogger { get; set; }
```
shaynevanasperen: It would be simpler to just expose these as normal method signatures rather than properties that redirect to an `Action` and a `Func`.

JustSaying/NullMessageResponseLogger.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Threading.Tasks;
+using JustSaying.AwsTools.MessageHandling;
+using JustSaying.Models;
+
+namespace JustSaying
+{
+    internal class NullMessageResponseLogger : IMessageResponseLogger
+    {
+        public Action<MessageResponse, Message> ResponseLogger { get; set; }
```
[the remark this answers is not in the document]
shaynevanasperen: Also, it might be simpler to just expose these methods as normal method signatures instead of properties of `Action` and `Func`. You could provide a base class that defines no-op implementations of both, and then implementors can just override one or both as they wish.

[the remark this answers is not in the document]
JosephWoodward: >You should make these actually return non-null instances, so that the calling code doesn't need to do null-conditional evaluations.

They were originally no-op implementations but we need to [detect if they're not set in order to call the synchronous version of the async version isn't provided](https://github.com/justeat/JustSaying/pull/350/files#diff-f6c5b29e7f5c2decc9b2ec6ff6a449c3R92).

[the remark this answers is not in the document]
shaynevanasperen: @JosephWoodward How about only exposing the asynchronous version and then if people want to implement it synchronously they can just return `Task.CompletedTask`?

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -82,20 +91,35 @@ public void Publish(Message message)
...
             try
             {
-                await Client.PublishAsync(request, cancellationToken).ConfigureAwait(false);
-
+                var response = await Client.PublishAsync(request, cancellationToken).ConfigureAwait(false);
                 _eventLog.LogInformation($"Published message: '{request.Subject}' with content {request.Message}");
+
+                if (_messageResponseLogger?.ResponseLoggerAsync != null)
+                {
+                    await _messageResponseLogger.ResponseLoggerAsync(new MessageResponse
```
slang25: `.ConfigureAwait(false)` please

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -46,10 +56,25 @@ public void Publish(Message message)
         public async Task PublishAsync(Message message, CancellationToken cancellationToken)
```
AnthonySteele: Why do we not return the response? We don't like consumers to interact with AWS?

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -46,10 +56,25 @@ public void Publish(Message message)
...
             var request = BuildSendMessageRequest(message);
-
             try
             {
-                await _client.SendMessageAsync(request, cancellationToken).ConfigureAwait(false);
+                var response = await _client.SendMessageAsync(request, cancellationToken).ConfigureAwait(false);
+
+                if (_messageResponseLogger?.ResponseLoggerAsync != null)
+                {
+                    await _messageResponseLogger.ResponseLoggerAsync(new MessageResponse
```
slang25: `.ConfigureAwait(false)` please

Directory.Build.props
```diff
@@ -13,6 +13,7 @@
     <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
     <PackageTags>aws;sns;sqs</PackageTags>
     <RepositoryType>git</RepositoryType>
+    <LangVersion>7.1</LangVersion>
```
slang25: `latest` please

JustSaying/IMessageResponseLogger.cs
AnthonySteele: I agree with Shayne. If it's properties that you have to set, it's simpler to have properties on a DTO `public interface MessageResponseLoggerConfig`. If it's an interface that you have to implement, then methods.

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
slang25: @JosephWoodward asked this question when starting the work.

I don't like it for 2 reasons.

First it's not an intuitive fit for the API, there is something very elegant and intuitive about a Publish method that is effectively a command.

Second, JustSaying tries not to expose AWSSDK types where possible, it is an abstraction, rather than a composable toolkit. There is an argument to be made to favour a composable toolkit, however this isn't what the library is, so we should stay with it IMO.

AnthonySteele: I agree with your second comment about what JustSaying _is_. Can have the discussion about which is preferable some other time.

JustSaying/MessagingConfig.cs
```diff
@@ -13,10 +16,12 @@ public MessagingConfig()
             PublishFailureBackoffMilliseconds = JustSayingConstants.DEFAULT_PUBLISHER_RETRY_INTERVAL;
             AdditionalSubscriberAccounts = new List<string>();
             Regions = new List<string>();
+            MessageResponseLogger = (r, m) => {};
```
slang25: May as well leave it as `null` as it's always invoked with the null conditional operator

PR #352, opened 2018-03-31
Dont expose blocking calls externally, except for fluent api

slang25:


JustSaying.Tools/Commands/MoveCommand.cs
```diff
@@ -62,9 +63,9 @@ public bool Execute()
             return true;
         }
 
-        private void EnsureQueueExists(SqsQueueByName queue)
+        private async Task EnsureQueueExists(SqsQueueByName queue)
```
JosephWoodward: Should this not be called `EnsureQueueExistsAsync`?

slang25: Thanks @JosephWoodward, this has been addressed

JosephWoodward:
@slang25 Done 👍

slang25:
@AnthonySteele You are right, at the moment they have been lifted up the fluent api methods, and we warn about this in the release notes (see first item under [known issues here](https://github.com/justeat/JustSaying/releases/tag/v5.0.0)).

This removes the other public apis with the blocking, that people may have accidentally called and get tripped up by. Pull async fluent api will come in a later release 👍

PR #356, opened 2018-05-29
Update testing tools

AnthonySteele:
Apply updates to testing tools.
In some cases we were still using a NETCore-compatible beta or RC, which is not needed any more, there are released versions.

Also updated `SourceLink.Embed.AllSourceFiles` to latest. Is this still the way forward? It's not the same as [this pr](https://github.com/justeat/JustEat.StatsD/pull/71/files) but then it's not the NETCore 2.1 SDK either. If necessary I can back that out and we'll address source linking separately. But since it's included, might as well use latest?

### DIFF
--- a/JustSaying.Models/JustSaying.Models.csproj
+++ b/JustSaying.Models/JustSaying.Models.csproj
@@ -6,4 +6,7 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="SourceLink.Embed.AllSourceFiles" Version="2.8.1" />
+  </ItemGroup>
 </Project>
--- a/JustSaying.Models/JustSaying.Models.csproj
+++ b/JustSaying.Models/JustSaying.Models.csproj
@@ -6,7 +6,4 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
   </ItemGroup>
-  <ItemGroup>
-    <PackageReference Update="SourceLink.Embed.AllSourceFiles" Version="2.8.1" />
-  </ItemGroup>
 </Project>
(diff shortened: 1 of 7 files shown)


JustSaying.Models/JustSaying.Models.csproj
```diff
@@ -6,4 +6,7 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="SourceLink.Embed.AllSourceFiles" Version="2.8.1" />
```
martincostello: If you put this into `Directory.Build.props` it will apply to every project in the solution.

JustSaying.Tools/JustSaying.Tools.csproj
```diff
@@ -20,4 +20,7 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="SourceLink.Embed.AllSourceFiles" Version="2.8.1" />
```
martincostello: Not needed if moved as commented above.

JustSaying/JustSaying.csproj
```diff
@@ -24,6 +24,9 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="SourceLink.Embed.AllSourceFiles" Version="2.8.1" />
```
martincostello: Ditto.

JustSaying.Models/JustSaying.Models.csproj
AnthonySteele: I didn't add that, so the tools must have. Should we move it?

martincostello:
AFAIK embedding source files in the PDB and SourceLink are different things. SourceLink puts a pointer into the NuGet package so that a compatible tool can fetch the source, where as the embedding puts the source into the PDB itself, meaning no additional fetching is required, though that's only useful if the PDB is in the NuGet package in the first place.

JustSaying.Models/JustSaying.Models.csproj
AnthonySteele: Yep, it needs to go in `Directory.Build.props` since the old version is there. Tooling has made a boo-boo.

Issue #351, opened 2018-03-28
QUESTION: AWS SNS/SQS Message Filtering?


eric-davis:
Do you all have any plans to support AWS SNS/SQS Message Filtering anytime soon?

stuart-lang:
I'd love this, and would like to open up the discussion around this.

Currently we have properties in the `Message` base class that we want to either move, or copy into message attributes (which is required for filtered subscriptions).

We'd need to balance backward compatibility with this feature, there is some overlap with the concept of "Message Envelopes" which has been discussed in other issues.

Maybe we start with a strawman implementation in a PR and work from there?

slang25:
Just to get the ball rolling on discussion:
```csharp
  var publisher = CreateMeABus.WithLogging(loggerFactory)
                    .InRegion(RegionEndpoint.EUWest1.SystemName)
                     .WithSnsMessagePublisher<GenericMessage>(cfg => {
                    cfg.MessageAttributes = new Dictionary<string,Func<GenericMessage,string>>{
                        [nameof(GenericMessage.Tenant)] = x => x.Tenant
                    };
                });
```

```csharp
var subscriber = CreateMeABus
                .WithLogging(new LoggerFactory())
                .InRegion(primaryRegion)
                .WithFailoverRegion(secondaryRegion)
                .WithActiveRegion(() => primaryRegion)
                .WithSqsTopicSubscriber()
                .IntoQueue("queuename")
                .WithFilter(nameof(GenericMessage.Tenant), "uk") // Filter here
                .WithMessageHandler(handler);
```

Issue #278, opened 2016-10-11
Generic Message types are not supported


eric-davis:
I've just started looking at JustSaying as a possible message bus solution, but have run into an issue with the handler registration system and message serialization when attempting to use generic message types. For example, I have an existing domain event class I would like to publish via JustSaying. I can't just have my event class derive from JustSaying.Models.Message since I am already deriving that event class from a base event type. And the event type has a property named Version that is of type integer that would collide with the Message class' string Version property. So, I decided to define a JustSaying message wrapper class like so:

```
public class MessageWrapper<TEvent> : Message where TEvent : IEvent
{
    public MessageWrapper(TEvent domainEvent)
    {
        this.DomainEvent = domainEvent;
    }

    public TEvent DomainEvent { get; private set; }
}
```

Then I configure the JustSaying bus like so...

```
var basicCredentials = new BasicAWSCredentials(awsAccessKeyId, awsAccessKeySecret);
CreateMeABus.DefaultClientFactory = () => new DefaultAwsClientFactory(basicCredentials);

var bus =
CreateMeABus.InRegion(awsRegion.SystemName)
    .ConfigurePublisherWith(c => c.PublishFailureReAttempts = 3)
    .WithSnsMessagePublisher<MessageWrapper<WidgetCreated>>()
    .WithSnsMessagePublisher<MessageWrapper<WidgetNameChanged>>()
    .WithSqsTopicSubscriber()
    .IntoQueue("justsaying_events")
    .WithMessageHandler<MessageWrapper<WidgetCreated>>(new WidgetCreatedHandler())
    .WithMessageHandler<MessageWrapper<WidgetNameChanged>>(new     WidgetNameChangedHandler());

bus.StartListening();
```

The issue is the way in which the message type's .Name property is serialized into the SNS message Subject and used for both the handler registration key and the serialization type. This method appears to be too simplistic to support generic message types. If I define and publish a message with a type of `MessageWrapper<WidgetCreated>`, the resulting message type's .Name value is "MessageWrapper'1". It does not include the generic parameter type. So, if I have registered handlers for multiple `MessageWrapper<>` message types, they all get incorrectly dispatched to just one of my registered message handlers...and most fail during de-serialization. The other concern with this approach is that it doesn't support non-generic types that may have the same type name, but different namespaces. Since the simplistic `Type.Name` property is being used as the key, I expect there could be collisions if you attempted to register handlers for both `MyApp.Foo.Created` and `MyApp.Bar.Created`.

[...]

PR #357, opened 2018-06-07
 Support the publishing of Message attributes

jaimalchohan:
Allow SNS Message Attributes to be specified and published.
This is an iteration that will eventually enable message filtering as requested in #26

I referenced the AWS SNS in the Message base class, as I thought it was much cleaner to simply pass MessageAttributes around rather than implement an abstraction on top of this. This can be changed if there's disagreement.

### DIFF
--- a/JustSaying.Models.version.props
+++ b/JustSaying.Models.version.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup Condition=" '$(MSBuildProjectName)' == 'JustSaying.Models' ">
-    <VersionPrefix>5.0.0</VersionPrefix>
-    <!-- <VersionSuffix></VersionSuffix> -->
+    <VersionPrefix>6.1.0</VersionPrefix>
+    <VersionSuffix>beta1</VersionSuffix>
 
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">$(APPVEYOR_BUILD_NUMBER)</BuildNumber>
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">0</BuildNumber>
--- a/JustSaying.Models.version.props
+++ b/JustSaying.Models.version.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup Condition=" '$(MSBuildProjectName)' == 'JustSaying.Models' ">
-    <VersionPrefix>6.1.0</VersionPrefix>
-    <VersionSuffix>beta1</VersionSuffix>
+    <VersionPrefix>5.0.0</VersionPrefix>
+    <!-- <VersionSuffix></VersionSuffix> -->
 
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">$(APPVEYOR_BUILD_NUMBER)</BuildNumber>
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">0</BuildNumber>
--- a/JustSaying.Models.version.props
+++ b/JustSaying.Models.version.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup Condition=" '$(MSBuildProjectName)' == 'JustSaying.Models' ">
-    <VersionPrefix>5.0.0</VersionPrefix>
+    <VersionPrefix>5.1.0</VersionPrefix>
     <!-- <VersionSuffix></VersionSuffix> -->
 
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">$(APPVEYOR_BUILD_NUMBER)</BuildNumber>
--- a/JustSaying.Models/Message.cs
+++ b/JustSaying.Models/Message.cs
@@ -1,4 +1,6 @@
+using Amazon.SimpleNotificationService.Model;
 using System;
+using System.Collections.Generic;
 
 namespace JustSaying.Models
 {
@@ -20,6 +22,7 @@ namespace JustSaying.Models
         public string ReceiptHandle { get; set; }
         public string QueueUrl { get; set; }
         public int? DelaySeconds { get; set; }
+        public Dictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
 
         //footprint in order to avoid the same message being processed multiple times.
         public virtual string UniqueKey() => Id.ToString();
--- a/JustSaying.Models/Message.cs
+++ b/JustSaying.Models/Message.cs
@@ -1,4 +1,3 @@
-using Amazon.SimpleNotificationService.Model;
 using System;
 using System.Collections.Generic;
 
@@ -22,7 +21,7 @@ namespace JustSaying.Models
         public string ReceiptHandle { get; set; }
         public string QueueUrl { get; set; }
         public int? DelaySeconds { get; set; }
-        public Dictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
+        public IDictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
 
         //footprint in order to avoid the same message being processed multiple times.
         public virtual string UniqueKey() => Id.ToString();
--- a/JustSaying.Models/MessageAttributeValue.cs
+++ b/JustSaying.Models/MessageAttributeValue.cs
@@ -31,7 +31,7 @@ namespace JustSaying.Models
         /// data, or images.
         /// </para>
         /// </summary>
-        public MemoryStream BinaryValue { get; set; }
+        public byte[] BinaryValue { get; set; }
 
         /// <summary>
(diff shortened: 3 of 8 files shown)


stuart-lang:
I think it's important to keep `Message` clean of AWS types, as it can be used in lots of simple models contracts, it already a bit much asking people to reference `JustSaying.Models`.

I made a suggestion here around filtering which would require the attributes to be configured at a publisher level, rather than per publish, but I'm not sure how I feel about it:
https://github.com/justeat/JustSaying/issues/351#issuecomment-377028548

This is a design decision that needs solving soon, there is the concept of a "message envelope" where you have a type that exposed the raw AWK types, and also presents a property of T where T is your message, this has been with regards to handlers, but could also be used for publishing. Or another proposal is to have some sort of mapper, where you can configure what properties go in the body, vs attributes. We need to pick something soon.

JustSaying.Models/Message.cs
```diff
@@ -19,9 +21,10 @@ protected Message()
         public string Conversation { get; set; }
         public string ReceiptHandle { get; set; }
         public string QueueUrl { get; set; }
-        public int? DelaySeconds { get; set; }
-
-        //footprint in order to avoid the same message being processed multiple times.
+        public int? DelaySeconds { get; set; }
+        public Dictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
```
martincostello: Using `IDictionary` here would be nicer so the model isn't tied to a specific implementation.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -111,12 +112,14 @@ public void Publish(Message message)
         private PublishRequest BuildPublishRequest(Message message)
         {
             var messageToSend = _serialisationRegister.Serialise(message, serializeForSnsPublishing: true);
-            var messageType = message.GetType().Name;
+            var messageType = message.GetType().Name;
```
martincostello: Would this create funky names if someone was using a generic type?

jaimalchohan: @martincostello This is a whitespace change, I haven't actually changed anything here.

jaimalchohan:
> I think it's important to keep Message clean of AWS types, as it can be used in lots of simple models contracts, it already a bit much asking people to reference JustSaying.Models.

I agree, it would be much nicer to get rid of JustSaying.Models.Message, and my current feeling is that adding a further abstraction over MessageAttributes makes this harder, not easier. 

> I made a suggestion here around filtering which would require the attributes to be configured at a publisher level, rather than per publish, but I'm not sure how I feel about it:
#351 (comment)

I saw that and actually started with that as a solution, however I noticed difficulties in implementing it. And I'm strongly against it, but I'll comment there, not here!

> This is a design decision that needs solving soon, there is the concept of a "message envelope" where you have a type that exposed the raw AWS types, and also presents a property of T where T is your message, this has been with regards to handlers, but could also be used for publishing. Or another proposal is to have some sort of mapper, where you can configure what properties go in the body, vs attributes. We need to pick something soon.

Yeah, personally I would love to get rid of the generic types altogether. 

Anyway, this is the smallest, simplest and cleanest change I could make to implement the ability to support MessageAttributes.

JustSaying.Models/Message.cs
jaimalchohan: Hmmm ....

`Amazon.SimpleNotificationService.Model.PublishRequest.MessageAttributes` is tied to `Dictionary` so if we changed JustSaying.Models.Message.MessageAttributes to be an `IDictionary` we'd have to convert it to a `Dictionary` eventually.  What benefits do you see using `IDictionary` providing?

martincostello: It's just that if this is a simple contract/model that implementers are expected to derive from, using the interface gives them more flexibility, like how `IList<T>` is better than `T[]` or `List<T>`.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
martincostello: [I was right](https://github.com/justeat/JustSaying/issues/278#issuecomment-395370692) though 😉

stuart-lang:
Cool, we are agreed on most things 😃 This is simplest in terms of code, but not in terms of the implications. First because of the SNS package dependency for projects containing models. Secondly because it forces a decision on the message envelope discussion (@AnthonySteele your view?)

If we tweaked this so that we have our own type for MessageAttributeValue, I'd be in favour.

JustSaying.Models/MessageAttributeValue.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class MessageAttributeValue
+    {
+        /// <summary>
+        /// Gets and sets the property BinaryValue. 
+        /// <para>
+        /// Binary type attributes can store any binary data, for example, compressed data, encrypted
+        /// data, or images.
+        /// </para>
+        /// </summary>
+        public MemoryStream BinaryValue { get; set; }
```
martincostello: Could _maybe_ make this a `Stream` so it's the least derived type, unless there's things that would consume this that are dependent on the fact it's a stream backed by memory rather than something else.

jaimalchohan: Err, so in JustSaying we would have to convert the Stream into a memory stream for the AWS Implementation... presumably by reading the stream into a memory stream?

martincostello: Isn't this an output from AWS more than an input to AWS? If so, the reader just uses the stream, so the implementation of the stream doesn't matter.

JustSaying.Models.version.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup Condition=" '$(MSBuildProjectName)' == 'JustSaying.Models' ">
-    <VersionPrefix>5.0.0</VersionPrefix>
-    <!-- <VersionSuffix></VersionSuffix> -->
+    <VersionPrefix>6.1.0</VersionPrefix>
```
slang25: Given this is not a breaking change, it makes sense to keep this at 5 something. This package is deliberately versioned separately from the others.

JustSaying.Models/MessageAttributeValue.cs
slang25: The binary value that this represents can only be up to 256 bytes long, so I think Stream might convey the wrong idea.

I think `MemoryStream` is a little weird here, and a `byte[]` would be the most idiomatic thing here (I know this was copy&pasta from AWS but doesn't feel like we have to keep it the same).

jaimalchohan:
I'm literally just going to take your comments and turn them into code.

As you have strong opinions around justsaying that I can't possibly have, then perhaps you could make any appropriate changes and we can close this long feedback loop?

slang25:
😄 I'm not sure I have strong opinions, just like to discuss code changes like this. I'm happy to make tweaks to this PR.

One thing I'd need some input on from other is what we do about versions, currently there is a 6.0.0-beta1 out, that should really be released before moving ahead with a beta release of this.

Can those at Just Eat who are trailing this confirm it's suitable for release.

AnthonySteele:
I have no objection to this PR, two things to consider
* putting the attributes on the message base class is probably the right call for now. It's not perfect but until we make breaking changes to how / if this base class is used, it's where things like this metadata are stored.

* This only affects publishing? The attributes will still be there on the message base class when reading, should be populated then as well at that point?

* After the last long-lived V5 branch, I would be cautious about versioning to 5.1 while there's a 6.0 beta as well. We can do it, but if we do, have a plan to make it _very_ short lived and to keep  the branches in sync.

slang25:
Agreed on point 1 @AnthonySteele 👍 

Point 2, good question, at the moment it will not be populated, it could be done as I understand by asking for `All` attributes when doing the receive request, then populating it manually, or maybe explicitly requesting the attributes that are wanted. The thing with the `Message` base class is that it's already asymmetric in the way we use it today.

Point 3, I think the intention here was to release it as 6.1.0 (and `JustSaying.Models` as 5.1.0), which would follow the 6.0.0 rtm. So we should look to get that out soon-ish (or should it include whatever breaking change #358 requires?).

jaimalchohan:
MemoryStream > byte[] and Models version set to 5.1.0

AnthonySteele:
This is good to merge, but

1) For simplicity, version number of models assembly should be same an the main JustSaying version number.

2) We will merge this to a 6.0 beta unless there's a good reason to lock down 6.0 beforehand. So version number would be 6.0

Issue #358, opened 2018-06-08
NamingStrategy messageType


duncanbrown:
Consumers can customise topic/queue names on a per-message-type basis by providing an `INamingStrategy`. However, this is somewhat difficult to work with because the message type itself is not passed into the relevant methods, but rather a stringified representation (coming from https://github.com/justeat/JustSaying/blob/4050749a2b3cef9cdf566a21ba513ea5241c14ef/JustSaying/Extensions/TypeExtensions.cs#L12). Getting back from this string to the original type isn't trivial - especially in the case of generic types - which is problematic for consumers who want to base the aws resource name on the type in question.

I guess the simplest fix would be to add an additional `Type` parameter to the `INamingStrategy` methods, for the message's type. This would be a breaking change, but I notice there's a major version release in the pipeline anyway: would you be open to a pr along these lines?

AnthonySteele:
Seems reasonable.  Is there a PR for this?

You might even change the signature to `string GetTopicName(string topicName, Type messageType);` and the naming strategy could call `ToTopicName` on that type _if_ that was the desired behaviour.

duncanbrown:
No PR yet, but happy to put one together @AnthonySteele. On your suggestion of dropping the string and just using the type: that would suit my purposes, but I notice that the`ToTopicName` extension is currently in an internal class - would you be happy for that to be publically exposed, or is it considered a private implementation detail?

AnthonySteele:
You're right, it is internal. I would be happy to see that as a `public static` method. Extension or not is a matter of style, it doesn't make a big difference.

PR #359, opened 2018-06-12
Pass message type to INamingStrategy

duncanbrown:
Addresses #358 

- Updates `INamingStrategy` to take actual message `Type` rather than pre-stringified version.

- Updates `DefaultNamingStategy` to replicate existing behaviour 

- Exposes `TypeExtensions` publicly so that consumers can replicate existing behaviour similarly

n.b. this is a **breaking change**, not sure if anything needs to be done to signify that? It would probably be helpful for something to go into release notes telling consumers how to keep compatibility when they upgrade (by using `ToTopicName` in their naming strategies)?


AnthonySteele:
breaking change would happen at a major version, and be documented in the release notes.

AnthonySteele:
We will merge this to a 6.0 beta unless there's a good reason to lock down 6.0 beforehand.

PR #360, opened 2018-06-18
Support for generic messages

duncanbrown:
Addresses lingering issues with generic message types discussed @ #278.

As mentioned there, the problem seems to lie with bits of code that stringify types using `Type.Name`, which loses information about the instance type's generic parameters. An added complication is that the Subject of SQS/SNS messages gets populated with this value;  changing that could break existing applications.

Here, I've attempted to remove all usages of `Type.Name` from the codebase in a way that won't break existing applications:

* For purely internal stuff, I've just replaced uses of `Type.Name` with something that preserves information about generic type parameters.

* For the message Subject, I've introduced an `IMessageSubjectProvider`. The default implementation preserves the current behaviour and compatibility, but has the existing problems with generics. An optional `GenericMessageSubjectProvider` breaks compatibility but supports generics (or users are free to supply their own impelmentation).

### DIFF
--- a/JustSaying.TestingFramework/Tasks.cs
+++ b/JustSaying.TestingFramework/Tasks.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading.Tasks;
 using NLog;
 
@@ -8,7 +8,7 @@ namespace JustSaying.TestingFramework
     {
         private static readonly Logger Log = LogManager.GetCurrentClassLogger();
 
-        private const int DefaultTimeoutMillis = 5000;
+        private const int DefaultTimeoutMillis = 10000;
         private const int DelaySendMillis = 200;
 
         public static async Task<bool> WaitWithTimeoutAsync(Task task)
--- a/JustSaying/Extensions/TypeExtensions.cs
+++ b/JustSaying/Extensions/TypeExtensions.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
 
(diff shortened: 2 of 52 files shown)


AnthonySteele:
Is this PR separate to #359, or does it replace #359 ?

duncanbrown:
It's separate. The topic name stuff #359 is involved with already supports generics.

JustSaying/Messaging/MessageSerialisation/IMessageSubjectProvider.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace JustSaying.Messaging.MessageSerialisation
+{
+    public interface IMessageSubjectProvider
+    {
+        string GetTypeForSubject(Type messageType);
```
AnthonySteele: Should this be `GetSubjectForType` ?

JustSaying/Extensions/TypeExtensions.cs
```diff
@@ -15,18 +14,7 @@ public static string ToTopicName(this Type type)
...
-                name = name.Substring(0, MAX_TOPIC_NAME_LENGTH - suffix.Length) + suffix;
-            }
-
-            return name;
-        }
-
-        private static int GetInvariantHashCode(this string value)
-        {
-            return value.Aggregate(5381, (current, character) => (current*397) ^ character);
+            return name.TruncateTo(MAX_TOPIC_NAME_LENGTH);
```
AnthonySteele: There should be tests on this

AnthonySteele:
Needs unit tests on `GenericMessageSubjectProvider` and `NonGenericMessageSubjectProvider`.

After review, we will merge this to a 6.0 beta unless there's a good reason to lock down 6.0 beforehand.

JustSaying/Messaging/MessageSerialisation/IMessageSubjectProvider.cs
duncanbrown: Yes, yes it should :) Updated

JustSaying/Extensions/TypeExtensions.cs
duncanbrown: Have added some (though I haven't added any new functionality here btw - just moved the truncation code elsewhere so it can be reused)

duncanbrown:
Requested subject provider unit tests added

JustSaying.TestingFramework/Tasks.cs
```diff
@@ -8,7 +8,7 @@ public static class Tasks
     {
         private static readonly Logger Log = LogManager.GetCurrentClassLogger();
 
-        private const int DefaultTimeoutMillis = 5000;
+        private const int DefaultTimeoutMillis = 10000;
```
duncanbrown: Was getting some (unrelated to this pr) timeouts (locally and on ci)

duncanbrown:
Not sure why that test is failing on travis... passes locally and in appveyor, and I don't think anything I've done around the bit that's failing. Flaky test?

AnthonySteele:
Must be a flakey test, it's green after I re-ran it. 🤷‍♂️

PR #361, opened 2018-06-18
Switch to netstandard2.0

slang25:
Responding to https://twitter.com/davidfowl/status/1008453902058917888?s=19

I had been thinking about this for the past few months, and cannot decide what to do with the older targets.

I think for 6.0 it makes sense to target only `netstandard2.0`, which can run on full framework 4.6.1.

See individual commits for the interim states.


AnthonySteele:
Interesting. We're lookinag as similar here: https://github.com/justeat/JustEat.StatsD/pull/73

martincostello:
Looks like the change broke something:

```
C:\projects\justsaying\JustSaying.Tools\JustSaying.Tools.csproj : error NU1701: Package 'Magnum 2.1.3' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.
```

slang25:
The tools command line parser doesn't support netstandard/netcoreapp so I'm targeting `net461` now

JustSaying.Models/JustSaying.Models.csproj
```diff
@@ -1,9 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.6;net451</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
duncanbrown: Do you think the Models project/package should target netstandard2.0, given it only contains a single POCO? Seems like it could maybe even just target netstandard1.0?

slang25: That's a good point. It seems like this package should be as portable as possible.

martincostello: Makes sense to target 1.0 if it doesn't depend on anything else.

slang25:
I've targeted both in the models project, as it's preferable to have `netstandard2.0` present (reduces package explosion, improves build time etc...)

slang25:
Conclusion from the gitter chat:

There is no version of `Microsoft.Extensions.Logging.Abstractions` that targets full framework directly and `netstandard2.0`, so we cannot get away from having netstandard somewhere in the dependency tree, so there's not much point in targeting `net4x` directly, other than maybe in the Models package.

slang25:
Actually, there is one advantage to targeting `net461` directly, which is we could keep the sync `Publish` methods.

Do we want to remove those in this release or not?

slang25:
The thing with the sync `Publish` methods is that consumers using maybe old versions of OpenRasta don't want to be calling async methods. I think we have to drop support at some point.

I've done a quick test targeting `net451` vs `netstandard20`, and it's better package wise to drop `net451` when adding the package to a `net461` project.

Package with `net451` (references ye olde `Microsoft.Extensions.Logging.Abstractions` which is `netstandard1x`)
js-net451

Package with `netstandard20` and no `net451`:
js-netstandard20

AnthonySteele:
Agreed, we can to drop support for sync `Publish` at some point. We might be at that point, hope to make a decision on that later today.

It's an interesting result that the `netstandard20` version is simpler.
What happens if the package is built for net461, and is referenced by another net461 project?

slang25:
Provided we take version 2 of `Microsoft.Extensions.Logging.Abstractions` then we'd get the same sort of packages as with `netstandard2.0`, but what would be the advantage of that? We'd be including a `netstandard2.0` dependency anyway.

AnthonySteele:
After discussion today,

* We _need_ a `NetStandard 2.0` target for new Netcore and `net472` apps.
* We want (if possible) to keep a `net452` target for legacy app compatibility.

slang25:
👍 Good to see there is a work force in place now 😃 
This thing with targeting `net452` is that we have to use an older version (1.x) of `Microsoft.Extensions.Logging.Abstractions`, so we'd probably want an additional `net461` target that references version 2 of that package, along with a `netstandard2.0` target.

It's not a problem, it's just that we go from 1 target to 3 ¯\_(ツ)_/¯

tonyharverson:
@slang25 what's the motivation for wanting the new logging abstractions library? Ideally we'd want to avoid conditional compilation.

slang25:
@tonyharverson both versions target the different netstandards which is where are issue is. We'd be source compatible with both versions, but it wouldn't be ideal

slang25:
@AnthonySteele rebased and good to go :+1:

Issue #362, opened 2018-06-22
Support for SQS SSE (Server-side Encryption)


payman81:
Ability to create queues (or modify existing queues) with Server-side encryption by specifying the customer master key (CMK) ID.

payman81:
PR #363 submitted

martincostello:
Resolved by #363?

slang25:
Thanks @payman81! Hope you are doing well these days :smiley:

PR #363, opened 2018-06-22
Issue-362 Ability to configure subscriptions in order to enable serve…

payman81:
…r-side encryptions for SQS queues


slang25:
Fixes #362

PR #364, opened 2018-06-24
Update AWSSDK packages

AnthonySteele:
Update AWSSDK nuget packages.

With the new tree-structured package model, the `AWSSDK.*` nuget packages will be implicit dependencies of the app consuming `JustSaying` meaning that they are often consumed at the _lowest_ version specified here, and never updated. 

So update them here in order to get a current version to consumers of `JustSaying`.

* Use a floating version `3.3.*` in `JustSaying.csproj`
* Projects that reference `JustSaying.csproj` do not need to mention these `AWSSDK` packages again, they are pulled in from there, via this tree of dependencies. In fact, mentioning them again is nothing but an opportunity for things to get out of sync, so avoid it. Same with `Newtonsoft.Json`.

### DIFF
--- a/JustSaying.Tools/JustSaying.Tools.csproj
+++ b/JustSaying.Tools/JustSaying.Tools.csproj
@@ -11,7 +11,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Magnum" Version="2.1.3" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.1.1" />  </ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.1.1" />
+  </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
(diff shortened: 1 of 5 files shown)


slang25:
Would referencing it by floating version fix this going forwards?
https://docs.microsoft.com/en-us/nuget/consume-packages/dependency-resolution#floating-wildcard-versions

AnthonySteele:
Floating version could help! So would we float these to minor or patch? e.g. `3.*` or `3.3..*` ?

slang25:
I was thinking `3.3.*`

JustSaying.Tools/JustSaying.Tools.csproj
```diff
@@ -10,12 +10,8 @@
     <ProjectReference Include="..\JustSaying\JustSaying.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="3.3.10.3" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.3.1.11" />
     <PackageReference Include="Magnum" Version="2.1.3" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.1.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
-  </ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.1.1" />  </ItemGroup>
```
martincostello: Indenting snaffu.

PR #365, opened 2018-06-26
Update sourcelink

AnthonySteele:
Update the NuGet package `SourceLink.Embed.AllSourceFiles` to latest.
It's in the common file `Directory.Build.props`
And because the VS nuget tooling does not expect to find it in this file, you have to do this update with a text editor not via clicking buttons.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -24,6 +24,6 @@
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(IsTestProject)' != 'true' AND '$(NCrunch)' == '' AND '$(SourceLinkEnabled)' != 'false' ">
-    <PackageReference Include="SourceLink.Embed.AllSourceFiles" Version="2.8.1" PrivateAssets="All" />
+    <PackageReference Include="SourceLink.Embed.AllSourceFiles" Version="2.8.3" PrivateAssets="All" />
   </ItemGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -21,9 +21,7 @@
     <ComVisible>false</ComVisible>
     <IsTestProject>$(MSBuildProjectName.Contains('Tests'))</IsTestProject>
     <IsTestProject Condition=" '$(IsTestProject)' != 'true' ">$(MSBuildProjectName.Contains('Testing'))</IsTestProject>
+    <EmbedAllSources Condition=" '$(IsTestProject)' != 'true' AND '$(NCrunch)' == '' AND '$(SourceLinkEnabled)' != 'false' ">true</EmbedAllSources>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


AnthonySteele:
I still don#t know how sourcelink works. If that's the way forward then I will withdraw this PR.

slang25:
I have a blog post that explains what we are doing here, but I haven't gotten around to publishing it 😞 
I personally like the embedded source approach here, but it's all subjective, there's no "better" approach. This repo is doing embedded source, so there's no download step, but it makes the package marginally larger than it needs to be.

slang25:
Looking at this again, we can now remove this package and add the property `<EmbedAllSources>true<EmbedAllSources>` to the `csproj` as it's natively supported by the compiler.

AnthonySteele:
I am not sure that this is working, if I put `<EmbedAllSources>false</EmbedAllSources>` or `<EmbedAllSources>tue</EmbedAllSources>` in the `Directory.Build.props` file, it makes no difference to the dll size.

Also If I reference the `JustSaying.dll` in another project with the `true` setting, I can't step into it.

slang25:
You're right, I'm not quite sure at the moment why it's not working, I'll do some digging.

slang25:
@AnthonySteele Found what it was, missing `<DebugType>embedded</DebugType>`. I rebased this PR and force pushed without thinking, apologies for the poor etiquette, shouldn't be force pushing in your branch.

slang25:
image

martincostello:
Nice work on getting to the bottom of it, Stu. How much smaller would it be if we used the GitHub version that just points there for the source rather than embedding it all in?

slang25:
110KB, so 6-7% larger by embedding

AnthonySteele:
So now the debug build of `JustSaying.dll` is 184Kb instead of 119Kb.
My test of if I can step into it still doesn't work, but that might be `JustSayingStack` between them getting in the way.

slang25:
Hmm, odd. What version? 5.0.0?

slang25:
Or just a local build of this branch?

AnthonySteele:
It was just a local build of this branch, yes.

slang25:
Odd, there are 2 ways (at least) to confirm that the sources are embedded. Extract the dll, and view it with DotPeek as I've shown above, looking for those nodes.
Or run
```
sourcelink print-documents .\JustSaying.dll
```
using this global tool:
```
dotnet tool install --global sourcelink
```

slang25:
If it still doesn't work then try checking this: (although I hope it's not required)
image

AnthonySteele:
- Checking that box makes no difference to me.
- dotPeek verifies that the source is there, so lets ,merge this and work out the issues later.
- The issue are probably to do with stepping into `JustSayingStack` first, it might need this source embedding as well?

PR #367, opened 2018-07-02
Add basic support for SQS Subscription FilterPolicies

jaimalchohan:
Message Filtering Docs:https://docs.aws.amazon.com/sns/latest/dg/message-filtering.html

The filter policy (string) is supposed to be JSON.

I refactored the subscription code a little as we appeared to have remove code that was dependant on the obtuse implementation 

The the [docs](https://docs.aws.amazon.com/sns/latest/dg/message-filtering.html) there is a Java sample helper implementation that constructs the JSON FilterPolicy from a HashMap. Is this helper something we'd like to introduce to JustSaying, or should we leave it to the user to construct the JSON?


slang25:
I would expect JustSaying to provide something, however if it just takes a string it's something we could add later.

jaimalchohan:
I agree, Modelling the Policies using something type safe, the produces the correct syntax would be really nice, and whilst I have stated working on this, it requires a bit more testing.  I'd think it'd be better to release that as a non-breaking change with a minor revision change rather than creating a dependency between this PR and it.

JustSaying/AwsTools/QueueCreation/AmazonQueueCreator.cs
```diff
@@ -68,9 +74,14 @@ private static bool TopicExistsInAnotherAccount(SqsReadConfiguration queueConfig
...
-        private Task<bool> EnsureQueueIsSubscribedToTopic(SnsTopicByName eventTopic, SqsQueueByName queue)
-        {
-            return eventTopic.SubscribeAsync(queue);
+        async Task SubscribeQueueAndApplyFilterPolicyAsync(IAmazonSimpleNotificationService amazonSimpleNotificationService, string topicArn, IAmazonSQS amazonSQS, string queueUrl, string filterPolicy)
+        {
+            var subscriptionArn = await amazonSimpleNotificationService.SubscribeQueueAsync(topicArn, amazonSQS, queueUrl).ConfigureAwait(false);
+
+            if (!string.IsNullOrEmpty(filterPolicy))
+            {
+                await amazonSimpleNotificationService.SetSubscriptionAttributesAsync(subscriptionArn, "FilterPolicy", filterPolicy);
```
martincostello: Does this need a `ConfigureAwait(false)` too?

PR #370, opened 2018-07-03
version metadata for 6.0.0 beta 2

AnthonySteele:
We are hoping to release another beta in a few days.
As far as I know, this is the version metadata necessary? i.e. everything is version `6.0.0 beta 2`


slang25:
My thoughts on the separate package versions are that they are 2 distinct packages, and we shouldn't be updating the models package that often. So controlling this version separately reduces package version proliferation in the wild, it's a small burden here vs a larger burden to consumers.

That said, I don't feel that strongly about it, and happy to hear your thoughts.

AnthonySteele:
I see the point, though we do have a change to the Message class [here](https://github.com/justeat/JustSaying/pull/357) and we should indicate that this version of the Models package works with JustSaying v6.

slang25:
I'm not great with semver, but wouldn't that be a minor release, as it's not breaking for current consumers?

AnthonySteele:
KISS. In general the messaging that "JustSaying v6 matches JustSaying.Models v5.1" is IMHO imposing too much cognitive load and reading of docs on consumers. People will understand what matches if they both have version 6.0. They likely won't understand anything else.

PR #371, opened 2018-07-17
Renamed JustSayingFluentlyLogging and JustSayingFluentlyDependencies

adammorr:
Since adding `MessageSubjectProvider` to the class, it feels appropriate to rename `JustSayingFluentlyLogging` to `JustSayingFluentlyDependencies`

### DIFF
--- a/JustSaying/CreateMeABus.cs
+++ b/JustSaying/CreateMeABus.cs
@@ -11,7 +11,7 @@ namespace JustSaying
         /// </summary>
         public static Func<IAwsClientFactory> DefaultClientFactory = () => new DefaultAwsClientFactory();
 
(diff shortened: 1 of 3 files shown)


AnthonySteele:
Would this go into 6.0 ?

martincostello:
This would need to be a candidate for `6.0.0` if we're going to do this rename, or we'll be doing another breaking change for `7`.

AnthonySteele:
I am OK for this in 6.0 - renaming is not a risky change operationally, although it can  break consumers when compiling.

adammorr:
OK does that mean we're good to merge this now or are we waiting on something else?

adammorr:
@AnthonySteele @martincostello any objection to me merging this?

Issue #380, opened 2018-08-18
Stop using JustBehave


martincostello:
As part of the investigation for #159 I've found that there's _a lot_ of use of `Task.GetAwaiter().GetResult()` and `Task.Wait()` to work around limitations with JustBehave's design of running test setup and teardown steps on synchronous code paths for constructors and disposal.

JustSaying is heavily asynchronous now, so this just profligates poor async usage throughout the tests.

I also think it is also causing deadlocks when I try and run them locally en-masse using the Visual Studio test runner with a local AWS simulator. This is a hunch, but it mainly comes from the fact that if I run all the tests they seem to get stuck, but if I run tests individually they pass.

The pattern of calling derived methods from constructors in JustBehave also prevents `ITestOutputHelper` (and other things) from being injected into the tests and being used as part of the test setup as the class' own constructor logic runs after the overridden classes, meaning any fields don't have their values set yet.

I'm happy to do the work to refactor this out of the integration test project myself as part of #159 if we agree that it's a necessary step to move the test suite forward.

I think the time has come to put it out to pasture for JustSaying and adopt a more conventional and xunit-friendly test style using fixtures.

martincostello:
Soon...

JosephWoodward:
You've got my vote 👍

martincostello:
Man, this is going to be a mission....

martincostello:
Had a quick hack-around, but it's usage is so pervasive that getting rid of it basically means slowly re-writing all the tests from scratch, deleting the old-style tests as they're switched, and then removing the JustBehave reference at the end once they're all done.

I think we'd all have to tackle a bit each over several weeks to scrub it entirely, short of deleting _everything_ and then starting from scratch, which doesn't seem the wisest idea, though could maybe done slowly in a long-lived branch with the "big bang" happening at the end with it being merged to master in one go.

martincostello:
Because JustBehave is not strong-named, we can't strong-name JustSaying until we remove it (see #405).

shaynevanasperen:
@slang25 Because of `InternalsVisibleTo`

slang25:
Damn, did not know that! We could introduce a configuration for "release but not signed", if running the tests in debug isn't desirable. It's a bit grim though 😞

Seems like yet another reason to overhaul the tests. The coverage is inconsistent, JustBehave forces bad async, we can't do strong naming... the list is long 😆

slang25:
[...]

Issue #159, opened 2016-02-08
Should run the Integration Tests as part of the continuous integration build


brainmurphy:
These tests give us greater confidence in the changes we make. There is an issue in that they require a functioning AWS environment to execute.

slang25:
We should definitely do something around this, cause the integration tests are constantly breaking as a result of this.
I have a branch locally that is working against [localstack](https://github.com/localstack/localstack), which works for most of the tests, but not for cross region tests, or iam policy checks. We could run that on travis with docker without too much effort, although it would be good to have a dedicated aws to run this in properly too.

martincostello:
I think this is one we shouldn't let go stale 😉

martincostello:
As discussed, I'm going to pick up looking at this for both real end-to-end tests with real AWS resources, as well as investigating using [localstack](https://github.com/localstack/localstack) for inner-loop unit/functional tests.

Once I've got a PoC for it sorted out, I'll raise a PR and we can go through it and decide how to proceed.

duncanbrown:
Just a random thought - might be worth considering goaws (https://github.com/p4tin/goaws) as an alternative to localstack: I've recently been using it with JustSaying for local development, and haven't had any issues yet (only been doing pretty simple stuff, though). I started using it as it seems much lighter-weight than localstack - docker image is just 12MB, and it runs in just a few MB of memory.

martincostello:
Cool, thanks @duncanbrown. I'll take a look at that too!

martincostello:
I've started having looking into this.

martincostello:
Begun in #381.

martincostello:
#381 has been merged which adds support for using goaws (thanks @duncanbrown 👍 ) during the Travis CI builds, as well as on Windows with `Build.ps1 -EnableIntegrationTests` if you have Docker installed and configured for Linux containers.

A future PR will validate the tests work with environment variables for real AWS resources, as well as a way of running the Windows build with integration tests in AppVeyor somehow given the free/OSS plan doesn't support Linux containers.

shaynevanasperen:
@martincostello The AppVeyor free/OSS plan does support Linux containers now.

martincostello:
Windows containers in Docker do not support Linux VMs on the free plan: https://github.com/justeat/JustSaying/pull/522#issuecomment-466799826

martincostello:
I've created an internal Jira to track getting us access to two real AWS accounts for doing testing with real AWS resources.

[...]

PR #381, opened 2018-08-18
Integration tests

martincostello:
Apologies in advance for the size of the diff.

This PR does _quite a lot_ of refactoring to support running the integration tests in CI using [goaws](https://github.com/p4tin/goaws).

It also _in theory_ supports running them against a real AWS account, but I haven't tested that yet.

A lot of the refactoring relates to how the region, logger factory etc. usage was spread all over the tests, making it hard to centralise into a single place to reconfigure. I've shifted a lot of things around to try and make it more-or-less changeable in a single place. This would allow, for example, a developer in Australia to run the integration tests against a real AWS account with `ap-southeast-2` as the primary region instead of `eu-west-1`.

I've also, where possible, hooked the logger factory into xunit's `ITestOutputHelper` so the logs get emitted into the test results, so you can read them if the tests fail. At a later point this could be extended into the uni tests as well.

There were also issues with tests hanging (see #380), though I've worked around these issues with usage of `ConfigureAwait(false)` in the paths for constructors that do async things.

Further, I've simplified the test project dependencies to leverage transient dependencies and reduce duplication, as well as target `netcoreapp2.1`.

Relates to #159.


martincostello:
Looks like Linux containers aren't supported in the free version of AppVeyor 😞

README.md
```diff
@@ -13,9 +13,9 @@ Before you can start publishing or consuming messages, you want to configure the
 			    new DefaultAwsClientFactory(new BasicAWSCredentials("accessKey", "secretKey"))
 ````
 
-You will also need to create a `ILoggerFactory` ([see here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging)), if you do not want logging then you can always create an empty logger factory like so:
+You will also need a `ILoggerFactory` ([see here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging)); if you do not want logging then you can use an empty logger factory like so:
 ````c#
-        var loggerFactory = new LoggerFactory();
+    ILoggerFactory loggerFactory = NullLoggerFactory.Instance;
```
slang25: TIL 👍

slang25:
Yeah, the deadlocks before were due to the blocking. It was due to the xunit enforcing a test concurrency limit, this introduced a SynchronizationContext that would lock up, often the tests would pass in isolation, but a full run would almost never work.

The work I did in JustBehave was primarily a pre-req for sorting out these tests, that said, I would like to see it gone from JustSaying completely.

On goaws vs localstack. I got about 80% of tests running with localstack, cause it also emulates iam etc... That said, goaws looks much nicer, and I'd be far more likely to PR to that than localstack.

This PR rocks, let me know if I can help.

slang25:
Shame about AppVeyor and linux images on docker ([related support question](https://help.appveyor.com/discussions/questions/8194-linux-docker-containers)), can we do it with AppVeyor linux build agent images or Travis?

martincostello:
My initial thinking yesterday was maybe we could run goaws in Travis (it's working now 🎉) and once I've done some local testing with it, we could setup AppVeyor to do the "real" integration tests as a "workaround".

Then we've got PRs testing both with the local emulator on Linux and against real SNS and SQS on Windows?

It's not ideal that we've got different coverage on different OSs, but it would spread things around, at least for an initial merge.

My initial aim was to get this at least green and in a decent working state for a jumping off point for merging to master, then we could do any further improvements on top of that (extending coverage, maybe removing JustBehave etc.).

martincostello:
Fun fact: dynamically changing the build Id in the build makes the pending GitHub status for AppVeyor 404:

```
Expected: https://ci.appveyor.com/project/justeattech/justsaying/build/0.0.386-utfjxwpu
Actual: https://ci.appveyor.com/project/justeattech/justsaying/build/6.0.0-beta2-build386-d5lx2cw2
```

martincostello:
I've got far enough along in this PR I think that there's enough value there to merge it now even though the overall task in #159 isn't yet complete.

Merging this adds integration tests with goaws for Travis CI, plus locally via opt-in in `Build.ps1` if you have Docker installed for Linux containers.

In a follow-up PR I'll do work to ensure we also support running the same test suite against AWS itself if the appropriate environment variables for keys etc. are configured.

martincostello:
This also gets enough of the refactoring into master to allow others to work on #380.

slang25:
This is amazing! 🎉 I bought Paul a coffee to say thanks for goaws

Issue #382, opened 2018-08-28
.NET Framework Support


slang25:
In the current 6.0.0 beta support for `net451` was dropped in favour of `netstandard20` which we understood to provide support for .NET 4.6.1 and up, however recently there is increasing consensus that .NET Standard 2.0 support for .NET Framework below 4.7.2 is problematic.

It might be a good idea to in this case explicitly multi-target 4.6.1, or alternatively make it clear that support below 4.7.2 will be non-ideal and possibly problematic.

Thoughts?

martincostello:
My opinion would be to go with adding a `net461` target back.

slang25:
Mine too, it was dropped because it felt a bit gross straddling 2 major versions of the logging abstractions package.

See here:
https://twitter.com/stuartblang/status/1034427988224421888

AnthonySteele:
We have to support .NET framework  `4.6.x` at present for a large number of codebases, so adding it back it might be necessary.

slang25:
We are back to this:
https://github.com/justeat/JustSaying/pull/361#issuecomment-399457168

😐

martincostello:
I guess fixing this is a must for 6.0.0 now?

AnthonySteele:
I think so, yes. We need a second target, be it `4.6.2` or `4.5.1`

martincostello:
I thing based on it being a major bump since 5 and 4.5.1 being out of support (?) now, we should go with `net462`.

PR #383, opened 2018-08-30
Bring back the .NET461 target

AnthonySteele:

Bring back the .NET461 target

https://github.com/justeat/JustSaying/issues/382#issuecomment-417276372

### DIFF
--- a/JustSaying.Models/JustSaying.Models.csproj
+++ b/JustSaying.Models/JustSaying.Models.csproj
@@ -1,5 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netstandard1.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard1.0;net461</TargetFrameworks>
   </PropertyGroup>
 </Project>
--- a/JustSaying/JustSaying.csproj
+++ b/JustSaying/JustSaying.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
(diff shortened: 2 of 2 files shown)


martincostello:
Do we need to do any malarkey with the logging abstractions package here @slang25 ?

AnthonySteele:
As it stands, it will pull in `Microsoft.Extensions.Logging.Abstractions` v2.1.1 which *only* has a NetStandard2.0 build.

The alternative is to conditionally use `v1.1.2`, which will use `.NETStandard 1.1`

slang25:
I have this:
https://github.com/justeat/JustSaying/compare/master...slang25:net461

I reference the targeting pack so it can be built on macOS (or any non-windows, non-mono setup)
Not sure what the right thing to do with the logging package to be honest.

AnthonySteele:
Could we go with the following and leave targeting packs for a separate pr?
```xml
  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.2" />
  </ItemGroup>
  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.1" />
  </ItemGroup>
```

slang25:
We could, some people on Hipchat were concerned when I did this during the netstandard2.0 update which is why I took it out, but I think it makes sense knowing what we know now.

slang25:
This situation feels familiar 🤔 

> net461: Okay, relax @slang25, it's me, it's net461!
> @slang25: No it can't be you... I just converted you to dropped you in favour of netstandard2.0!
> net461: No, I know, you did drop me, but I'm back!
> @slang25: Great Scott!

image

martincostello:
Martin McFly approves: ✔️

_Gets in DeLorean_ 🚗

PR #386, opened 2018-08-31
[DNM] Proof-of-Concept HttpClientFactory integration

martincostello:
A proof-of-concept of integrating HttpClientFactory into JustSaying.

This isn't intended to be merged, it's just an easy way to make it easy to share for review and consideration.

It's all very unextensible and what-have-you in this form, but there's enough here to show the concept of how we could hook newer .NET Core 2.1 concepts into the AWS SDK and JustSaying without having to touch the SDK itself.

The same concept could be used with the Dynamo and S3 clients as well.

### DIFF
--- a/JustSaying.Extensions/AmazonServiceClientExtensions.cs
+++ b/JustSaying.Extensions/AmazonServiceClientExtensions.cs
@@ -1,0 +1,32 @@
+using System.ComponentModel;
+using System.Net.Http;
+using Amazon.Runtime;
+using Amazon.Runtime.Internal;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace JustSaying.Extensions
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    internal static class AmazonServiceClientExtensions
+    {
+        internal static void ConfigureHttpHandler<T>(this T client, RuntimePipeline pipeline)
+            where T : AmazonServiceClient
+        {
+            string name = client.GetType().Name;
+
+            var services = new ServiceCollection();
+
+            services
+                .AddHttpClient()
+                .AddAwsHttpClient(name, client.Config);
+
+            var serviceProvider = services.BuildServiceProvider();
+            var httpClientFactory = serviceProvider.GetRequiredService<IHttpClientFactory>();
+
+            var httpRequestFactory = new CustomHttpRequestFactory(name, client.Config, httpClientFactory);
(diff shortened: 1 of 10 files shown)


slang25:
This is a great idea, this would allow you to get some logging handler middleware in there, at the moment AWS SDK for .NET Core lacks any ability to enable any kind of logging (https://github.com/aws/aws-sdk-net/pull/915)

martincostello:
Yeah, the whole load of HTTP handler stuff we already have (StatsD, logging) would a slot straight in.

JosephWoodward:
Nice!

ilike

AnthonySteele:
Would this help us diagnose slow NSN publishes? 

We know that a slow publish stalls somewhere in the call stack down near

```
   at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
```

It looks like the async awaits are fine, no task blocking issues.

It could be the `HttpRequestMessageFactory` with the cache and locks, and we might be able to update that code since it's not in the `NET35` build, but we don't know if that code is causing the problem, or is even used in that path.
It could be a connection cached longer than it should be after DNS has changed. 

Both of these are fixable, but we don't know if it's either of them, or something else.

slang25:
It can only help. It would certainly help us rule out JustSaying from the problem and let you measure things at the HTTP level.

martincostello:
We can discuss this in the next JustSaying "Funday Friday" we have once I'm back in the office (21st?) and report back with some suggestions on steps forward, if any, for JustSaying itself.

Of course, this PR just shows how it _could_ be done, so any temporary need to instrument things like @AnthonySteele mentions could be done ad-hoc in any apps needing to do that, without requiring JustSaying itself to change(?).

martincostello:
Closing.

martincostello:
I’ve folded the ideas here into an internal library for now to try it out in one of my team’s apps with the S3 SDK to see what mileage it gets in production. I’ve got at least enough proof-of-concept for integrating with Dynamo, S3, SNS and SQS.

Assuming it gives some non-negligible gains, I’ll look into making it into its own OSS library as some building blocks for wider use.

Unfortunately the way the AWS SDK is designed doesn’t lend itself to a simple way to drop this in, so (at least with where I got to so far) a bit of bespoke boilerplate would always be needed in the consuming app to make it truly re-usable.

PR #391, opened 2018-09-17
Add BaseTopicName to SnsWriteConfiguration

adammorr:
Currently the `BaseTopicName` is taken from the `_subscriptionConfig` when registering an SNS publisher which means you have to perform some fluent API gymnastics to set it correctly, such as;
```
CreateMeABus
    .WithLogging(NullLoggerFactory.Instance)
    .InRegion("eu-west-1")
    .WithSqsTopicSubscriber()
    .IntoQueue("queue")
    .ConfigureSubscriptionWith(cfg =>
    {
        cfg.BaseTopicName = "lardy-dah";
    })
    .WithMessageHandler(new FakeMessageHandler())
    .WithSnsMessagePublisher<FakeMessage>();
```

This proposed change allows you to do this;
```
CreateMeABus
    .WithLogging(NullLoggerFactory.Instance)
    .InRegion("eu-west-1")
    .WithSnsMessagePublisher<FakeMessage>(cfg =>
    {
        cfg.BaseTopicName = "lardy-dah";
    });
```

### DIFF
--- a/JustSaying.IntegrationTests/JustSayingFluently/GivenANotificationStack.cs
+++ b/JustSaying.IntegrationTests/JustSayingFluently/GivenANotificationStack.cs
@@ -84,11 +84,15 @@ namespace JustSaying.IntegrationTests.JustSayingFluently
                     c.PublishFailureBackoffMilliseconds = _config.PublishFailureBackoffMilliseconds;
                     c.PublishFailureReAttempts = _config.PublishFailureReAttempts;
                 })
-                .WithSnsMessagePublisher<SimpleMessage>()
+                .WithSnsMessagePublisher<SimpleMessage>(cf =>
+                {
+                    cf.BaseTopicName = "this-is-my-topic";
+                })
                 .WithSqsTopicSubscriber()
                 .IntoQueue(TestFixture.UniqueName)
                 .ConfigureSubscriptionWith(cf =>
                 {
+                    cf.BaseTopicName = "this-is-my-topic";
                     cf.MessageRetentionSeconds = 60;
                     cf.VisibilityTimeoutSeconds = JustSayingConstants.DEFAULT_VISIBILITY_TIMEOUT;
                     cf.InstancePosition = 1;
--- a/JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
+++ b/JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using JustSaying.Models;
 
 namespace JustSaying.AwsTools.QueueCreation
@@ -10,5 +10,6 @@ namespace JustSaying.AwsTools.QueueCreation
         /// </summary>
         /// <returns>Boolean indicating whether the exception has been handled</returns>
         public Func<Exception, Message, bool> HandleException { get; set; }
+        public string BaseTopicName { get; set; }
     }
 }
--- a/JustSaying/DefaultNamingStrategy.cs
+++ b/JustSaying/DefaultNamingStrategy.cs
@@ -5,7 +5,7 @@ using JustSaying.Extensions;
 namespace JustSaying
 {
     /// <summary>
-    /// A default namign strategy for JustSaying bus.
+    /// A default naming strategy for JustSaying bus.
     /// Topic names are defaulted to message type name, lowercase (one topic per message type).
     /// Queue name is default to queue name.
     /// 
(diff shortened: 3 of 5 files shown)


JustSaying/JustSayingFluently.cs
```diff
@@ -75,12 +75,11 @@ public virtual INamingStrategy GetNamingStrategy()
             var snsWriteConfig = new SnsWriteConfiguration();
             configBuilder?.Invoke(snsWriteConfig);
 
-            _subscriptionConfig.Topic = typeof(T).ToTopicName();
             var namingStrategy = GetNamingStrategy();
 
             Bus.SerialisationRegister.AddSerialiser<T>(_serialisationFactory.GetSerialiser<T>());
 
-            var topicName = namingStrategy.GetTopicName(_subscriptionConfig.BaseTopicName, typeof(T));
+            var topicName = namingStrategy.GetTopicName(snsWriteConfig.BaseTopicName, typeof(T));
```
slang25: I've always found the whole `Base...Name` stuff a bit weird, so excuse my ignorance. The value we are no longer using (`_subscriptionConfig.BaseTopicName`) used to be populated by `WithSqsTopicSubscriber` which took an optional topic name, how does that play with this after the change?

adammorr: The behaviour of `WithSqsTopicSubscriber` remains the same with this change as `_subscriptionConfig.BaseTopicName` will continue to be used when configuring a topic subscriber (https://github.com/justeat/JustSaying/blob/master/JustSaying/JustSayingFluently.cs#L339)

JustSaying.IntegrationTests/JustSayingFluently/GivenANotificationStack.cs
```diff
@@ -84,11 +84,15 @@ protected override IAmJustSayingFluently CreateSystemUnderTest()
...
-                .WithSnsMessagePublisher<SimpleMessage>()
+                .WithSnsMessagePublisher<SimpleMessage>(cf =>
+                {
+                    cf.BaseTopicName = "this-is-my-topic";
+                })
                 .WithSqsTopicSubscriber()
                 .IntoQueue(TestFixture.UniqueName)
                 .ConfigureSubscriptionWith(cf =>
                 {
+                    cf.BaseTopicName = "this-is-my-topic";
```
martincostello: Should this test set a different name in step 2 to verify that it overwrites the one from step 1?

JustSaying/AwsTools/QueueCreation/SnsWriteConfiguration.cs
```diff
@@ -9,6 +9,7 @@ public class SnsWriteConfiguration
         /// Extension point enabling custom error handling on a per notification basis, including ability handle raised exceptions.
         /// </summary>
         /// <returns>Boolean indicating whether the exception has been handled</returns>
-        public Func<Exception, Message, bool> HandleException { get; set; }
+        public Func<Exception, Message, bool> HandleException { get; set; }
+        public string BaseTopicName { get; set; }
```
martincostello: Needs `///` docs.

JustSaying/DefaultNamingStrategy.cs
```diff
@@ -5,19 +5,22 @@
...
     /// 
     /// Such configuration gives a queue for each IntoQueue configuration , and a queue is subcribed to multiple topics, where one topic per message.
     /// </summary>
     class DefaultNamingStrategy : INamingStrategy
     {
-        public string GetTopicName(string topicName, Type messageType) => messageType.ToTopicName().ToLower();
+        public string GetTopicName(string baseTopicName, Type messageType)
+            => string.IsNullOrEmpty(baseTopicName)
+                ? messageType.ToTopicName().ToLower()
+                : baseTopicName;
```
shaynevanasperen: Shouldn't this be concatenating the base topic name with the message type?

JustSaying.IntegrationTests/JustSayingFluently/GivenANotificationStack.cs
shaynevanasperen: Why have two ways of doing the same thing?

martincostello: I assume the new method is a "quick and simple" way, and the other is the "advanced" way.

JustSaying/DefaultNamingStrategy.cs
adammorr: I've just mirrored what we're doing in `GetQueueName(...)` below when a `BaseQueueName` is provided. If we change this, we should change both.

Back to Stu's original point, the whole base stuff is a bit weird (as well as the configuration syntax) so it could be a good thing to revisit when we plan JustSaying 7

shaynevanasperen: I see. However, from looking at this change alone (not digging any deeper) they both look wrong.

adammorr: Yea, it's a bit weird as base implies part of something not the actual "thing"

adammorr:
@martincostello @shaynevanasperen @AnthonySteele @stuart-lang I'm going to close this now that 6 is released fully and instead do it properly when we rewrite the fluent API for V7

PR #395, opened 2018-09-22
Bump version to 7.0.0

martincostello:
* Bump the version to 7.0.0.
  * Set the initial version in AppVeyor to `7` rather than `0`, even if we do change it during the build (which breaks pending GitHub status check links once the build starts).


martincostello:
I was under the impression we'd agreed that we'd keep the versions in-step to prevent confusion, but yes in that case we could consolidate.

martincostello:
Pending the `6.0.1` release now.

martincostello:
This can go in whenever now once ✔️'d. We can refactor the targets to remove duplication separately.

PR #396, opened 2018-09-28
Microsoft.Extensions.Logging.Abstractions dependency version downgrade

duncanbrown:
Changes the dependency on Microsoft.Extensions.Logging.Abstractions for the netstandard2.0 target from >= 2.1.1 to >= 2.0.0.

This is useful because the Microsoft.AspNetCore.App metapackages have dependencies on specific versions of Microsoft.Extensions.Logging.Abstractions. So as things stand, if you're targeting, say, Microsoft.AspNetCore.App 2.1.0 and JustSaying, you'll get errors unless you explicitly reference Microsoft.Extensions.Logging.Abstractions 2.1.1, and even then you'll get a warning.

I assume that JustSaying isn't actually depending on anything in 2.1.1, because the net461 target uses 1.1.2?

### DIFF
--- a/JustSaying/JustSaying.csproj
+++ b/JustSaying/JustSaying.csproj
@@ -15,6 +15,6 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.2" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
   </ItemGroup>
 </Project>


AnthonySteele:
Seems reasonable. No. we don't have a need for the very latest version of `Microsoft.Extensions.Logging.Abstractions` and can be more lax about the required version in order to allow these consumers here.

However it would be different if it was an `AWSSDK.`  package, for instance

martincostello:
I get the motivation for this, but two caveats:
  1. Is this a breaking change in a "strict" sense? To me it seems fine, but I'm no expert. Would this be OK as far as Semantic Versioning is concerned as being a `6.0.1`?
  2. If we're going to reduce this version, we might as well reduce their others to the minimum acceptable versions for compilation as well. Then we leave it to consumers to manage their versions (_including_ the AWS SDK), rather than mandating versions ranges based on what we (Just Eat) want as the minimums in our applications?

If memory serves me, we noticed a while ago that these had been over-eagerly updated to newer versions, but we left them as is as we weren't sure on the SemVer consequences so left them as-is.

AnthonySteele:
1 ) I don't _think_ that the lower version requirement on logging is a breaking change, as everything (combo of package versions) that was allowed before is allowed afterwards.

2) Seems reasonable, and in scope for next point release, but may be out of scope for this PR.

martincostello:
I'll merge this and do an imminent PR to do the others.

duncanbrown:
Quick turnaround! Thanks :)

martincostello:
@duncanbrown This has been pushed to NuGet in [6.0.1](https://www.nuget.org/packages/JustSaying/6.0.1).

PR #397, opened 2018-09-28
Versioning Updates

martincostello:
As-per the discussion in #396, this PR updates our versioning as thus:
  1. JustSaying now targets the minimum NuGet version required for compilation for it's dependencies, _except_ the AWS SDK, which I've just pinned to the same versions that we shipped `6.0.0` with.
  1. Updates the other projects to specify the _latest_ versions available for the transient dependencies from JustSaying.
  1. Updates some direct test dependencies to the latest relevant versions.
  1. Bumps the package version to `6.0.1` so we can ship these updates to NuGet.

### DIFF
--- a/JustSaying.Tools/JustSaying.Tools.csproj
+++ b/JustSaying.Tools/JustSaying.Tools.csproj
@@ -9,6 +9,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Magnum" Version="2.1.3" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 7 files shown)


JustSaying.Tools/JustSaying.Tools.csproj
```diff
@@ -9,6 +9,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Magnum" Version="2.1.3" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
```
AnthonySteele: Should these versions also be lower?

martincostello: Do we ship this? I thought it was an executable tool not a library, so would want the latest versions.

AnthonySteele: Ah, that's correct, it's not referenced by `JustSaying`

martincostello:
We should look at publishing a `6.0.1` with these changes and those from #396 on Monday.

PR #403, opened 2018-10-23
Dispose of the CancellationTokenSource in SqsNotificationListener

adammorr:
Disposes of the `CancellationTokenSource` that is used in the `ListenLoop` in `SqsNotificationListener`


martincostello:
From looking at the code, it seems more that we should be implementing `IDisposable` on this type so it's disposed of for other code paths as well, as it's owned by the class, rather than disposing of it on a specific call path?

adammorr:
That's a good point, I'll look at what is involved in doing that

slang25:
I looked at this a while ago and from memory @martincostello is right, there are a few things that should be disposable. If people are starting and stopping with JustSaying, my bet is that stuff will leak at the moment.

adammorr:
Tests will fail but some feedback on the approach would be great!

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -86,39 +85,30 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
...
 
-            _cts = new CancellationTokenSource();
-            Task.Factory.StartNew(async () =>
-                {
-                    while (!_cts.IsCancellationRequested)
-                    {
-                        await ListenLoop(_cts.Token).ConfigureAwait(false);
-                    }
-                })
+            Task.Factory.StartNew(async () => { await ListenLoop(cancellationToken).ConfigureAwait(false); },
```
martincostello: This block looks like it might need some TLC with regards to the recommendations in the doc David Fowler shared as well, due to the `ContinueWith()` usage. We could do that separately though.

JustSaying/JustSayingBus.cs
```diff
@@ -122,11 +124,13 @@ public void Start()
                     return;
                 }
 
+                _subscriberCancellationTokenSource = new CancellationTokenSource();
```
martincostello: Yeah, seems like we'd want to make `IDisposable` available in this class also, otherwise we're just moving the problem around.

JustSaying/Messaging/INotificationSubscriber.cs
```diff
@@ -8,8 +9,7 @@ namespace JustSaying.Messaging
     public interface INotificationSubscriber : INotificationSubscriberInterrogation
     {
         void AddMessageHandler<T>(Func<IHandlerAsync<T>> handler) where T : Message;
-        void Listen();
-        void StopListening();
+        void Listen(CancellationToken cancellationToken);
```
martincostello: We'd need to document this and make it part of `7.0.0` as it's a breaking interface change.

JustSaying/JustSayingFluently.cs
```diff
@@ -406,5 +406,11 @@ public IMayWantOptionalSettings WithAwsClientFactory(Func<IAwsClientFactory> aws
             _awsClientFactoryProxy.SetAwsClientFactory(awsClientFactory);
             return this;
         }
+
+        public void Dispose()
```
martincostello: If the class is public and not sealed, it should implement with the `protected void Dispose(bool disposing)` pattern with a finalizer so that derived types can participate in disposal too.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
adammorr: Issue https://github.com/justeat/JustSaying/issues/404 created for this...hopefully you can find it :trollface:

JustSaying/JustSayingFluently.cs
slang25: The dispose pattern has been an anti-pattern for a number of years now, so I'd suggest leaving it. I'll sends some links tomorrow to clarify.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -86,22 +85,16 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
...
```
slang25: `ListenLoop` very misleadingly is actually just one batch receive, the `while` you've removed was the loop.

JustSaying/Messaging/INotificationSubscriber.cs
slang25: This could be an overload and the original passes `CancellationToken.None`, but I feel like it should probably go out with a major release anyway, if the fluent api gets updating for example.

JustSaying/JustSayingFluently.cs
martincostello: What I was trying to say (badly) was "seal this class".

martincostello: When inheritance is expected and you don't use the pattern, integrators have a bad time when they want to dispose things they add in a derived types, leading to me having to raise PRs like this as otherwise you end up with _horrible_ code using `new` to replace implementations that cause whole heaps of sadness: aspnet/Mvc#7637

slang25: Agreed, in the example you sent that could have been avoided if they had made the dispose method virtual. To be fair, the dispose pattern makes sense for Microsoft as they are building the building blocks for all types of crazy that people might do, but I really think they should change the guidance for everyone else.

Agreed with "seal this class" 👍

AnthonySteele:
The `SqsNotificationListener` owns a `CancellationTokenSource` so that when `StopListening` is called  then it will trigger cancellation and fall out of the loop and end. This is typically done  by some other thread, while it is listening, because the app is shutting down. I.e. all Listeners are shut down at the same time by the runtime.

I had another design in mind, where the `SqsNotificationListener` is given a `CancellationTokenSource` which the caller owns and decides when to trigger.  there is no need for `StopListening` - the caller performs shutdown instead by triggering the cts, and the same cts can be given to all Listeners. There is also no need for the `SqsNotificationListener` to dispose of the cts, as it does not own it.

adammorr:
@AnthonySteele is what I've put together here what you were getting at e.g. `StopListening()` has been removed from `INotificationSubscriber` and the `CancellationToken` from `JustSayingBus` is passed through when `Start()` is called and then cancelled when `Stop()` is

slang25:
I think we should adopt @AnthonySteele recommendation, because:
- We'd have 2 ways to cancel/stop, not problematic, but not a simple design
- We sometimes have ownership of disposable resources
- The more general use case for JustSaying is to start listening, stopping doesn't warrant its own api, a cancellation token is adequate.

JustSaying/IAmJustSaying.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace JustSaying
 {
-    public interface IAmJustSaying : IMessagePublisher
+    public interface IAmJustSaying : IMessagePublisher, IDisposable
```
slang25: See comment, I think we shouldn't own a `CancellationTokenSource`, so we wouldn't need to implement `IDisposable`

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -86,22 +85,16 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
             _handlerMap.Add(typeof(T), handlerFunc);
         }
 
-        public void Listen()
+        public void Listen(CancellationToken cancellationToken)
```
slang25:
```suggestion
        public void Listen(CancellationToken cancellationToken = default)
```

slang25: Does this work? I've not tried it before

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -143,54 +135,52 @@ private static string AggregateExceptionDetails(AggregateException ex)
             return innerExDetails.ToString();
         }
 
-        public void StopListening()
-        {
-            _cts.Cancel();
-            _log.LogInformation($"Stopping Listening - Queue: {_queue.QueueName}, Region: {_queue.Region.SystemName}");
-        }
-
```
slang25: 👍

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -143,54 +135,52 @@ private static string AggregateExceptionDetails(AggregateException ex)
...
+                try
                 {
-                    foreach (var message in sqsMessageResponse.Messages)
+                    if (sqsMessageResponse != null)
                     {
-                        HandleMessage(message);
+                        foreach (var message in sqsMessageResponse.Messages)
+                        {
+                            HandleMessage(message);
+                        }
```
slang25: 👍

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
slang25: I mean the GitHub suggestions thing, you do three backticks and "suggestion" followed by the new code

adammorr:
I'll resolve the conflicts when we get to a consensus on how this change is implemented

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -141,7 +141,7 @@ internal async Task ListenLoop(CancellationToken ct)
             var region = _queue.Region.SystemName;
             ReceiveMessageResponse sqsMessageResponse = null;
 
-            do
+            while(!ct.IsCancellationRequested)
```
martincostello: Nit: space between the `while` and `(`.

JustSaying/JustSayingBus.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace JustSaying
 {
-    public class JustSayingBus : IAmJustSaying, IAmJustInterrogating
+    public sealed class JustSayingBus : IAmJustSaying, IAmJustInterrogating
     {
         public bool Listening { get; private set; }
```
slang25: Do we still need this property (I think it's edge case, but yet). If so, we need to set it to false when cancelling.

adammorr: It would be easier to delete it and have whoever consumes this keep track of the listen status as they will own the cancellation token

slang25: Posible solution:
```csharp
public void Start(CancellationToken cancellationToken = default)
{
    cancellationToken.Register(() => Listening = false);
    ...
```

It might be better to handle it in a more coordinated way with the task that launches the listen loop though.

slang25: I've just gone through some scenarios in my head, and all of them convince me that we should get rid of it as you suggest 👍

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
AnthonySteele: Is the `= default` needed? It is going to be quite unusual to use this class directly, so we don't have to concentrate on API ergonomics.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -86,22 +85,16 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
...
 
-            _cts = new CancellationTokenSource();
-            Task.Factory.StartNew(async () =>
-                {
-                    while (!_cts.IsCancellationRequested)
-                    {
-                        await ListenLoop(_cts.Token).ConfigureAwait(false);
-                    }
-                })
+            Task.Factory.StartNew(async () => { await ListenLoop(cancellationToken).ConfigureAwait(false); },
```
shaynevanasperen: I think we should be passing the `TaskCreationOptions.LongRunning` enum here.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -86,22 +85,16 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
...
+                    cancellationToken)
                 .Unwrap()
-                .ContinueWith(t => LogTaskEndState(t, queueInfo, _log));
+                .ContinueWith(t => LogTaskEndState(t, queueInfo, _log), cancellationToken);
```
shaynevanasperen: Will this even fire if the cancellation was requested? Perhaps it should happen inside `ListenLoop` instead, just after checking that cancellation was requested?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -143,53 +135,51 @@ private static string AggregateExceptionDetails(AggregateException ex)
...
-                var msg = $"Issue receiving messages for queue {queueName}, region {region}";
-                _log.LogError(0, ex, msg);
-            }
+                    _log.LogTrace(
+                        $"Polled for messages - Queue: {queueName}, Region: {region}, MessageCount: {messageCount}");
+                }
+                catch (InvalidOperationException ex)
+                {
+                    _log.LogTrace(
+                        $"Could not determine number of messages to read from [{queueName}], region: [{region}]. Ex: {ex}");
```
shaynevanasperen: This can only assume the reason for `InvalidOperationException` because it knows what happens inside `GetNumberOfMessagesToReadFromSqs()`. It shouldn't be assuming what happens inside other methods. Better to just let this get caught by the catch-all block.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -143,53 +135,51 @@ private static string AggregateExceptionDetails(AggregateException ex)
...
+                }
+                catch (OperationCanceledException ex)
+                {
+                    _log.LogTrace($"Suspected no message in queue [{queueName}], region: [{region}]. Ex: {ex}");
```
shaynevanasperen: This could occur due to a timeout or because cancellation was requested by the calling code. Therefore this error message is misleading.

shaynevanasperen: There's a new `CancellationTokenSource` being created inside `GetMessagesFromSqsQueue` below, which is not being disposed.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
AnthonySteele: We should yes, there is an analyser warning that says the same. But I think it's separate to this PR about `CancellationToken` and `IDisposable`.

slang25: I disagree on this one, it's an async loop with very short lived thread-bound work, perfect for the thread-pool.

shaynevanasperen: @slang25 Huh? This task runs forever (or until cancellation is requested). It's the tasks inside it for each message handler that are short-lived. Or have I got this wrong?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
shaynevanasperen: See [line 249](https://github.com/justeat/JustSaying/pull/403/files#diff-9005575960680ff0f5e43cfda505b5a2R249)

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
shaynevanasperen: See [line 249](https://github.com/justeat/JustSaying/pull/403/files#diff-9005575960680ff0f5e43cfda505b5a2R249)

AnthonySteele: The loop should be refactored any to remove the `ContinueWith` - I think that the threading should be reviewed in a separate PR.

slang25: The `Task` never completes, but it's not a thread bound thing that runs forever. Its a leaky abstraction I guess. See this: https://twitter.com/davidfowl/status/1049198826668486657?s=19

slang25: I don't think newing up a thread solves anything. What would you do? Have a synchronous loop that blocks each Listen iteration? It's not solving any problems.

If there was part of the loop that was blocking for a non-ideal period of time, then we'd need to think about how to solve it. But as it is, this is a perfect use case for the thread pool.

Think of the loop as a dashed line, where the solid part is the thread bound part. This is 1 line, with massive gaps in (not too important), but the solid parts are really short (very important).

slang25: I've just concluded that "TaskCreationOptions.LongRunning is useless for Task.Run with async/await", and then found this blog post: 
http://blog.i3arnon.com/2015/07/02/task-run-long-running/

shaynevanasperen: @slang25 Thanks. That's cleared it right up for me!

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
slang25: :+1:, we should address that in separate PR IMO. It actually causes no issues, there was a leak with the linked token source leaking memory which was addressed. CTS is one of those things where most times Dispose is a no-op (see here for the full story: https://stackoverflow.com/a/30855490/2225808), but we shouldn't rely on this knowledge, we should do as the analyzer says here IMO.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
shaynevanasperen: There should be some tests proving that this code does what we expect it to do.

AnthonySteele:
Can we agree that this PR is for `CancellationTokenSource` \ `CancellationToken` \ `IDisposable` issues, and Task Creation issues are rout of scope, and should be addressed in a sperate PR?

slang25:
Agreed, we will have to talk about it a little bit cause task creation effects our ability to cancel in this case.

adammorr:
With that being said, are there any other points we want to address or I've missed in this PR and if not, are you all happy for me to proceed with cleaning it up and fixing the tests?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
shaynevanasperen: Cancelling should work in our case though, because the cancellation token is passed into the `ListenLoop` method. It's just the `LogTaskEndState` method which will not work properly.

shaynevanasperen:
I agree with @AnthonySteele about keeping this PR to just `CancellationTokenSource` \ `CancellationToken` \ `IDisposable` issues, and Task Creation issues are out of scope, and should be addressed in a separate PR.

However, I'm always weary that when this gets done (as it should, to keep changes focused on a single thing), the valuable comments from a PR are not always taken into account when doing the subsequent work. As long as whoever does this, remembers to come back and pick up all theses comments for the subsequent work, that's fine.

shaynevanasperen:
@slang25 We can still cancel the way we are doing it now (as the cancellation token is passed into the loop method). It's just the logging of task end state which won't run properly when we cancel.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -86,22 +85,16 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
             _handlerMap.Add(typeof(T), handlerFunc);
         }
 
-        public void Listen()
+        public void Listen(CancellationToken cancellationToken = default)
```
slang25: I agree with @AnthonySteele we don't need an optional parameter for the Listen method, because we have handled it in the Start method. I think I suggested it earlier not realising we we're in the "inner call"

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
slang25: If can remove cancellation token here and we'll fix this bit of code with a separate PR & discussion.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -86,22 +85,16 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
...
-            _cts = new CancellationTokenSource();
-            Task.Factory.StartNew(async () =>
-                {
-                    while (!_cts.IsCancellationRequested)
-                    {
-                        await ListenLoop(_cts.Token).ConfigureAwait(false);
-                    }
-                })
+            Task.Factory.StartNew(async () => { await ListenLoop(cancellationToken).ConfigureAwait(false); },
+                    cancellationToken)
```
slang25: Could you remove this cancellation token, again, this bit of code can be addressed in a separate PR

adammorr:
@shaynevanasperen I'd suggest raising issues for anything that was discussed in this PR which is outside the scope of the changes at this stage so we don't lose them, like you said 👍

slang25:
@adammorr I'm happy with what's being suggested, I have 3 minor outstanding comments

adammorr:
It's never good enough for you is it @stuart-lang

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
adammorr: Done

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
adammorr: Done

adammorr:
@slang25 I've addressed two comments so have I missed one? 🙉

adammorr:
@martincostello @AnthonySteele @shaynevanasperen are there any other changes we want as part of this PR or shall I go ahead and clean things up?

JustSaying/Messaging/INotificationSubscriber.cs
AnthonySteele: While you can use `CancellationToken.None` here, I would prefer that users are steered by the docs down the path of "graceful shutdown when the app exits" which means that the host app should use a `CancellationTokenSource`.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
AnthonySteele: Is it it possible to keep the ` _log.LogInformation($"Stopping Listening...` message in some form? 
It can be useful in determining if a worker shut down gracefully or not.
Maybe just put it after the end of the `while (!ct.IsCancellationRequested) {  }` loop?

adammorr: `LogTaskEndState` will log a very similar message which includes queue name and region

JustSaying/JustSayingBus.cs
```diff
@@ -126,31 +125,12 @@ public void Start()
...
-
-                foreach (var regionSubscriber in _subscribersByRegionAndQueue)
-                {
-                    foreach (var queueSubscriber in regionSubscriber.Value)
-                    {
-                        queueSubscriber.Value.StopListening();
-                    }
-                }
-                Listening = false;
+                cancellationToken.Register(() => _subscribersAreListening = false);
```
slang25: @adammorr Just realised we should probably not do this, as this creates a disposable subscription and can leak (I think). Might need this to set `_subscribersAreListening` with a try-finally inside the listen loop.

AnthonySteele:
I'm happy with it now.

martincostello:
Has conflicts that need resolving, btw.

adammorr:
@slang25 thanks for fixing the build...I forgive you for all your prior flakiness!

martincostello:
I'll let everyone else give the final ✔️ on this as they've chipped in much more feedback.

slang25:
My only outstanding concern is here:
https://github.com/justeat/JustSaying/pull/403/files#diff-033da0d02936ea6781ce382cc01f72f7R132
We are not disposing that registration. I think we should keep track of `_subscribersAreListening` when the list loop is entered and exited.

adammorr:
@slang25 addressed your comment about `_areSubscribersListening`

adammorr:
Whoops, I’ll fix the tests

adammorr:
Hmmpf need to try something else...currently the test that is failing will pass when run in isolation. @martincostello I tried your suggestion which worked once for d7711bb but not since 😞 any other thoughts off the top of your head?

martincostello:
@adammorr Bah - it must not have _really_ worked then and the timing gods just happened to be happy.

adammorr:
I broke it more 😞

adammorr:
@slang25 @martincostello @shaynevanasperen @AnthonySteele could I have another review of this please?

JustSaying/Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
```diff
@@ -21,7 +22,8 @@ public interface IMessageProcessingStrategy
         /// Launch a worker to start processing a message.
         /// </summary>
         /// <param name="action"></param>
-        void StartWorker(Func<Task> action);
+        /// <param name="ct">The cancellation token</param>
+        void StartWorker(Func<Task> action, CancellationToken ct);
```
slang25: We must remember to document this in a change log or release note

adammorr: Yeah, I’ll make a note of all the breaking changes in this PR once it’s merged

martincostello: This should have the proper name of `cancellationToken` if it’s public, rather than just `ct`.

JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -18,10 +18,16 @@ public Throttled(int maxWorkers, IMessageMonitor messageMonitor)
...
-        public void StartWorker(Func<Task> action)
+        public void StartWorker(Func<Task> action, CancellationToken ct)
         {
             var messageProcessingTask = new Task<Task>(() => ReleaseOnCompleted(action));
-            _semaphore.Wait();
+            _semaphore.Wait(ct);
+
+            if (ct.IsCancellationRequested)
+            {
+                return;
```
slang25: Before returning, be sure to call `_semaphore.Release()` to keep the count correct

JustSaying/Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs
adammorr: Renamed @martincostello

JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs
adammorr: I've changed this to catch `OperationCanceledException` instead

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
slang25: Update on this; this has now been clarified in the async guidance (see the 3rd lightbulb):
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/7e773b7ac2e266c1368113c63db5a8a3b2f5334d/AsyncGuidance.md#avoid-using-taskrun-for-long-running-work-that-blocks-the-thread

slang25:
Out of interest @adammorr, is this what fixed the tests? Was it that workers were being queued, but the tests expect that once the bus was stopped no more handler should get invoked?

adammorr:
@slang25 yea, that looked to be the case as the cancellation token was only checked once the current batch of messages were handled

slang25:
Cool, so before cancellation meant "stop listening", now it means "stop listening, and ignore any internally queued messages", which seems appropriate 👍

JustSaying.IntegrationTests/JustSayingFluently/MultiRegion/WithSqsPointToPointSubscriber/WhenAFailoverRegionIsSetup.cs
```diff
@@ -12,6 +12,7 @@
 
 namespace JustSaying.IntegrationTests.JustSayingFluently.MultiRegion.WithSqsPointToPointSubscriber
 {
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable", Justification = "Acceptable in tests")]
```
martincostello: If the blanket statement in the tests is to just turn this off, it would be less effort to just add a `<NoWarn>` for `CA1001` to the `csproj` file.

adammorr: Good point, will do

adammorr: It is done

adammorr:
Is there anything else that is required on this otherwise, can we start approving it 😄

slang25:
The implementation looks good. The only thing I'm still thinking about is the implications of not having Stop/Dispose. It could still be that `SqsNotificationListener` should implement `IDisposable`, in which case how would the `Dispose` method and the cancellation token play with each other, given we wouldn't own the cancellation token.

Issue #402, opened 2018-10-17
Review compared to "Open-source library guidance"


martincostello:
Before we release version `7.0.0`, we should review the codebase with respect to the new [_"Open-source library guidance"_](https://docs.microsoft.com/en-gb/dotnet/standard/library-guidance/) documentation from Microsoft and determine what, if anything, we _should_ do for JustSaying while we have the opportunity to make breaking changes.

The only obvious thing to consider that we don't currently do is [strong-naming](https://docs.microsoft.com/en-gb/dotnet/standard/library-guidance/strong-naming), which would allow the library to be consumed by applications/libraries that are/need strong-naming themselves.

Similar changes elsewhere on the same theme include:
  * justeat/JustEat.StatsD#147
  * justeat/httpclient-interception#56

AnthonySteele:
I'm sure that we will _debate_ strong naming.

AnthonySteele:
We have agreed to allow strong naming.

martincostello:
I'll look at doing a PR like the other two, unless someone else has a burning desire to do it.

slang25:
On strong-naming, JustBehave is now signed, but structuremap isn't and it looks like there is no intention from them to do it, not sure the best way forwards with this 😕

martincostello:
This _might_ fit in with what I'm doing in #431, where I'm trying to make the DI more pluggable as a whole.

If that all works, we could replace StructureMap with something else that is, like the Microsoft one?

martincostello:
Re-reviewing this, it looks like the path forward is to move the tests for StructureMap into their own test assembly that doesn't need `[assembly: InternalsVisibleTo("...")]` and by necessity not strong-name `JustSaying.Extensions.DependencyInjection.StructureMap`. Then everything else should be sign-able and we can turn it on for the rest of the assemblies.

PR #405, opened 2018-10-26
Best practices as-per "Open-source library guidance"

martincostello:
Relates to #402.

Update some of the build properties as per [_"Open-source library guidance"_](https://docs.microsoft.com/en-gb/dotnet/standard/library-guidance/).

## Changes
  1. Use NuGet profile name for the author.
  1. Use 64x64 logo image.
  1. Use commas to separate tags, not semi-colon.
  1. _Prepare_ to strong-name the assembly (blocked by #380).
  1. Fix the _assembly_ version to the major version number with all other components as zero, not the _package_ version number.


version.props
```diff
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <AssemblyVersion>7.0.0.0</AssemblyVersion>
```
slang25: It's already got an `AssemblyVersion` of 7.0.0.0, is this so we can manage this separately?

martincostello: This is to pin it to `7.0.0.0` for when `7.0.1` happens, which is why I've left it separate, even though it looks duplicated at the moment.

slang25: Cool, we did use to do this:
https://github.com/justeat/JustSaying/commit/0257f8a68dcc4951fa42708f8aa22a8e3253556c#diff-4b2f402501d23abbede6eac0f47783e4L35

I reverted it cause assembly binding redirects are less of a problem these days, but it likely makes it easier for some consumers still, so worth the effort.

slang25: I think this is another area that needs guidance on, cause Microsoft have stopped doing this for all of their packages now, even ones that target full framework.

martincostello: It is called out in the guidance on strong-naming:

image

slang25: Oh nice, I hadn't seen that

Issue #401, opened 2018-10-16
Use code analysers


AnthonySteele:
As an important library, JustSaying should have high code standards.

we can use [code analysers like is done here](https://github.com/justeat/JustEat.StatsD/pull/126) to enforce standards.

My experience with adding those analysers is that it's intensive work for a day or so while all the issues are worked through, and then is almost invisible. I have confirmed that there are > 100 such analyser warnings in JustSaying. This is not an unusual number.

Most common issue are in my experience around:

* `await`  without `ConfigureAwait(false)`
* string operations without a specified culture
* methods that can be static
* public fields that should not be mutable
* correct `Exception` declarations
* correct implementation of `IDisposable`, and flagging classes that own disposable resources but are not `IDisposable` in turn.

AnthonySteele:
A goof way to do it is [like this, by linking in a ruleset](https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper/NuKeeper.csproj#L10). The repo can contain multiple rulesets, for different goals. e.g. [This ruleset for tests allows underscores in method names](https://github.com/NuKeeperDotNet/NuKeeper/blob/master/CodeAnalysisRulesForTests.ruleset), whereas the shipping code does not.

martincostello:
You can also use the `NoWarn` property like this: 

https://github.com/justeat/httpclient-interception/blob/fa872e0b7d36d3ec7e53c5ab49e7fa26d90edd4b/tests/HttpClientInterception.Tests/JustEat.HttpClientInterception.Tests.csproj#L4

I prefer that approach personally, as you can have a default `ruleset` file in `Directory.Build.props` and just ignore things that you don't want in specific projects.

PR #409, opened 2018-10-30
ToLowerInvariant

AnthonySteele:

use `ToLowerInvariant()` not `ToLower()` throughout.
This should be a compatible change in almost all cases, whereas using `ToUpperInvariant()` as some guidance suggests, could introduce tricky breakages.

Work towards using code analysers


martincostello:
Relates to #401.

slang25:
I am putting together a blog post that goes into this rule, I think it's BS. There are some characters that don't round trip when you go from source->lower->upper, but also some that don't round trip when you do source->upper->lower. The perf difference is not worth talking about either, so this is good 👍

slang25:
Related, there are places we should use `CultureInfo.InvariantCulture` and `StringComparison.Ordinal` in a few places

martincostello:
@slang25 I believe @AnthonySteele is doing PRs to fix one rule at a time.

PR #411, opened 2018-11-01
Naming of constants

AnthonySteele:

Naming in `JustSayingConstants` as per code analysers rule [ca1707](https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca1707-identifiers-should-not-contain-underscores?view=vs-2017)
Change `SCREAMING_SNAKE_CASE` to `CamelCase`.

This is a breaking change.
I am doing this on it's own as it touches a few files.
NB: some of these constants can vary. What to do with them is a separate issue.

#401

### DIFF
--- a/JustSaying/AwsTools/JustSayingConstants.cs
+++ b/JustSaying/AwsTools/JustSayingConstants.cs
@@ -2,70 +2,70 @@ namespace JustSaying.AwsTools
 {
     public static class JustSayingConstants
     {
-        public const string ATTRIBUTE_REDRIVE_POLICY = "RedrivePolicy";
-        public const string ATTRIBUTE_ARN = "QueueArn";
-        public const string ATTRIBUTE_RETENTION_PERIOD = "MessageRetentionPeriod";
-        public const string ATTRIBUTE_VISIBILITY_TIMEOUT = "VisibilityTimeout";
-        public const string ATTRIBUTE_DELIVERY_DELAY = "DelaySeconds";
-        public const string ATTRIBUTE_POLICY = "Policy";
-        public const string ATTRIBUTE_ENCRYPTION_KEY_ID = "KmsMasterKeyId";
-        public const string ATTRIBUTE_ENCRYPTION_KEY_REUSE_PERIOD_SECOND_ID = "KmsDataKeyReusePeriodSeconds";
+        public const string AttributeRedrivePolicy = "RedrivePolicy";
+        public const string AttributeArn = "QueueArn";
+        public const string AttributeRetentionPeriod = "MessageRetentionPeriod";
+        public const string AttributeVisibilityTimeout = "VisibilityTimeout";
+        public const string AttributeDeliveryDelay = "DelaySeconds";
+        public const string AttributePolicy = "Policy";
+        public const string AttributeEncryptionKeyId = "KmsMasterKeyId";
+        public const string AttributeEncryptionKeyReusePeriodSecondId = "KmsDataKeyReusePeriodSeconds";
 
         /// <summary>
         /// Default visibility timeout for message in seconds
         /// </summary>
-        public static int DEFAULT_VISIBILITY_TIMEOUT = 30;
+        public static int DefaultVisibilityTimeout = 30;
         
         /// <summary>
         /// Number of times a handler will retry a message until a message 
         /// is sent to error queue
         /// </summary>
-        public static int DEFAULT_HANDLER_RETRY_COUNT = 5;
+        public static int DefaultHandlerRetryCount = 5;
 
         /// <summary>
         /// Number of times publisher will retry to publish a message if destination is down.
         /// </summary>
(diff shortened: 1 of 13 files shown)


JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -2,70 +2,70 @@ namespace JustSaying.AwsTools
...
     {
-        public const string ATTRIBUTE_REDRIVE_POLICY = "RedrivePolicy";
-        public const string ATTRIBUTE_ARN = "QueueArn";
-        public const string ATTRIBUTE_RETENTION_PERIOD = "MessageRetentionPeriod";
-        public const string ATTRIBUTE_VISIBILITY_TIMEOUT = "VisibilityTimeout";
-        public const string ATTRIBUTE_DELIVERY_DELAY = "DelaySeconds";
-        public const string ATTRIBUTE_POLICY = "Policy";
-        public const string ATTRIBUTE_ENCRYPTION_KEY_ID = "KmsMasterKeyId";
-        public const string ATTRIBUTE_ENCRYPTION_KEY_REUSE_PERIOD_SECOND_ID = "KmsDataKeyReusePeriodSeconds";
+        public const string AttributeRedrivePolicy = "RedrivePolicy";
```
martincostello: We can look at this separately, but some of these (where they're things we've defined rather than AWS) might be candidates for properties rather than constants, so that there's scope for drop-in upgrades rather than requiring recompilation, particularly where an application might have dependencies that have dependencies on JustSaying, meaning that an change in a newer version applied to a top-level app wouldn't take affect in the dependency without it also being recompiled.

JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -2,70 +2,70 @@ namespace JustSaying.AwsTools
...
+        public const string AttributeDeliveryDelay = "DelaySeconds";
+        public const string AttributePolicy = "Policy";
+        public const string AttributeEncryptionKeyId = "KmsMasterKeyId";
+        public const string AttributeEncryptionKeyReusePeriodSecondId = "KmsDataKeyReusePeriodSeconds";
 
         /// <summary>
         /// Default visibility timeout for message in seconds
         /// </summary>
-        public static int DEFAULT_VISIBILITY_TIMEOUT = 30;
+        public static int DefaultVisibilityTimeout = 30;
```
martincostello: Again, separate to renaming, but I feel these should be properties, not fields, for the same reasons as above.

JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -2,70 +2,70 @@ namespace JustSaying.AwsTools
...
         /// </summary>
-        public static int DEFAULT_PUBLISHER_RETRY_COUNT = 3;
+        public static int DefaultPublisherRetryCount = 3;
 
         /// <summary>
         /// Every time a publisher is not able to deliver a message, it will 
         /// wait {interval}*{attemptCount} miliseconds before retrying,
         /// </summary>
-        public static int DEFAULT_PUBLISHER_RETRY_INTERVAL = 100;//100 milliseconds
+        public static int DefaultPublisherRetryInterval = 100;//100 milliseconds
```
martincostello: Just an observation, it's worrying that in the same constants we use different units of time for constants. Some are seconds, some are milliseconds. That's almost asking for someone to make a mistake.

JustSaying/AwsTools/JustSayingConstants.cs
AnthonySteele: Changing to properties _should_ be transparent to the consuming files. They won't need altering.

JustSaying/AwsTools/JustSayingConstants.cs
AnthonySteele: Yep, the analysers complain that these are public mutable fields. It should not be at least one of those things. But in a separate PR.

JustSaying/AwsTools/JustSayingConstants.cs
AnthonySteele: Seperate from naming, but we can store them all as what? `Milliseconds`? `Timespan`?

martincostello: Personally I prefer `TimeSpan` as they can support more-or-less arbitrary precision, and are easy to unambiguously define, such as with `TimeSpan.FromMinutes(2)`.

AnthonySteele: Agreed, though if we e.g. change `int JustSayingConstants.DefaultVisibilityTimeout` to  `TimeSpan JustSayingConstants.DefaultVisibilityTimeout` then we will also want to change `SqsReadConfiguration.VisibilityTimeoutSeconds` as that's what it's a default for.

martincostello: Personally I'm not concerned _too_ much with the ripple effects if we're redesigning the API to make it "nicer" and those are the consequences.

PR #412, opened 2018-11-02
Use interface not array on MessageAttributeValue data

AnthonySteele:

In line with [CA1819: Properties should not return arrays](https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca1819-properties-should-not-return-arrays?view=vs-2017)
We use IEnumerable instead, which is both more widely compatible (e.g is implemented by lists) and less abusable as it's read-only

And convert it to array just before sending.

#401

### DIFF
--- a/JustSaying.Models/MessageAttributeValue.cs
+++ b/JustSaying.Models/MessageAttributeValue.cs
@@ -4,7 +4,6 @@
 
 
 using System.Collections.Generic;
-using System.IO;
 
 
 namespace JustSaying.Models
--- a/JustSaying.Models/MessageAttributeValue.cs
+++ b/JustSaying.Models/MessageAttributeValue.cs
@@ -3,6 +3,7 @@
  */
 
 
+using System.Collections.Generic;
 using System.IO;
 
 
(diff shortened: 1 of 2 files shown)


JustSaying.Models/MessageAttributeValue.cs
```diff
@@ -31,7 +32,7 @@ public class MessageAttributeValue
         /// data, or images.
         /// </para>
         /// </summary>
-        public byte[] BinaryValue { get; set; }
+        public IEnumerable<byte> BinaryValue { get; set; }
```
martincostello: `ICollection<T>` or `IList<T>` is much better as you can use `.Count` rather than needing to turn it into a more derived type. `IEnumerable<T>` as an output type (rather than an input) also can lead to bugs due to unmaterialized collections being passed around, causing enumeration multiple times, which is wasteful.

`IList<T>` would be the most compatible replacement as it has the larger surface area, and also retains the indexer, which `ICollection<T>` would not.

AnthonySteele: Hm, we don't actually count it, that's the only use in `SnsTopicBase`. 
I favour `IReadOnlyCollection<T>` in other cases.

martincostello: `I*Collection<T>` or `I*List<T>` is preferable, IMHO, and still makes the rule happy.

I think the fix hints for this rule are poor, as I believe the real intention of the rule is not suggest against not using concrete types (`T[]`, `List<T>`, `Dictionary<K, V>`), rather than having to use `IEnumerable<T>` _specifically_.

PR #413, opened 2018-11-02
QueueUrl as Uri not string

AnthonySteele:

`QueueUrl` should be strongly typed as `Uri` not string
In line with [CA1056: URI properties should not be strings](https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca1056-uri-properties-should-not-be-strings?view=vs-2017).

#401

### DIFF
--- a/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
+++ b/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
@@ -81,7 +81,7 @@ namespace JustSaying.IntegrationTests.AwsTools
                     entriesAdded++;
                 }
 
-                await client.SendMessageBatchAsync(queue.Url.ToString(), entries);
+                await client.SendMessageBatchAsync(queue.Uri.ToString(), entries);
             }
             while (entriesAdded < throttleMessageCount);
 
--- a/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
+++ b/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
@@ -81,7 +81,7 @@ namespace JustSaying.IntegrationTests.AwsTools
                     entriesAdded++;
                 }
 
-                await client.SendMessageBatchAsync(queue.Url, entries);
+                await client.SendMessageBatchAsync(queue.Url.ToString(), entries);
             }
             while (entriesAdded < throttleMessageCount);
 
--- a/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
+++ b/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
@@ -81,7 +81,7 @@ namespace JustSaying.IntegrationTests.AwsTools
                     entriesAdded++;
                 }
 
-                await client.SendMessageBatchAsync(queue.Uri.ToString(), entries);
+                await client.SendMessageBatchAsync(queue.Uri.AbsoluteUri, entries);
             }
             while (entriesAdded < throttleMessageCount);
 
--- a/JustSaying.Models/Message.cs
+++ b/JustSaying.Models/Message.cs
@@ -19,7 +19,7 @@ namespace JustSaying.Models
         public string Tenant { get; set; }
         public string Conversation { get; set; }
         public string ReceiptHandle { get; set; }
-        public Uri QueueUrl { get; set; }
+        public Uri QueueUri { get; set; }
         public int? DelaySeconds { get; set; }
         public IDictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
 
--- a/JustSaying.Models/Message.cs
+++ b/JustSaying.Models/Message.cs
@@ -19,7 +19,7 @@ namespace JustSaying.Models
         public string Tenant { get; set; }
         public string Conversation { get; set; }
         public string ReceiptHandle { get; set; }
-        public string QueueUrl { get; set; }
+        public Uri QueueUrl { get; set; }
         public int? DelaySeconds { get; set; }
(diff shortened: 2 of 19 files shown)


JustSaying.Models/Message.cs
```diff
@@ -1,29 +1,29 @@
...
+
+        public Guid Id { get; set; }
+        public DateTime TimeStamp { get; set; }
+        public string RaisingComponent { get; set; }
+        public string Version{ get; private set; }
+        public string SourceIp { get; private set; }
+        public string Tenant { get; set; }
+        public string Conversation { get; set; }
+        public string ReceiptHandle { get; set; }
+        public Uri QueueUrl { get; set; }
```
martincostello: If we're changing this, we might as well rename it to `QueueUri` as well.

JustSaying/AwsTools/ErrorQueue.cs
```diff
@@ -38,7 +38,7 @@ public override async Task UpdateQueueAttributeAsync(SqsBasicConfiguration queue
 
             var request = new SetQueueAttributesRequest
             {
-                QueueUrl = Url,
+                QueueUrl = Url.ToString(),
```
martincostello: You _should_ be able to use the `AbsoluteUri` property, rather than `ToString()` it, given that these should always be HTTP(S) URIs, not `file://` and the like.

JustSaying/AwsTools/MessageHandling/SqsQueueBase.cs
```diff
@@ -12,7 +13,7 @@ namespace JustSaying.AwsTools.MessageHandling
     public abstract class SqsQueueBase
     {
         public string Arn { get; protected set; }
-        public string Url { get; protected set; }
+        public Uri Url { get; protected set; }
```
martincostello: `QueueUri`? I'd suggest `Uri`, but then it's the same name as the type.

JustSaying.Models/Message.cs
AnthonySteele: Yep. let's get the build green first though.

JustSaying/AwsTools/MessageHandling/SqsQueueBase.cs
AnthonySteele: I'm in favour of calling it `Uri` since often it is used as e.g. ` _queue.Uri`. 
Second choice would be `QueueUri` to match the name in message.

JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
```diff
@@ -81,7 +81,7 @@ public async Task HandlingManyMessages(int throttleMessageCount)
                     entriesAdded++;
                 }
 
-                await client.SendMessageBatchAsync(queue.Url, entries);
+                await client.SendMessageBatchAsync(queue.Uri.ToString(), entries);
```
martincostello: Use `AbsoluteUri` rather than `ToString()`? (And all the other places in the diff).

JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/BaseQueuePollingTest.cs
```diff
@@ -71,7 +71,7 @@ protected override async Task When()
 
             SystemUnderTest.StopListening();
 
-            doneOk.ShouldBeTrue("Timout occured before done signal");
+            doneOk.ShouldBeTrue("Timeout occured before done signal");
```
martincostello: 👍

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -43,7 +43,7 @@ public async Task PublishAsync(Message message, CancellationToken cancellationTo
             catch (Exception ex)
             {
                 throw new PublishException(
-                    $"Failed to publish message to SQS. QueueUrl: {request.QueueUrl} MessageBody: {request.MessageBody}",
+                    $"Failed to publish message to SQS. QueueUri: {request.QueueUrl} MessageBody: {request.MessageBody}",
```
martincostello: Maybe we should `nameof()` the two property names here?

slang25:
Changing the `Message` base class makes me nervous for 2 reasons:
- We need to be sure that serialization -> deserialization can happen across any version of the message class without introducing exceptions or breaking functionality. If someone updates JustSaying on either side it shouldn't break others.
- There were versions of JustSaying (I want to say pre-4) that included their own copy of JustSaying.Models in the nupkg, which means that binary incompatible changes such as this one would cause runtime exceptions. I hope all of those versions have been removed and this is no longer an issue.

This change looks good however as the property is only set after deserialization. If there's no old version of JustSaying lying around then that second one isn't a problem either.

AnthonySteele:
I believe that this change would make `JustSaying.Models` the first project free of analyser warnings.

AnthonySteele:
`QueueUrl` appears, I think, in V5.0.0

slang25:
It was added in a version 4 minor release, and caused a lot of problem at the time due to the thing where JustSaying.Models.dll was shipped in 2 different packages, so depending which one was written last at compile time you could get runtime errors. Thinking about it, that can't cause any more problems, cause consumers that publish would have already hit that bug in that previous version and dealt with it.

AnthonySteele:
We expect to publish new versions of `JustSaying` and `JustSaying.Models` for v7, and upgrade advice will be to update `JustSaying` and `JustSaying.Models` together. 

Keeping them in sync should solve the possible problems with this PR that can occur on 1 machine, and since the wire format doesn't change, this PR doesn't introduce any problems across machines.

slang25:
To be clear, I don't have a point to make other than I can see that you have been diligent in changes to this class, and that is good 👍

PR #417, opened 2018-11-05
Turn public fields to properties or constants

AnthonySteele:

Turn public fields to properties or constants. Some of them were writable desipite the name "constants"
This makes then read-only. This should be fine, but is a breaking change.

Analyser rules:

[CA1501: Do not declare visible instance fields](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1051-do-not-declare-visible-instance-fields) and [CA2211: Non-constant fields should not be visible](https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca2211-non-constant-fields-should-not-be-visible?view=vs-2017)

#401

### DIFF
--- a/JustSaying/Messaging/MessageProcessingStrategies/MessageConstants.cs
+++ b/JustSaying/Messaging/MessageProcessingStrategies/MessageConstants.cs
@@ -1,8 +1,8 @@
-﻿namespace JustSaying.Messaging.MessageProcessingStrategies
+namespace JustSaying.Messaging.MessageProcessingStrategies
 {
     public static class MessageConstants
     {
         public const int MaxAmazonMessageCap = 10;
-        public static int ParallelHandlerExecutionPerCore = 8;
+        public const int ParallelHandlerExecutionPerCore = 8;
     }
 }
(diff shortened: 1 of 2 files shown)


JustSaying/Messaging/MessageProcessingStrategies/MessageConstants.cs
```diff
@@ -1,8 +1,8 @@
-﻿namespace JustSaying.Messaging.MessageProcessingStrategies
+namespace JustSaying.Messaging.MessageProcessingStrategies
 {
     public static class MessageConstants
     {
         public const int MaxAmazonMessageCap = 10;
-        public static int ParallelHandlerExecutionPerCore = 8;
+        public const int ParallelHandlerExecutionPerCore = 8;
```
martincostello: IMHO this one should be a property.

AnthonySteele: because we might change it and not rebuild the client?

martincostello: Correct, so that if in the future we find a better value for some reason, the values aren't compiled and baked into everything that ever referenced it.

AnthonySteele: The same logic probably also applies to `MaxAmazonMessageCap` .

martincostello: Indeed, public constants rarely need to be such.

Issue #418, opened 2018-11-06
More sensible default for sqs retention period 


payman81:
**Is your feature request related to a problem? Please describe.**
Current default for SQS retention is a mere 10 minutes which means with default settings you will lose messages if an outage last longer than 10 minutes.

**Describe the solution you'd like**
A more sensible default is to default to the maximum retention that SQS offers.

martincostello:
Fixed by #419.

PR #419, opened 2018-11-06
Issue 418 Increase default SQS retention to maximum of 4 days

payman81:
Link to issue #418

### DIFF
--- a/JustSaying/AwsTools/JustSayingConstants.cs
+++ b/JustSaying/AwsTools/JustSayingConstants.cs
@@ -41,7 +41,7 @@ namespace JustSaying.AwsTools
         /// <summary>
         /// Default message retention period on a queue in seconds
         /// </summary>
-        public static int DefaultRetentionPeriod => 60 * 10;    //10 minutes
+        public static int DefaultRetentionPeriod => MaximumRetentionPeriod;    //14 days
 
         /// <summary>
         /// Maximum message retention period on a queue in seconds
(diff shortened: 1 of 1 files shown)


martincostello:
Fixes #418.

slang25:
Deja vu: https://github.com/justeat/JustSaying/pull/272

slang25:
I think we should default to the SQS default which is 4 days

slang25:
Inline with version 4: https://github.com/justeat/JustSaying/blob/v4/JustSaying.AwsTools/JustSayingConstants.cs#L42

This must have got lost when the v4 commits were picked across into v5, that's a worry.

JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -41,7 +41,7 @@ public static class JustSayingConstants
         /// <summary>
         /// Default message retention period on a queue in seconds
         /// </summary>
-        public static int DefaultRetentionPeriod => 60 * 10;    //10 minutes
+        public static int DefaultRetentionPeriod => MaximumRetentionPeriod;    //14 days
```
slang25:
```suggestion
        public static int DEFAULT_RETENTION_PERIOD = 345600;    //4 days
```

payman81:
Greetings old friends 👋

martincostello:
Looks like this broke a test.

slang25:
Sorry, I made an invalid suggestion, and updated it 3 times in quick succession to correct it. It must have been accepted while it was invalid. A race condition in a review 😆

PR #420, opened 2018-11-06
Fix CA2208 in MessagingConfig

AnthonySteele:

Fix [analyser issue CA2208](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2208-instantiate-argument-exceptions-correctly) in the `MessagingConfig` class
use `InvalidOperationException` not `ArgumentNullExcpetion` because they're not arguments,  and often not null.

#401

### DIFF
--- a/JustSaying/MessagingConfig.cs
+++ b/JustSaying/MessagingConfig.cs
@@ -34,18 +34,21 @@ namespace JustSaying
                 throw new InvalidOperationException($"Config needs values for the {nameof(Regions)} property.");
             }
 
-            if (string.IsNullOrWhiteSpace(Regions.First()))
+            if (Regions.Any(string.IsNullOrWhiteSpace))
             {
                 throw new InvalidOperationException($"Config cannot have a blank entry for the {nameof(Regions)} property.");
             }
 
-            var duplicateRegion = Regions
+            var duplicateRegions = Regions
                 .GroupBy(x => x)
-                .FirstOrDefault(y => y.Count() > 1);
+                .Where(y => y.Count() > 1)
+                .Select(r => r.Key)
+                .ToList();
 
-            if (duplicateRegion != null)
+            if (duplicateRegions.Count > 0)
             {
-                throw new InvalidOperationException($"Config has a duplicate in {nameof(Regions)} for '{duplicateRegion.Key}'.");
+                var regionsText = string.Join(",", duplicateRegions);
+                throw new InvalidOperationException($"Config has duplicates in {nameof(Regions)} for '{regionsText}'.");
             }
 
             if (MessageSubjectProvider == null)
--- a/JustSaying/MessagingConfig.cs
+++ b/JustSaying/MessagingConfig.cs
@@ -29,9 +29,14 @@ namespace JustSaying
 
         public virtual void Validate()
         {
-            if (!Regions.Any() || string.IsNullOrWhiteSpace(Regions.First()))
+            if (!Regions.Any())
             {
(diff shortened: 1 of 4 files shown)


JustSaying/MessagingConfig.cs
```diff
@@ -24,25 +23,30 @@ public MessagingConfig()
...
+        public IList<string> Regions { get; }
         public Func<string> GetActiveRegion { get; set; }
         public IMessageSubjectProvider MessageSubjectProvider { get; set; }
 
         public virtual void Validate()
         {
             if (!Regions.Any() || string.IsNullOrWhiteSpace(Regions.First()))
             {
-                throw new ArgumentNullException("config.Regions", "Cannot have a blank entry for config.Regions");
+                throw new InvalidOperationException("Cannot have a blank entry for config.Regions");
```
martincostello: Maybe update the message to `$"Cannot have a blank entry for the {nameof(Regions)} property."`?

JustSaying/MessagingConfig.cs
```diff
@@ -24,25 +23,30 @@ public MessagingConfig()
...
             if (duplicateRegion != null)
             {
-                throw new ArgumentException($"Region {duplicateRegion.Key} was added multiple times");
+                throw new InvalidOperationException($"Region {duplicateRegion.Key} was added multiple times");
             }
 
             if (MessageSubjectProvider == null)
-                throw new ArgumentNullException("config.MessageSubjectProvider");
+            {
+                throw new InvalidOperationException("config.MessageSubjectProvider cannot be null");
```
martincostello: `$"{nameof(MessageSubjectProvider)} cannot be null"`

JustSaying/MessagingConfig.cs
```diff
@@ -24,25 +23,35 @@ public MessagingConfig()
...
         {
-            if (!Regions.Any() || string.IsNullOrWhiteSpace(Regions.First()))
+            if (!Regions.Any())
             {
-                throw new ArgumentNullException("config.Regions", "Cannot have a blank entry for config.Regions");
+                throw new InvalidOperationException($"Config needs values for the {nameof(Regions)} property.");
             }
 
-            var duplicateRegion = Regions.GroupBy(x => x).FirstOrDefault(y => y.Count() > 1);
+            if (string.IsNullOrWhiteSpace(Regions.First()))
```
AnthonySteele: I am tempted to replace this with `if (Regions.Any(string.IsNullOrWhiteSpace))`
What do you think?

martincostello: Looks more readable.

AnthonySteele: Ana also different: if there are 2 regions in the list (primary and secondary), it will check them both for blank values.

JustSaying/MessagingConfig.cs
```diff
@@ -24,25 +23,35 @@ public MessagingConfig()
...
+            }
+
+            var duplicateRegion = Regions
+                .GroupBy(x => x)
+                .FirstOrDefault(y => y.Count() > 1);
+
             if (duplicateRegion != null)
             {
-                throw new ArgumentException($"Region {duplicateRegion.Key} was added multiple times");
+                throw new InvalidOperationException($"Config has a duplicate in {nameof(Regions)} for '{duplicateRegion.Key}'.");
```
shaynevanasperen: This will only describe the first duplicate. We may as well describe all found duplicates.

AnthonySteele: Done

PR #423, opened 2018-11-07
Review impl of RegionResourceCache

AnthonySteele:

review `RegionResourceCache` because the analyser doesn't like it.

Inheriting from `Dictionary<K,V>` comes with baggage for [naming](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1710-identifiers-should-have-correct-suffix) and [serialisation](https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca2237-mark-iserializable-types-with-serializableattribute)

better to encapsulate the dictionary and seal the containing class.
And the interface serves no role either - there's no injection happening on it, and no reason or mechanism for  swapping in a different implementation.

#401

### DIFF
--- a/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
+++ b/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Concurrent;
 
 namespace JustSaying.AwsTools.QueueCreation
@@ -5,7 +6,7 @@ namespace JustSaying.AwsTools.QueueCreation
     public sealed class RegionResourceCache<T>
     {
         private readonly ConcurrentDictionary<string, ConcurrentDictionary<string, T>> _regionsData
-            = new ConcurrentDictionary<string, ConcurrentDictionary<string, T>>();
+            = new ConcurrentDictionary<string, ConcurrentDictionary<string, T>>(StringComparer.InvariantCultureIgnoreCase);
 
         public T TryGetFromCache(string region, string key)
         {
@@ -21,7 +22,7 @@ namespace JustSaying.AwsTools.QueueCreation
         public void AddToCache(string region, string key, T value)
         {
             var regionDict = _regionsData.GetOrAdd(region,
-                r => new ConcurrentDictionary<string, T>());
+                r => new ConcurrentDictionary<string, T>(StringComparer.InvariantCultureIgnoreCase));
 
             regionDict[key] = value;
         }
--- a/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
+++ b/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using System.Collections.Concurrent;
 
 namespace JustSaying.AwsTools.QueueCreation
--- a/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
+++ b/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
@@ -2,7 +2,7 @@ using System.Collections.Generic;
 
 namespace JustSaying.AwsTools.QueueCreation
 {
-    public sealed class RegionResourceCache<T> : IRegionResourceCache<T>
+    public sealed class RegionResourceCache<T>
     {
         private readonly Dictionary<string, Dictionary<string, T>> _regionsData
             = new Dictionary<string, Dictionary<string, T>>();
--- a/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
+++ b/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
@@ -1,5 +1,4 @@
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 
 namespace JustSaying.AwsTools.QueueCreation
 {
--- a/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
+++ b/JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
@@ -9,27 +9,27 @@ namespace JustSaying.AwsTools.QueueCreation
 
         public T TryGetFromCache(string region, string key)
         {
-            if (!_regionsData.ContainsKey(region))
+            if (! _regionsData.TryGetValue(region, out var regionDict))
             {
                 return default;
             }
 
-            var regionDict = _regionsData[region];
-            if (! regionDict.ContainsKey(key))
+            if (! regionDict.TryGetValue(key, out var value))
             {
                 return default;
             }
 
-            return regionDict[key];
+            return value;
         }
 
         public void AddToCache(string region, string key, T value)
         {
-            if (!_regionsData.ContainsKey(region))
+            if (!_regionsData.TryGetValue(region, out var regionDict))
             {
-                _regionsData[region] = new Dictionary<string, T>();
+                regionDict = new Dictionary<string, T>();
+                _regionsData[region] = regionDict;
             }
-            var regionDict = _regionsData[region];
+
             regionDict[key] = value;
         }
     }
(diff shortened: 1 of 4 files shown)


JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,36 @@
...
+{
+    public sealed class RegionResourceCache<T>
     {
+        private readonly Dictionary<string, Dictionary<string, T>> _regionsData
+            = new Dictionary<string, Dictionary<string, T>>();
+
         public T TryGetFromCache(string region, string key)
         {
-            if (! ContainsKey(region))
+            if (!_regionsData.ContainsKey(region))
```
martincostello: Use `TryGetValue()`. It's cheaper than a `Contains()` call and then a `[]` call,  and is race-condition safe.

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,36 @@
...
             {
-                return default(T);
+                return default;
             }
 
-            var regionDict = this[region];
+            var regionDict = _regionsData[region];
             if (! regionDict.ContainsKey(key))
```
martincostello: As above.

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,36 @@
...
+                return default;
             }
 
             return regionDict[key];
         }
 
         public void AddToCache(string region, string key, T value)
         {
-            if (!ContainsKey(region))
+            if (!_regionsData.ContainsKey(region))
```
martincostello: Same here.

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,36 @@
 using System.Collections.Generic;
 
 namespace JustSaying.AwsTools.QueueCreation
-{
-    public class RegionResourceCache<T> : Dictionary<string, Dictionary<string, T>>, IRegionResourceCache<T>
+{
+    public sealed class RegionResourceCache<T>
     {
+        private readonly Dictionary<string, Dictionary<string, T>> _regionsData
```
martincostello: Maybe we should use ConcurrentDictionary instead?

adammorr: I had some weird issues when trying to parallelalalalalalize subscription creation which came down to this class and specifically reading / writing to the dictionary

martincostello: image

AnthonySteele: So we currently have `Dictionary<string, Dictionary<string, T>>` 
Do we need `ConcurrentDictionary<string, Dictionary<string, T>>` or `ConcurrentDictionary<string, ConcurrentDictionary<string, T>>` ?

martincostello: Something like IDictionary<string, IDictionary<string, T>>, then the actual implementations can be concurrent dictionaries. The type of the field doesn’t have to actually mention ConcurrentDictionary and can just leverage the co/contra/whichever-variance.

martincostello: Something like IDictionary<string, IDictionary<string, T>>, then the actual implementations can be concurrent dictionaries. The type of the field doesn’t have to actually mention ConcurrentDictionary and can just leverage the co/contra/whichever-variance.

martincostello:
I was leaving a comment but it got lost (I think) as a commit was pushed while I was writing it, but you can use IDictionary<string, IDictionary<string, T>> and happen to use ConcurrentDictionary while assigning them, rather than needing to nest the concrete type itself.

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,31 @@
...
-}
\ No newline at end of file
+using System.Collections.Concurrent;
+
+namespace JustSaying.AwsTools.QueueCreation
+{
+    public sealed class RegionResourceCache<T>
+    {
+        private readonly ConcurrentDictionary<string, ConcurrentDictionary<string, T>> _regionsData
+            = new ConcurrentDictionary<string, ConcurrentDictionary<string, T>>();
```
martincostello: So `ID<string, ID<string, T>> d = new CD<string, ID<string, T>>();` (shorthand as on phone) and then just use CD again when adding items.

AnthonySteele: We are now using `GetOrAdd`, which is not present on `IDictionary`.

martincostello: That’s true, but if it was using the interfaces it could’ve just used the simpler TryGetValue() and the rest of the code wouldn’t have needed to know the detail.

Meh, don’t have to change it now, but using the interfaces would have made it less verbose but still benefited from the concurrent-ness as AFAIK the interface methods leverage the concrete methods where the concurrentness is made explicit.

AnthonySteele: I am finding lines 26-27 less verbose than the `if TryGetValue` equivalent. It's no important though, as concurrency on adding regions (not keys within regions) seems to be an edge case,

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,31 @@
...
+
+        public T TryGetFromCache(string region, string key)
+        {
+            if (_regionsData.TryGetValue(region, out var regionDict))
```
martincostello: Could do this as just this instead?

```csharp
if (_regionsData.TryGetValue(region, out var regionDict) &&
    regionDict.TryGetValue(key, out var value))
{
    return value;
}

return default;
```

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,30 @@
...
\ No newline at end of file
+using System;
+using System.Collections.Concurrent;
+
+namespace JustSaying.AwsTools.QueueCreation
+{
+    public sealed class RegionResourceCache<T>
+    {
+        private readonly ConcurrentDictionary<string, ConcurrentDictionary<string, T>> _regionsData
+            = new ConcurrentDictionary<string, ConcurrentDictionary<string, T>>(StringComparer.InvariantCultureIgnoreCase);
```
martincostello: `OrdinalIgnoreCase` should be used, I believe, if memory serves this one will raise a different analyzer warning as I'm 99% sure the equivalent `StringComparison` does if you don't use one of the two ordinal ones.

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
```diff
@@ -1,33 +1,30 @@
...
+                return value;
+            }
+
+            return default;
+        }
+
+        public void AddToCache(string region, string key, T value)
+        {
+            var regionDict = _regionsData.GetOrAdd(region,
+                r => new ConcurrentDictionary<string, T>(StringComparer.InvariantCultureIgnoreCase));
```
martincostello: As above.

JustSaying/AwsTools/QueueCreation/RegionResourceCache.cs
martincostello: Explanation: https://stackoverflow.com/a/72766/1064169

martincostello:
Very detailed explanation of why anything other than `Ordinal*` is almost never what you want in library code: https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#string-operations-that-use-the-invariant-culture

PR #424, opened 2018-11-10
Add IAM actions to README

adammorr:
Pull out the required IAM actions from release notes and add them to the README.

- https://github.com/justeat/JustSaying/releases/tag/v5.0.0
- https://github.com/justeat/JustSaying/releases/tag/v6.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,29 @@ Here's a snippet of the expected configuration:
 JustSaying supports error queues and this option is enabled by default. When a handler is unable to handle a message, JustSaying will attempt to re-deliver the message up to 5 times (Handler retry count is configurable) and if the handler is still unable to handle the message then the message will be moved to an error queue.
 You can opt out during subscription configuration.
 
+## IAM Requirements
+
+JustSaying requires the following IAM actions to run smoothly;
+
+```text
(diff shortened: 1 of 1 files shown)


martincostello:
Nice. Maybe we should include an example IAM policy document somewhere too?

adammorr:
@martincostello good point, I'll add one in

adammorr:
I've excluded `sns:ListSubscriptionsByTopic` as I don't believe this is required any longer for two reasons;

- My tests didn't fail when I removed this
- I _think_ we removed the check for an existing topic subscription if favour of subscribing a queue each time, regardless of whether it exists already

Does the 2nd point ring a bell to anyone?

adammorr:
No more amendments from me on this @martincostello, promise!

martincostello:
@adammorr Now you definitely can't 😉

slang25:
@adammorr You are right, there was a change that improved startup perf by doing what you described.
We should probably remove this line here in the default SNS policy:
https://github.com/justeat/JustSaying/blob/master/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs#L49

PR #425, opened 2018-11-10
Retention period as TimeSpan not int

AnthonySteele:

Retention period typed as as `TimeSpan` not `int`. Convert to and from seconds only when needed by AWS.

Strong typing is good. Constants are now self-documenting - e.g. `int DefaultRetentionPeriod => 345600;    //4 days` becomes `TimeSpan DefaultRetentionPeriod => TimeSpan.FromDays(4);`

I have one concern, if  `duration.TotalSeconds.ToString(CultureInfo.InvariantCulture)` is the same as before, because `TotalSeconds` is typed as `double` not `int`. A helper / extension method may be needed to handle the conversions to Seconds, to int and then to string.

There are other duration constants still typed as `int` and they are a mixture of milliseconds and seconds scales, but out of scope as small PRs work better.

Based on discussion here https://github.com/justeat/JustSaying/pull/411#discussion_r230061498

### DIFF
--- a/JustSaying/AwsTools/ErrorQueue.cs
+++ b/JustSaying/AwsTools/ErrorQueue.cs
@@ -25,7 +25,7 @@ namespace JustSaying.AwsTools
             return new Dictionary<string, string>
             {
                 { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD, queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)},
-                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT, JustSayingConstants.DefaultVisibilityTimeout.ToString(CultureInfo.InvariantCulture)},
+                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT, JustSayingConstants.DefaultVisibilityTimeout.ToString("F0", CultureInfo.InvariantCulture)},
             };
         }
 
--- a/JustSaying/AwsTools/ErrorQueue.cs
+++ b/JustSaying/AwsTools/ErrorQueue.cs
@@ -24,8 +24,8 @@ namespace JustSaying.AwsTools
         {
             return new Dictionary<string, string>
             {
-                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD , queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString(CultureInfo.InvariantCulture)},
-                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT  , JustSayingConstants.DefaultVisibilityTimeout.ToString(CultureInfo.InvariantCulture)},
+                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD, queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)},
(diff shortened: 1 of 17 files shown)


JustSaying/AwsTools/ErrorQueue.cs
```diff
@@ -24,7 +24,7 @@ public ErrorQueue(RegionEndpoint region, string sourceQueueName, IAmazonSQS clie
         {
             return new Dictionary<string, string>
             {
-                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD , queueConfig.ErrorQueueRetentionPeriodSeconds.ToString(CultureInfo.InvariantCulture)},
+                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD , queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString(CultureInfo.InvariantCulture)},
```
martincostello: These `ToString()`s could be fixed as per your concern by specifying a format string that requires no decimal places, such as `"F0"`, which appears to do rounding rather than truncation.

Using [this](https://www.microsoft.com/net/learn/dotnet/in-browser-tutorial/1) for a quick REPL, the following output comes from this code:

```csharp
using System;

namespace Formatting
{
    internal static class Program
    {
        internal static void Main()
        {
            var duration = TimeSpan.FromSeconds(1234.56);
            Console.WriteLine(duration.TotalSeconds.ToString(System.Globalization.CultureInfo.InvariantCulture));
            Console.WriteLine(duration.TotalSeconds.ToString("N0", System.Globalization.CultureInfo.InvariantCulture));
            Console.WriteLine(duration.TotalSeconds.ToString("F0", System.Globalization.CultureInfo.InvariantCulture));
        }
    }
}
```

```sh
1234.56
1,235
1235
```

JustSaying/AwsTools/ErrorQueue.cs
```diff
@@ -24,8 +24,8 @@ public ErrorQueue(RegionEndpoint region, string sourceQueueName, IAmazonSQS clie
         {
             return new Dictionary<string, string>
             {
-                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD , queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString(CultureInfo.InvariantCulture)},
-                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT  , JustSayingConstants.DefaultVisibilityTimeout.ToString(CultureInfo.InvariantCulture)},
+                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD, queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)},
+                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT, JustSayingConstants.DefaultVisibilityTimeout.ToString(CultureInfo.InvariantCulture)},
```
martincostello: Did you miss this one?

AnthonySteele:
[I found the docs here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SetQueueAttributes.html)

> Valid values: An integer representing seconds, from 60 (1 minute) to 1,209,600 (14 days). Default: 345,600 (4 days).

So it is correct to round to the nearest whole second.

JustSaying/AwsTools/ErrorQueue.cs
AnthonySteele: I did

PR #428, opened 2018-11-12
Remove ListSubscriptionsByTopic

adammorr:
Removes `ListSubscriptionsByTopic` from the default SNS policy as no longer needed


adammorr:
@slang25 as you wished 🎆

PR #427, opened 2018-11-11
DeliveryDelay as TimeSpan not int

AnthonySteele:
_Summarise the changes this Pull Request makes._

`DeliveryDelay` as `TimeSpan` not `int`.

Follow on from #425

_Please include a reference to a GitHub issue if appropriate._

PR #429, opened 2018-11-12
Constants as TimeSpan not int

AnthonySteele:

`DefaultVisibilityTimeout` and `DefaultPublisherRetryInterval` as `TimeSpan` not `int`.
These are the last 2 in `JustSayingConstants`.

Follow on from #425 and #427

### DIFF
--- a/JustSaying/AwsTools/ErrorQueue.cs
+++ b/JustSaying/AwsTools/ErrorQueue.cs
@@ -8,6 +8,7 @@ using Amazon.SQS.Model;
 using Amazon.SQS.Util;
 using JustSaying.AwsTools.MessageHandling;
 using JustSaying.AwsTools.QueueCreation;
+using JustSaying.Extensions;
 using Microsoft.Extensions.Logging;
 
 namespace JustSaying.AwsTools
@@ -24,8 +25,8 @@ namespace JustSaying.AwsTools
         {
             return new Dictionary<string, string>
             {
-                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD, queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)},
-                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT, JustSayingConstants.DefaultVisibilityTimeout.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)},
+                { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD, queueConfig.ErrorQueueRetentionPeriod.AsSecondsString() },
+                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT, JustSayingConstants.DefaultVisibilityTimeout.AsSecondsString() },
             };
         }
 
@@ -42,8 +43,7 @@ namespace JustSaying.AwsTools
                 Attributes = new Dictionary<string, string>
                 {
                     {
-                        JustSayingConstants.AttributeRetentionPeriod,
-                        queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)
+                        JustSayingConstants.AttributeRetentionPeriod, queueConfig.ErrorQueueRetentionPeriod.AsSecondsString()
                     }
                 }
             };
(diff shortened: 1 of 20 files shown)


JustSaying/AwsTools/ErrorQueue.cs
```diff
@@ -25,7 +25,7 @@ public ErrorQueue(RegionEndpoint region, string sourceQueueName, IAmazonSQS clie
             return new Dictionary<string, string>
             {
                 { SQSConstants.ATTRIBUTE_MESSAGE_RETENTION_PERIOD, queueConfig.ErrorQueueRetentionPeriod.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)},
-                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT, JustSayingConstants.DefaultVisibilityTimeout.ToString("F0", CultureInfo.InvariantCulture)},
+                { SQSConstants.ATTRIBUTE_VISIBILITY_TIMEOUT, JustSayingConstants.DefaultVisibilityTimeout.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture)},
```
martincostello: Maybe we've done this `F0` call so many times across the code base now we should add an internal helper/extension method to do it?

JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -31,9 +31,9 @@ public static class JustSayingConstants
 
         /// <summary>
         /// Every time a publisher is not able to deliver a message, it will 
-        /// wait {interval}*{attemptCount} miliseconds before retrying,
+        /// wait {interval}*{attemptCount}  before retrying,
```
martincostello: Two spaces between `}` and `before`.

JustSaying/JustSayingBus.cs
```diff
@@ -216,7 +216,8 @@ private async Task PublishAsync(IMessagePublisher publisher, Message message, in
                 }
 
                 _log.LogWarning(0, ex, $"Failed to publish message {message.GetType()}. Retrying after attempt {attemptCount} of {Config.PublishFailureReAttempts}");
-                await Task.Delay(Config.PublishFailureBackoffMilliseconds * attemptCount, cancellationToken).ConfigureAwait(false);
+                var millisDelay = (int)(Config.PublishFailureBackoff.TotalMilliseconds * attemptCount);
```
martincostello: This is just an internal delay, so I don't think we really need to worry about doing the truncation here?

For 2.0 (not sure if it works in non-.NET Standard TFMs), could just do this ([docs](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.op_multiply?view=netcore-2.1&viewFallbackFrom=netframework-4.7.2)):

```csharp
await Task.Delay(Config.PublishFailureBackoff * attemptCount, cancellationToken).ConfigureAwait(false);
```

martincostello:
@AnthonySteele Add the milestone for future PRs when you open them please.

JustSaying/JustSayingBus.cs
AnthonySteele: That doesn't work for me,  `Config.PublishFailureBackoff * attemptCount`  gives 
`Error CS0019 Operator '*' cannot be applied to operands of type 'TimeSpan' and 'int`, There's no "multiply" method found iether in this version of .NET.

It is truncated to int because `TotalMilliseconds` is typed as `double`, and the next line is using `Task.Delay(int millisecondsDelay`

martincostello: Fair enough regarding the `*` operator - it must not be available for 100% of our TFMs.

Why not use the overload that takes a `TimeSpan` then?

JustSaying/AwsTools/ErrorQueue.cs
AnthonySteele: Agreed, done

JustSaying/JustSayingBus.cs
AnthonySteele: Ah yes, that version is just as succinct, because `TimeSpan.FromMilliseconds`  _does_ accept a double value.

JustSaying/Extensions/TimespanExtensions.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using System.Globalization;
+
+namespace JustSaying.Extensions
+{
+    public static class TimespanExtensions
```
martincostello: If you add an attribute [like this](https://github.com/martincostello/sqllocaldb/blob/d734da85cd2e4355a3e662df7d4b73b9c4e5dd55/src/SqlLocalDb/ISqlLocalDbInstanceInfoExtensions.cs#L13) you can hide the class from Intellisense so people don't get suggestions to write code like `TimespanExtensions.AsSecondsString(TimeSpan.Zero)`.

PR #426, opened 2018-11-10
Make integration test fully async

slang25:
Attempt to kill almost all blocking code, this removes all the occurences in the test projects. There are only 2 places left now:
- [Fluent API](https://github.com/justeat/JustSaying/search?q=GetAwaiter+filename%3AJustSayingFluently.cs&unscoped_q=GetAwaiter+filename%3AJustSayingFluently.cs), this we are well aware of
- ~~[Thottled.cs](https://github.com/justeat/JustSaying/blob/fa06e1b5f5ca35f11d734c9877141f26c77d369c/JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs#L24), this one is tricky and could be hit under contention when at throttling limits, I have an idea on how to fix this in a non-breaking way. For another PR.~~

Something is going on with line-endings, I don't know if it's me, or if the code base needs all lines normalizing. I'll investigate.

Issue #414, opened 2018-11-02
Add XML documentation to public members


martincostello:
Public members of the assemblies we ship via NuGet should have appropriate `///` documentation so that consumer of the libraries get nice Intellisense documentation, making more "traditional" documentation of methods, properties etc. less important.

martincostello:
This is still being worked on.

PR #330, opened 2017-08-16
Explore - Rewrite Fluent API

adammorr:
@slang25 first pass of exploring/rewriting the fluent API and splitting out/changing the underlying bus implementations

```
// Configure bus (regions, dependencies etc.)
var configuredBus = new BusBuilder().UsingAws(cfg =>
{
    cfg.UseRegions("eu-west-1", "eu-central-1");
    cfg.ConfigureDependencies(dep =>
    {
        dep.MessageMonitor = new NullOpMessageMonitor();
    });
});

// Publisher
var publisher = await configuredBus.CreatePublishers()
    .AddTopicPublisher<MessageOne>(cfg =>
    {
        cfg.TopicNameOverride = "override-topic-name";
        cfg.AdditionalSubscribers = new[] {"123", "456", "789"};
    })
    .AddQueuePublisher<MessageTwo>(cfg =>
    {
        cfg.QueueNameOverride = "override-queue-name";
        cfg.DeliveryDelaySeconds = 10;
        cfg.ErrorQueueOptOut = true;
        cfg.ErrorQueueRetentionPeriodSeconds = 300;
        cfg.MessageRetentionSeconds = 30;
        cfg.RetryCountBeforeSendingToErrorQueue = 8;
        cfg.VisibilityTimeoutSeconds = 30;
    })
    .BuildAsync();

await publisher.PublishAsync(new MessageOne());

// Subscriber
var subscriber = await configuredBus.CreateSubscribers()
    .AddTopicSubscriber(cfg =>
    {
        cfg.TopicNameOverride = "override-topic-name";
    })
    .WithHandler(new MessageOneHandler())
    .AddQueueSubscriber(cfg =>
    {
        cfg.QueueNameOverride = "override-queue-name";
    })
    .WithHandler(new MessageTwoHandler())
    .BuildAsync();

subscriber.StartListening();
subscriber.StopListening();
```

PR #431, opened 2018-11-13
[WIP] BusBuilder prototype first-step

martincostello:
Add new `BusBuilder` class as a prototype of a replacement for `CreateMeABus`.


slang25:
Would the entry point here be static? Or is the idea that the builder is resolved from DI with some dependencies injected?

slang25:
I assume this would also look to solve the lack of async and deferred execution of the current builder, in which case we no longer require a logger as the first thing to be configured.

martincostello:
My idea would be that there's one _somewhere_ and you get it however you want (your own static, DI, whatever) and then chain from there.

Basically I want to get rid of all the static stuff as it's the source of 90% of the pain in the integration tests with the shared static for mocking out the factory to do things.

This is just the barest skeleton of the sort of shape things _could_ take with an arbitrary name I've picked, then if we like where it's going I can start actually trying to plumb it in and cast off the `static`s as much as possible (but after #426 gets merged).

martincostello:
Indeed, with the builder pattern we could make things very lazy/configurable over time, and then it would be up to the integrator to either use it once and use that forever, or use the builder as part of DI registration and get things as they are needed.

slang25:
Cool, I agree on all points. It'd be nasty if there were some dependencies that were baked into a static factory method for example.

I was thinking that potentially we'd just be building up a data structure with the fluent api, and the starting point would be blank, so it's kind of just a style thing whether it's static or not.

If we are going to allow cross cutting things (let's say ILogger, perfect example) to be injected via DI, then I could see it being beneficial to go down the instance method entrypoint route.

martincostello:
Any thoughts on this from anyone else (@AnthonySteele, @brainmurphy) before I start thinking about making this PR actually do something working?

martincostello:
FYI I've started working on a working prototype for this. I'm doing it as an _alternate_ implementation for now, rather than replacing anything already in the library.

martincostello:
So I've got a basic skeleton of a API surface we could look towards doing. An example of it is below, which is just a simple unit test.

The general gist is:
  1. You create a builder;
  1. You configure everything you need to on it (config, handlers, publisher, subscribers, regions etc.);
  1. You call `Build()` when you're done configuring;
  1. You get back an `IMessagingBus`;
  1. The host calls `Start(...)` on the bus.

The `MessagingBusBuilder` acts as the root, and you could pass it around to split up configuration/registration into chunks.

Then once everything is done, you just get the simple `IMessagingBus` instance which only lets you start and stop processing messages, which the host application would use at the appropriate point (e.g. the end of `void Configure(IApplicationBuilder app)` in a ASP.NET Core application).

One (of the many things) missing so far is a need to hook DI in to the builder for use during `Build()`, though this is _partially_ helped by the `Func<T>` overloads for various things on the builder.

```csharp
[Fact]
public static void Can_Create_Messaging_Bus_Fluently()
{
    // Arrange
    var builder = new MessagingBusBuilder()
        .Client()
            .WithBasicCredentials("accessKey", "secretKey")
        .And().Parent
        .Messaging()
            .WithRegions("eu-west-1", "eu-central-1")
            .And()
            .WithActiveRegion("eu-west-1")
        .And().Parent
        .WithLoggerFactory(new LoggerFactory());

    // Assert
    IMessagingBus bus = builder.Build();

    bus.Start(new CancellationToken(canceled: true));
}
```

adammorr:
This looks good and a lot nicer than the current API...adios `CreateMeABus`

One thought I had (which sort of leans towards what Stu had mentioned) was to configure the bus with common configuration and then branch off of that into the more specific publisher / subscriber logic....maybe something like this;

```csharp
var configuredBus = new MessagingBus()
    .Messaging(cfg => {
        // Regions / Serialisation / Naming conventions
    })
    .Monitoring(cfg => {
        // Message monitor / logging / ...?
    })

IMessagePublisher publisher = await configuredBus.ConfigurePublishers()
    .AddTopicPublisher<T>()
    .AddQueuePublisher<T>()
    .BuildAsync();

IMessageSubscriber subscriber = await configuredBus.ConfigureSubscribers()
    .AddTopicSubscriber<TMessage, THandler>(cfg => {})
    .AddQueueSubscriber<TMessage, THandler>(cfg => {})
    .BuildAsync();
```
Just another idea but feel free to 🔥 it!

martincostello:
I like both of those suggestions. There's already a lot of plumbing going on for this, so I don't think the possible overhead of a few extra types for the nested builders is a bad shout.

The `And().Parent` was just a first stab at trying to _"get back"_ from a more specific API configuration back to the root while preserving a single builder chain.

It probably means I can ditch the need for the base class for the builders, which seemed like a good idea at the start, but as nothing's _really_ being subclassed within a particular builder type (and I've `sealed`'d them), it's just extra fluff now. I did a similar pattern for Selenium `IWebDriver`, where there actually was a lot of sub-classing (Chrome, Firefox, IE etc.), so that pattern actually solved a lot of duplication.

I'll mull on this and look at doing some more hacking around tomorrow to work towards the suggested nested builders.

martincostello:
I've made a first pass at the suggested approach with the nesting, and also made some tweaks here and there and fixed a few bugs along the way (I'll cherry-pick those into separate PRs later today) so that the two tests both pass again (locally at least) by pointing at goaws.

I'll come back to looking at separating out subscribing and messaging later on.

The updated sample from the test is now (I've left `And()` in for now):

```csharp
[Fact]
public void Can_Create_Messaging_Bus_Fluently()
{
    // Arrange
    var services = new ServiceCollection()
        .AddLogging((p) => p.AddXUnit(OutputHelper))
        .AddJustSaying(
            (builder) =>
            {
                builder.Client(
                            (options) => options.WithBasicCredentials("accessKey", "secretKey"))
                       .Messaging(
                            (options) => options.WithRegions("eu-west-1", "eu-central-1")
                                                .And()
                                                .WithActiveRegion("eu-west-1"))
                       .Subscriptions(
                            (options) => options.WithHandler<MyMessage>());
            })
        .AddJustSayingHandler<MyMessage, MyHandler>();

    IServiceProvider serviceProvider = services.BuildServiceProvider();

    // Assert
    var bus = serviceProvider.GetRequiredService<IMessagingBus>();
    bus.Start(new CancellationToken(canceled: true));
}
```

slang25:
It's looking really nice 👍 I'll have a deeper dive later today and feedback

slang25:
Just for reference, @adammorr did a prototype a while ago of the fluent api ([see here](https://github.com/justeat/JustSaying/pull/330))

Something that I quite liked in that prototype was making it clear if you were dealing with topics or just queues in subscriptions and publishing, with methods like `AddTopicPublisher` and `AddTopicSubscriber`, vs `AddQueuePublisher` and  `AddQueueSubscriber`.

One of the things that I find slightly annoying with the way JustSaying currently works is that you might expect those 4 apis to be relatively symmetrical,  but in the common case of `AddTopicSubscriber` it does more for you, it checks if the topic exists, creates it if not, and same for the subscription to the queue. I'm not suggesting that we change that here, but again it's something to keep in mind while giving the apis an overhaul.

While you're in the guts of this, what's your thoughts?

martincostello:
Having a quick look at @adammorr's PR, it does look like there's a lot of common ideas going on between that and this.

I think the only thing I disagree with from there is the use of interfaces. I think concrete types are much better for this scenario as it makes them easy to extend with minimal fuss and/or breaking chances, rather than the explosion of interfaces that exists in [JustSayingFluentlyInterfaces.cs](https://github.com/justeat/JustSaying/blob/master/JustSaying/JustSayingFluentlyInterfaces.cs).

martincostello:
Have cherry-picked some of the changes here into #455 and #456. Will rebase this once those are merged.

martincostello:
Rebased.

shaynevanasperen:
@martincostello This is nice work so far, so please don't take this the wrong way. I very much like @adammorr's suggestion of splitting out the publish and subscribe parts into two branches.

`AddQueuePublisher` doesn't make sense because you don't publish to queues (you publish to topics). Likewise, `AddTopicSubscriber` doesn't make sense because you don't subscribe to topics (you subscribe to queues). I'd like to see something that makes it more clear what is actually happening behind the scenes:

```cs
var configuredBus = new MessagingBus()
    .Messaging(cfg => {
        // Regions / Serialisation / Naming conventions
    })
    .Monitoring(cfg => {
        // Message monitor / logging / ...?
    })

IMessagePublisher publisher = await configuredBus.ConfigurePublishers()
    .EnsureTopicExists<T1>() // for when we want to publish to a topic which will allow fan-out to many queues
    .BuildAsync();

IMessageSender sender = await configuredBus.ConfigureSenders()
    .EnsureQueueExists<T2>() // for when we want point-to-point queues with no topic
    .BuildAsync();

IMessageSubscriber subscriber = await configuredBus.ConfigureSubscribers()
    .EnsureTopicExists<T1>(topic =>
        topic.EnsureQueueExists<THandler1>(cfg => {})) // here we are adding a queue for a specific topic (we could continue to chain more EnsureQueueExists calls here to fan-out to many queues)
    .EnsureQueueExists<T2, THandler2>(cfg => {}) // here we are adding a queue without a topic
    .BuildAsync();
```
`EnsureTopicExists` and `EnsureQueueExists` make it clear that it's creating a new topic or queue (only if they don't already exist)

Usage would be:
```cs
sender.SendAsync(...)
publisher.PublishAsync(...)
subscriber.StartListeningAsync(...)
```

martincostello:
I'll look at splitting things up more into publish/subscribe/listen when more of this PoC is wired-up.

I still need to implement _any_ publishing at all, and I only got subscriptions working last night. Publishers and listeners aren't yet wired up into the DI or available directly on the builder.

Once I've got the basics of JustSaying working end-to-end with a goaws integration that passes, I can start refactoring things to fit the suggestions more.

Writing this PR is also very much a learning process for me of what JustSaying _actually does_ under the covers.

slang25:
That's a fair point @shaynevanasperen, it was done like that because JustSaying didn't have separate semantics for message sends, we should take the time here to get it right.

adammorr:
@shaynevanasperen just on your point about `AddQueuePublisher doesn't make sense because you don't publish to queues` you can publish directly to queues (point to point in current lingo)

slang25:
I was thinking about the same thing @shaynevanasperen with regards to explicitly asking for topic/queue/subscription creation, it would need to be something that wouldn't be easy to miss for existing consumers relying on that behaviour.

@adammorr You are right, I think what Shayne means is that the existing JustSaying lingo is wrong, cause you send to queues, and publish to topics. So if we introduced a `Send` method, then we'd change the fluent api to match.

martincostello:
We can definitely tidy up the lingo, but from the descriptions it sounds more like it's the actual objects that do the work that need the naming changes, rather than the config/builder API.

martincostello:
_Also_, but unrelated, there's been some discussion of having `BuildAsync()`. This in fine _in theory_ but is problematic still anyway because if you take the ASP.NET Core startup behaviour as an example, those methods aren't `async` and would be where this code would be called, so there'd still have to be a sync-wait-on-async in that case.

I'm not sure what the best path forward for that would be, other than deferring it until the first usage in a request/handler, where things _would_ be async.

shaynevanasperen:
@martincostello I think having the _sync-wait-on-async_ for ASP.NET Core startup should be fine. If I'm not mistaken, they're planning to make that async in version 3 anyway. But until then, I think it would be okay to have sync wait on async in startup, or we could defer to first use as an optional thing (maybe we should leave that as a choice for consumers).

adammorr:
@slang25 so something like `PublishAsync()` and `SendAsync()`? This would be nice and would also open up the opportunity to have types that could be published to both queues and topics rather than the black box `PublishAsync()` and trust it goes to the right place based on how it has been registered

slang25:
Oh man, that's not idea. Looks like Shayne is right and there was talk a while back of revisiting this in 3.0: https://github.com/aspnet/Hosting/issues/1088#issuecomment-311741002

This also looks like a handy solution for now: https://www.thomaslevesque.com/2018/09/25/asynchronous-initialization-in-asp-net-core-revisited/

martincostello:
Doing some sort of async deferral looks nice, we'd just have to make sure it works for ASP.NET Core and full framework scenarios so it's not tied just to how the best solution to ASP.NET Core might be.

JustSaying/IMessagingBus.cs
```diff
@@ -0,0 +1,16 @@
+using System.Threading;
+
+namespace JustSaying
+{
+    /// <summary>
+    /// Defines a messaging bus.
+    /// </summary>
+    public interface IMessagingBus
```
slang25: When you get around to publishers, I suggest renaming `IMessagingBus` to something that indicates it's subscriber specific

shaynevanasperen: Not sure what you mean by this?

slang25: This interface has a `Start` method, which means nothing if you are just publishing

shaynevanasperen: Ah, okay. Well if we do as @adammorr and I suggested above where we make the fluent API branch into publisher/subscriber config, then this `Start` method would disappear.

shaynevanasperen:
I'd also like to see an extension method where you can do something along the lines of:

```cs
.ApplyContributionsFromAssembly(Assembly assembly)
```
...so that it uses reflection to find all the `IMessageBusContributor` (or similarly-named) implementations and instantiates them (possibly passing the DI service locator) and calls methods on them to build up the configuration. This allows putting all the related startup logic for a _feature_ inside the respective _feature folder_ instead of all lumped together in one huge `Startup` class/method.

martincostello:
To be honest that sounds fairly icky to me.

There's hooks here for plugging in DI, and there's nothing that forces to to register everything in one place. You'd just pass the builder off to some other code yourself, and as long as you wait for that to run before you do your `Build()` (or whatever we end up with) calls, it'll all be registered.

We should provide hooks for plugging in your own DI container (which we have) and then the user can plug in whatever that is themselves. Trying to roll an advanced DI resolution system for users' own types sounds like something we'd not want to have to maintain.

We could maybe provide something like `IHostingStartup` where you can _register_ an interface and we'll call back to it during `Build()`, but I wouldn't want to add any code that has to actively do scanning through assemblies with reflection to create objects ourselves to _then_ call.

slang25:
It's not as bad as it sounds (might be worth @shaynevanasperen slacking across the implementation you are talking about). It's something that should be possible regardless of whether it's in the box or not.

If you moved the DI extensions into a separate package it wouldn't be out of place there, it feels like something that can be looked at after the initial work merged.

martincostello:
You just say "reflection" and them I'm thinking:

  * Would we support `internal` types?
  * Would we support nested types?
  * What if it doesn't have a simple default constructor?

And then suddenly we're maintaining custom type activation code. I'd rather we leave out any reflection, and leave all activation to the DI container (like I've done with `AddJustSayingHandler<TMessage, THandler>()`).

I'd be OK with some sort of abstraction to provide the hooks to do the callbacks into the extra bits, but I don't think we should be activating themselves other than asking the `IServiceResolver` interface to do so (if there's one registered).

shaynevanasperen:
@martincostello Good points! But the problem is that we haven't yet constructed the `IServiceProvider` by the time we want to fetch all the _registrations_ as this code will actually be running where you're still building up the `IServiceCollection`. A classic chicken-and-egg problem!

martincostello:
Myself and @shaynevanasperen have just had a quick chat and we're basically on the same page:
  1. Registering stuff is setup as "lazy", so you can register stuff to be used at `Build()` time, that you'd do in methods like `Configure(IApplication app)` in an ASP.NET Core web app.
  1. JustSaying should delegate all object activation that isn't something internal (like `DefaultAwsClientFactory`) to a DI container with a BYO model where you just register it by type with the builder.
  1. The builder will split up more before this PR is "done", but I'll get a basic publish implementation working here first.

Then we can review the end state and either make more changes to this PR or otherwise create tasks for what we want to do next once/if this gets merged into master. Then we can start splitting up the work to flesh out the implementation more before removing the "old way" of doing things.

JustSaying/Fluent/DefaultServiceResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                return new NewtonsoftSerializationFactory();
+            }
+            else if (desiredType == typeof(IMessageSerializationRegister))
+            {
+                return new MessageSerializationRegister(ResolveService<IMessageSubjectProvider>());
+            }
+            else if (desiredType == typeof(IMessageSubjectProvider))
+            {
+                return new NonGenericMessageSubjectProvider();
```
adammorr: Seeing that we are changing a whole load of things, I think we should return GenericMessageSubjectProvider by default

JustSaying/Fluent/QueuePublicationBuilder`1.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace JustSaying.Fluent
+{
+    /// <summary>
+    /// A class representing a builder for a queue publication. This class cannot be inherited.
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of the message published to the queue.
+    /// </typeparam>
+    public sealed class QueuePublicationBuilder<T> : IPublicationBuilder<T>
```
adammorr: Name of this file needs fixing

JustSaying/Fluent/SubscriptionBuilder`1.cs
```diff
@@ -0,0 +1,128 @@
...
+
+namespace JustSaying.Fluent
+{
+    /// <summary>
+    /// A class representing a builder for a subscription. This class cannot be inherited.
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of the message.
+    /// </typeparam>
+    public sealed class SubscriptionBuilder<T> : ISubscriptionBuilder<T>
```
adammorr: File name needs fixing

martincostello:
I haven't fully split everything out yet, but I've made some progress towards the suggestions from @adammorr and @shaynevanasperen:

  1. Things like `ILoggerFactory`, `IMessageLockAsync` etc. are now configured through `.Services(...)`;
  1. Added publish support through `Publications(...)` with `WithQueue<T>()` and `WithTopic<T>()`;
  1. Updated the integration test to publish and receive messages.

An example registration now looks something like this:

```csharp
builder.Client(
            (options) => options.WithBasicCredentials("accessKey", "secretKey")
                                .WithServiceUrl("http://localhost:4100"))
       .Messaging(
            (options) => options.WithRegions("eu-west-1", "eu-central-1")
                                .WithActiveRegion("eu-west-1"))
       .Publications(
            (options) => options.WithQueue<Message1>()
                                .WithTopic<Message2>())
       .Subscriptions(
            (options) => options.WithSubscription<Message1>((p) => p.IntoQueue("foo"))
                                .WithSubscription<Message2>((p) => p.IntoQueue("bar")))
       .Services(
            (options) => options.WithMessageMonitoring(() => new MyMonitor()));

```

JustSaying/Fluent/QueuePublicationBuilder`1.cs
martincostello: To what? This is intentional.

JustSaying/Fluent/SubscriptionBuilder`1.cs
martincostello: To what? This is intentional.

JustSaying/Fluent/DefaultServiceResolver.cs
martincostello: My understanding is that this is as-is because it needs to be backwards compatible for the wire format/queue naming, rather than the C# code.

JustSaying/Fluent/ISubscriptionBuilder`1.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace JustSaying.Fluent
+{
+    /// <summary>
+    /// Defines a builder for a subscription.
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of the messages to subscribe to.
+    /// </typeparam>
+    internal interface ISubscriptionBuilder<out T>
```
shaynevanasperen: Looks like this type argument is superfluous?

martincostello: It's not - if it isn't there, [this code](https://github.com/justeat/JustSaying/blob/624a11e33b462c830c18c5915fc7ca6d1393643a/JustSaying/Fluent/SubscriptionsBuilder.cs#L101) doesn't work.

JustSaying/Fluent/SubscriptionBuilder`1.cs
adammorr: Ignore me x 2 🤦‍♂️

JustSaying/Fluent/QueuePublicationBuilder`1.cs
adammorr: Ignore me 🤦‍♂️

martincostello:
I don't see much value in drip feeding this in smaller PRs into master (other than cherry-picking bug fixes, which I've already been doing), particularly before now given how in flux this PR has been as I've tried different approaches.

slang25:
Agreed with Martin on this one, this PR is actually smaller than I expected. Previous attempts were parked as they either didn't get enough interest and consensus at the time, or they didn't fix enough of the crazy. It doesn't all have to happen in one PR, but it needs to tackle enough of the issues that there is a clear way forwards because this first attempt is a big commitment.

I really like where this PR is heading 👍

martincostello:
To stop this PR snowballing indefinitely, I think the things left I need/want to do before this PR is potentially mergaeable are:

- [ ] Complete and tidy up the topic/queue config split.
- [ ] Check I haven't missed off any "big features" that are in `JustSayingFluently` in this.
- [ ] Add a handful more integration tests for using the builder to validate it works for the common cases.
- [ ] Prototype a way to register "config callbacks" on build as suggested yesterday.

Once that's done, we can do a proper review and discuss at next week's "funday Friday" once @brainmurphy is back. Then once that's complete, if this PR is a go, we can merge it to master.

After that, we can break down tasks to spread around everyone and start fully fleshing things out and ripping out `CreateMeABus` so that this becomes the new entry point into JustSaying.

martincostello:
I'll look at getting more of those tasks above done today, BTW.

JustSaying/Fluent/ISubscriptionBuilder`1.cs
shaynevanasperen: But you could remove it from that code too, not so? In fact, that looks like it won't even work properly anyway, as it ends up calling `.WithMessageHandler<T>(resolver)` where `T` is `Message` rather than the specific message type.

martincostello: No, because I need the type to register the handlers. The types are the specific type (see the test). The internals use `out T` so it doesn't matter what the concrete type is, it's just calling a method on a type to do building.

martincostello:
A good follow-up PR to this when this is merged would be to go through the integration tests and switch them to the new pattern. That'll be a good way of flushing out any things missing/not working here.

shaynevanasperen:
@martincostello My `EnsureXXXExists(...)` methods are just suggestions of better names for what you've called `WithQueue<T>`, `WithTopic<T>`, `ForQueue<T>` and `ForTopic<T>` (because part of what they do is to check for the existence of a queue or topic and then create it if it doesn't already exist). Perhaps we can think of even better names. The right word for a topic is _topic_. A _queue_ is just an endpoint for a topic (and you can define many queues for a particular topic). SNS _pushes_ messages to a queue, that when you _subscribe_ to, get _pulled_ from the queue. There is no such thing as "subscribing" to a topic. Instead, you subscribe to a queue, which may or may not be configured as one of the potentially many endpoints of a topic.

This [StackOverflow post](https://stackoverflow.com/questions/13681213/what-is-the-difference-between-amazon-sns-and-amazon-sqs) helps explain it.

And regarding you comment about being constrained by what's _already_ in JustSaying: We're going to be breaking the public interface anyway, so we may as well start fresh and build from the ground up until we support every scenario already supported by JustSaying (rather than try to make incremental backwards-compatible changes).

martincostello:
So `TopicBuilder<T>` (i.e. drop `Subscription` from the name)? You confused me because it already had the word `Topic` in it.

I disagree on exposing `Exists...(...)` in the builder - that's an implementation detail, surely? The simple intention is that someone just wants to listen to something, right?

Regarding breaking things, I'm intentionally **not** breaking things in this PR. This PR is mostly to get the surface area, not to start ripping the internals to pieces. You've already said this PR is too big, doing that would make it even bigger. Replacing the the current internals is a job for future PRs.

I can make shape and naming changes to the builder, but I'd rather if/when this is merged it at least works to some degree with the current constraints. Then we can start taking a hatchet to things later. I can put methods on that throw `NotImplementedException` for more advanced scenarios I don't get to here, but I don't want to touch the core code yet. The intention is to add something new, which we slowly transition over to and light-up as we delete old things.

martincostello:
To be honest I still don't quite get the naming nuances/issue having read the post linked to.

Messages are published to a topic or a queue, and people subscribe to messages from them, separate to whether it's a queue or fan-out/notifications, so I don't _quite_ get what the naming issue is.

martincostello:
Also the "wrong" lingo you refer to in your example is internals to map the new shape to the existing implementation, so it doesn't matter if it's wrong at this stage because that's what's already there, it's not new public surface. Otherwise that's changing the existing API, which is outside the scope of _this_ PR>

AnthonySteele:
> We ultimately want to go back to the drawing board and redesign this from the ground up

We really do need to consider the impact of breaking changes on existing consumers. Updating hundreds of codebases is hard, we need to minimise that.

shaynevanasperen:
No, because the new thing can be incrementally built up side-by-side with the existing version (or replacing and breaking the existing version in a long-lived branch), with several small PRs.

martincostello:
goaws isn't working in Docker on my JE laptop right now, but I reckon I've got a bare minimum on 3 of the 4 tasks done now. I'll add a few more tests this evening on my own laptop to verify the core scenarios appear to work. Looks like I need to fix a test as well.

Then I guess after that it's just review feedback to address and a design meeting/discussion next week.

martincostello:
We need to start somewhere, that was the whole point of this PR.

martincostello:
Right - I [think I'm done](https://github.com/justeat/JustSaying/pull/431#issuecomment-442038926).

Latest examples from the tests for the simplest cases:

## Queue

```csharp
var services = new ServiceCollection()
    .AddJustSaying(
        (builder) =>
        {
            builder.Client((options) => options.WithBasicCredentials("accessKey", "secretKey"))
                   .Messaging((options) => options.WithRegions("eu-west-1"))
                   .Publications((options) => options.WithQueue<QueueMessage>())
                   .Subscriptions((options) => options.ForQueue<QueueMessage>())
                   .Services((options) => options.WithMessageMonitoring(() => new MyMonitor()));
        })
    .AddJustSayingHandler<QueueMessage, QueueHandler>();

IServiceProvider serviceProvider = services.BuildServiceProvider();

IMessagePublisher publisher = serviceProvider.GetRequiredService<IMessagePublisher>();
IMessagingBus listener = serviceProvider.GetRequiredService<IMessagingBus>();
```

## Topic

```csharp
var services = new ServiceCollection()
    .AddJustSaying(
        (builder) =>
        {
            builder.Client((options) => options.WithBasicCredentials("accessKey", "secretKey"))
                   .Messaging((options) => options.WithRegions("eu-west-1"))
                   .Publications((options) => options.WithTopic<TopicMessage>())
                   .Subscriptions((options) => options.ForTopic<TopicMessage>());
        })
    .AddJustSayingHandler<TopicMessage, TopicHandler>();

IServiceProvider serviceProvider = services.BuildServiceProvider();

IMessagePublisher publisher = serviceProvider.GetRequiredService<IMessagePublisher>();
IMessagingBus listener = serviceProvider.GetRequiredService<IMessagingBus>();
```

## Delegating configuration

```csharp
var services = new ServiceCollection()
    .AddJustSaying()
    .AddSingleton<IMessageBusConfigurationContributor, AwsContributor>()
    .AddSingleton<IMessageBusConfigurationContributor, QueueContributor>()
    .AddSingleton<IMessageBusConfigurationContributor, RegionContributor>()
    .AddJustSayingHandler<QueueMessage, QueueHandler>()
    .AddSingleton<MyMonitor>();

IServiceProvider serviceProvider = services.BuildServiceProvider();

IMessagePublisher publisher = serviceProvider.GetRequiredService<IMessagePublisher>();
IMessagingBus listener = serviceProvider.GetRequiredService<IMessagingBus>();

public class AwsContributor : IMessageBusConfigurationContributor
{
    public void Configure(MessagingBusBuilder builder)
    {
        builder.Client(
            (options) => options.WithSessionCredentials("accessKeyId", "secretKeyId", "token"));
    }
}

public class QueueContributor : IMessageBusConfigurationContributor
{
    public void Configure(MessagingBusBuilder builder)
    {
        builder.Publications((options) => options.WithQueue<QueueMessage>())
               .Subscriptions((options) => options.ForQueue<QueueMessage>());
    }
}

public class RegionContributor : IMessageBusConfigurationContributor
{
    public void Configure(MessagingBusBuilder builder)
    {
        builder.Messaging(
            (options) => options.WithRegion("eu-west-1"));
    }
}
```

JustSaying/Microsoft/Extensions/DependencyInjection/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,200 @@
...
+using JustSaying.AwsTools;
+using JustSaying.AwsTools.QueueCreation;
+using JustSaying.Fluent;
+using JustSaying.Messaging.MessageHandling;
+using JustSaying.Messaging.MessageSerialization;
+using JustSaying.Messaging.Monitoring;
+using JustSaying.Models;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection
```
slang25: Could we add a `JustSaying.Extensions.DependencyInjection.csproj` project as part of this? (and your thoughts on naming)

JustSaying.IntegrationTests/Fluent/MessagingBusBuilderTests.cs
```diff
@@ -0,0 +1,271 @@
...
+
+                // Assert
+                while (!source.IsCancellationRequested && !TopicHandler.MessageIds.Contains(message.Id))
+                {
+                    await Task.Delay(TimeSpan.FromSeconds(1), source.Token);
+                }
+
+                TopicHandler.MessageIds.ShouldContain(message.Id);
+            }
+        }
```
slang25: This test here is what it's about 👏 The api here looks much more approachable than v6

JustSaying/CreateMeABus.cs
```diff
@@ -9,10 +9,12 @@ public static class CreateMeABus
         /// <summary>
         /// Allows to override default <see cref="IAwsClientFactory"/> globally.
         /// </summary>
+        ////[Obsolete("Use the BusBuilder class to create message buses.")]
         public static Func<IAwsClientFactory> DefaultClientFactory { get; set; }
             = () => new DefaultAwsClientFactory();
 
-        public static JustSayingFluentlyDependencies WithLogging(ILoggerFactory loggerFactory) => 
+        ////[Obsolete("Use the BusBuilder class to create message buses.")]
```
slang25: Is `////` intentional?

JustSaying/Fluent/DefaultServiceResolver.cs
slang25: Possibly one to discuss in a separate PR, but it looks like `GenericMessageSubjectProvider` would be a better default, and compatible with the default, it only extends and fixes the cases where the type name exceeds 100 chars, or the type is generic

JustSaying/Fluent/ServiceProviderResolver.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using JustSaying.Messaging.MessageHandling;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace JustSaying.Fluent
+{
+    /// <summary>
+    /// A class that implements <see cref="IServiceResolver"/> and <see cref="IHandlerResolver"/>
+    /// for <see cref="IServiceProvider"/>. This class cannot be inherited.
+    /// </summary>
```
slang25: Are these autogenerated out of interest? There's nothing in this doc comment that isn't in the type definition

JustSaying/Fluent/SubscriptionsBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            configure(builder);
+
+            Subscriptions.Add(builder);
+
+            return this;
+        }
+
+        /// <summary>
+        /// Configures a topic subscription for the default topic name.
```
slang25: I think we should explain this a bit more, given it will be the thing that most people will want to reach for.
Something that explains that you'll end up pulling from a queue with a subscription to the topic.

JustSaying/Microsoft/Extensions/DependencyInjection/IServiceCollectionExtensions.cs
martincostello: Plan was to do that separately once/if this gets merged, then JustSaying itself doesn’t get this dependency and remains “clean”.

JustSaying/CreateMeABus.cs
martincostello: Habit to keep StyleCop happy, but I should revert this change as it’s not so much obsolete as gonna be deleted later I guess?

JustSaying/Fluent/ServiceProviderResolver.cs
martincostello: No, I wrote it, but for public shipping libraries I’m a believer in adding _something_.

JustSaying/Fluent/SubscriptionsBuilder.cs
martincostello: We can beef-up the docs once we’ve committed to this new design, rather than write too much while it’s first-steps-y.

JustSaying/Fluent/ServiceProviderResolver.cs
martincostello: As-per #414. Otherwise it’ll just generate compiler warnings.

martincostello:
Reminder: ping #466 when this PR (or some future alternative) gets merged to master.

JustSaying.IntegrationTests/Fluent/MessagingBusBuilderTests.cs
```diff
@@ -0,0 +1,271 @@
...
+        public async Task Can_Create_Messaging_Bus_Fluently_For_A_Queue()
+        {
+            // Arrange
+            var services = new ServiceCollection()
+                .AddLogging((p) => p.AddXUnit(OutputHelper))
+                .AddJustSaying(
+                    (builder) =>
+                    {
+                        builder.Client((options) => options.WithBasicCredentials("accessKey", "secretKey").WithServiceUri(TestEnvironment.SimulatorUrl))
+                               .Messaging((options) => options.WithRegions("eu-west-1"))
```
AnthonySteele: Why `(options) =>` instead of `options =>` or `o =>`

JustSaying.IntegrationTests/Fluent/MessagingBusBuilderTests.cs
```diff
@@ -0,0 +1,271 @@
...
+                listener.Start(source.Token);
+
+                var message = new QueueMessage();
+
+                await publisher.PublishAsync(message, source.Token);
+
+                // Assert
+                while (!source.IsCancellationRequested && !QueueHandler.MessageIds.Contains(message.Id))
+                {
+                    await Task.Delay(TimeSpan.FromSeconds(1), source.Token);
```
AnthonySteele: I would wait in smaller increments than that, e.g. 200ms. It is probably better for test speed, and not noticeably worse for CPU.

JustSaying.IntegrationTests/Fluent/MessagingBusBuilderTests.cs
martincostello: Purely because that's how I like to write code, and it also makes the code's intention clearer at this exact point while it serves as the only documentation.

AnthonySteele: YMMV 🤷‍♂️

AnthonySteele:
This Looks Good To Me, and I am very happy with the general approach. But I am not the expert on the fluent interface (it's the part of JustSaying that I am least familiar / concerned with) so someone else will have to say if the details fit with the various consumers.

martincostello:
I'd say that if @adammorr seems OK with it, we can get this merged so we can start fleshing it out more and starting to switch over/remove the old code over the quiet period between now and January.

adammorr:
I’ll take a look when I get back from holiday tomorrow 👍

adammorr:
This looks good to me 👍 the only question I have (which may have already been considered or will come as part of fleshing this out once merged) is around `.AddJustSayingHandler<TMessage, THandler>` being outside of the configuration within `Subscriptions(...)`.

As subscriber configuration and its handler are closely related, would we want to either;
- add safeguards to notify the user if they have configured a queue but not added a handler?
- should defining the handler be part of `SubscriptionsBuilder<T>`?
- something else?

JustSaying/Fluent/SubscriptionsBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <summary>
+        /// Configures the subscriptions for the <see cref="JustSayingFluently"/>.
+        /// </summary>
+        /// <param name="bus">The <see cref="JustSayingFluently"/> to configure subscriptions for.</param>
+        /// <exception cref="InvalidOperationException">
+        /// No instance of <see cref="IHandlerResolver"/> could be resolved.
+        /// </exception>
+        internal void Configure(JustSayingFluently bus)
+        {
+            IHandlerResolver resolver = Parent.ServiceResolver.ResolveService<IHandlerResolver>();
```
adammorr: I'm being super picky but should this be `var` or should we change line 103 to declare the type instead?

martincostello:
The `AddJustSayingHandler<,>` stuff is a DI extension more than part of the builder, so it's separate. Otherwise the builder (probably) will get more complicated as it'll need more abstract-DI-awareness baked into it.

Being able to do warnings would also mean needing introspecting the DI and/or maybe prematurely activating services, so while that might be a nice-to-have, it might get complicated.

adammorr:
OK cool - it's probably not worth incurring any additional complexity for this given the existing fluent API doesn't give you any kind of warning either!

martincostello:
Regarding the above, actually, maybe we could put in some lightweight stuff into the default `IHandlerResolver` implementation at a later point in the fleshing-out phase if it seems easy to do.

martincostello:
Right I think we're at the point now this can be merged and we can start some Christmas-period-lull _hax0r_ activities to start cutting over and expanding the new API.

Just needs a ✔️ now...

martincostello:
💥

martincostello:
Later today I'll look at making some issues (i.e. tasks) that can be doled out amongst the interested parties to start the switch-over.

PR #432, opened 2018-11-13
Bus should be a property not a field

AnthonySteele:

Bus should be a get-only property not a read-only field
Compliance with analyser rule: https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca1051-do-not-declare-visible-instance-fields
This is shallow fix in order to get the analysers passing, pending better design

#401

### DIFF
--- a/JustSaying/JustSayingFluently.cs
+++ b/JustSaying/JustSayingFluently.cs
@@ -28,7 +28,7 @@ namespace JustSaying
         private readonly ILogger _log;
         private readonly IVerifyAmazonQueues _amazonQueueCreator;
         private readonly IAwsClientFactoryProxy _awsClientFactoryProxy;
-        protected readonly IAmJustSaying Bus;
+        protected IAmJustSaying Bus { get; }
         private SqsReadConfiguration _subscriptionConfig = new SqsReadConfiguration(SubscriptionType.ToTopic);
         private IMessageSerialisationFactory _serialisationFactory;
         private Func<INamingStrategy> _busNamingStrategyFunc;


AnthonySteele:
[WAT](https://github.com/justeat/JustSaying/blob/master/JustSaying.UnitTests/JustSayingFluentlyTestBase.cs#L45)

martincostello:
https://github.com/justeat/JustSaying/blob/1aab3228bea66035cda5b22ac3acba8a8dbd1bc7/JustSaying.UnitTests/JustSayingFluentlyTestBase.cs#L40

AnthonySteele:
That's not wrong.

JustSaying/JustSayingFluently.cs
```diff
@@ -28,7 +28,7 @@ public class JustSayingFluently : ISubscriberIntoQueue, IHaveFulfilledSubscripti
         private readonly ILogger _log;
         private readonly IVerifyAmazonQueues _amazonQueueCreator;
         private readonly IAwsClientFactoryProxy _awsClientFactoryProxy;
-        protected IAmJustSaying Bus { get; }
+        protected IAmJustSaying Bus { get; private set; }
         private SqsReadConfiguration _subscriptionConfig = new SqsReadConfiguration(SubscriptionType.ToTopic);
```
slang25: Could we get away with internal and use InternalsVisibleTo?

martincostello: If the test isn't super important and just belt-and-braces, we should remove it. It's a self-documented cludge, so if it's not adding much value we should just 🗑 it if it's causing us trouble here.

AnthonySteele: The test looks like it's a base class for a lot of things :(

martincostello: Bah. A super-cludge.

slang25: This is the theme in general with a lot of the tests

martincostello: Once #426 gets merged I'll see if I can get time to flex my ex-QA muscles and start slowly re-writing things for #380...

AnthonySteele: @slang25  apprently not, setting it to `protected IAmJustSaying Bus { get; internal set; }` produces

```
JustSayingFluently.cs(31,53,31,56): 
error CS0273: 
The accessibility modifier of the 'JustSayingFluently.Bus.set' accessor 
must be more restrictive than the property or indexer 'JustSayingFluently.Bus'
```

martincostello: `protected internal`?

AnthonySteele: Same.

martincostello: I meant for the whole property, not just the setter.

slang25: Well TIL. https://stackoverflow.com/a/1063920/2225808
You could have a method that sets it, and is internal. 🤔 🤢

slang25: Yeah, `protected internal` for the property is what we want

slang25:
Turns out there are more of those reflection setters 😆

PR #433, opened 2018-11-13
Update .NET Core SDK

martincostello:
* Update to .NET Core SDK 2.1.500.
  * Use new `PackageLicenseExpression` MSBuild property ([docs](https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg-(Technical-Spec)#license-expression-pack)) which causes a compiler warning (`NU5125`) about deprecation with the new SDK if it isn't used.


slang25:
We could also start using the new lock file:
https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#enabling-lock-file
or maybe in a separate PR, up to you

martincostello:
I’d do that separate. I only made the second change because it introduces a warning.

slang25:
It looks like we shouldn't be getting `NU5125` if we remove the licence url, might be worth raising a bug

PR #434, opened 2018-11-16
Fix analyser issues on testing framework code

AnthonySteele:

Fix various analyser issues on testing framework code

#401

### DIFF
--- a/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
+++ b/JustSaying.IntegrationTests/AwsTools/BasicHandlingThrottlingTest.cs
@@ -40,7 +40,7 @@ namespace JustSaying.IntegrationTests.AwsTools
             {
                 await queue.CreateAsync(new SqsBasicConfiguration());
 
-                if (!fixture.IsSimulator)
+                if (!JustSayingFixture.IsSimulator)
                 {
                     // Wait for up to 60 secs for queue creation to be guaranteed completed by AWS
                     using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(1)))
@@ -106,7 +106,7 @@ namespace JustSaying.IntegrationTests.AwsTools
             {
                 do
                 {
-                    if (!fixture.IsSimulator)
+                    if (!JustSayingFixture.IsSimulator)
                     {
                         await Task.Delay(TimeSpan.FromSeconds(5));
                     }


JustSaying.TestingFramework/TestEnvironment.cs
```diff
@@ -37,12 +37,24 @@ public static RegionEndpoint Region
...
+        {
+            get
+            {
+                var awsEnv = Environment.GetEnvironmentVariable("AWS_SERVICE_URL");
+                if (string.IsNullOrWhiteSpace(awsEnv))
+                {
+                    return null;
+                }
+
+                return new Uri(awsEnv);
```
martincostello: Suggest using the constructor that specifies `UriKind.Absolute`.

AnthonySteele:
I wanted to enable most analyser rules on the code, but it seems that we aren't ready for that yet. Getting a clean bill of health from the analyser on e.g. `JustSaying.csproj` is when it _starts_ analysing `JustSaying.TestingFramework`, and when that's clear then the 250+ messages come in from the unit tests.

It might be possible soon to enable analysers with only a few rules disabled, on the non-test projects.

JustSaying.TestingFramework/TestEnvironment.cs
AnthonySteele: Is that done to prevent invalid data? Should it also be done on other code that calls `new Uri` ? or is it just a concern in this case since Env vars are more prone to bad/missing data?

martincostello:
There's also a level of pragmatism to apply to analyzers with regards to the test code. For example, requiring `ConfigureAwait(...)` just adds visual noise. I think it's fine for the test _infrastructure_ parts, but in the tests themselves it's not necessary.

In this lib I turned off that rule, plus the ban on underscores and something else that I can't remember what it is off the top of my head.

https://github.com/justeat/httpclient-interception/blob/6756febbd23879077bea037c24a5b7014c8b4fc9/tests/HttpClientInterception.Tests/JustEat.HttpClientInterception.Tests.csproj#L4

AnthonySteele:
Right, [what I did here](https://github.com/NuKeeperDotNet/NuKeeper) Is to use a second, more forgiving ruleset for the tests.

martincostello:
I find one rule set with overrides per-project via `NoWarn` easier to maintain than files which are 90% the same, but whichever.

JustSaying.TestingFramework/TestEnvironment.cs
martincostello: Basically yes, it flushes out misconfiguration earlier than at the point it gets consumed by something trying to do a request. I do it most times I create a `Uri` it as it's very rare I'd ever need a relative URI. Depends on context obviously, but in this case I can't see any reason why you'd be pointing at a relative URI for an AWS resource.

martincostello:
Suggest running `Build.ps1` locally before committing to check it actually compiles, rather than relying on the CI to do it.

PR #436, opened 2018-11-16
Analysers in place

AnthonySteele:

Code analysers are in place
There are exclusions, globally in `CodeAnalysisRules.ruleset` and in individual test projects.

It's not perfect yet, but it's a big step in the right direction to change from "no analysers automatically run, checks performed as and when someone feels like it" to "analysers are automatically run, with significant exclusions". The exclusions can be pared down to the minimum in following steps.

In particular, there is work pending ( #403 )  on `IDisposable` and use of Disposable resources, where the code currently sets off analyser errors ([CA1001](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1001-types-that-own-disposable-fields-should-be-disposable) )  for valid reasons, but this is being addressed in another PR.

#401 - Use code analysers

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,17 +1,18 @@
 <Project>
   <Import Project="version.props" />
+  <PropertyGroup>
+    <AnalyzersPackageVersion>2.6.2</AnalyzersPackageVersion>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysisRules.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
     <PackageReference Include="OpenCover" Version="4.6.519" PrivateAssets="All" />
     <PackageReference Include="ReportGenerator" Version="3.1.2" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.2" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="2.6.2" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.NetCore.Analyzers" Version="2.6.2" PrivateAssets="All" />
-    <PackageReference Include="Text.Analyzers" Version="2.6.2" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.NetCore.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
+    <PackageReference Include="Text.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
   </ItemGroup>
-  <PropertyGroup>
-    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysisRules.ruleset</CodeAnalysisRuleSet>
-  </PropertyGroup>
   <PropertyGroup>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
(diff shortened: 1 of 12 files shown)


Directory.Build.props
```diff
@@ -4,9 +4,15 @@
...
     <PackageReference Include="OpenCover" Version="4.6.519" PrivateAssets="All" />
     <PackageReference Include="ReportGenerator" Version="3.1.2" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.2" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="2.6.2" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.NetCore.Analyzers" Version="2.6.2" PrivateAssets="All" />
+    <PackageReference Include="Text.Analyzers" Version="2.6.2" PrivateAssets="All" />
   </ItemGroup>
   <PropertyGroup>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <CodeAnalysisRuleSet>..\CodeAnalysisRules.ruleset</CodeAnalysisRuleSet>
```
martincostello: This only works in the root directory. Use this instead:

```xml
<CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysisRules.ruleset</CodeAnalysisRuleSet>
```

Directory.Build.props
```diff
@@ -4,9 +4,15 @@
...
+  </PropertyGroup>
+  <PropertyGroup>    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
```
martincostello: Indenting.

Directory.Build.props
```diff
@@ -4,9 +4,15 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
     <PackageReference Include="OpenCover" Version="4.6.519" PrivateAssets="All" />
     <PackageReference Include="ReportGenerator" Version="3.1.2" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.2" PrivateAssets="All" />
```
martincostello: I've started parameterising the version in other repos because they update in sync, so it's only one thing to change, rather than 4.

JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;CA2007;CA1054;CA1307;CA1063;CA1822;CA1052;CA1034;CA1816</NoWarn>
```
martincostello: Did you miss prefixes on 1701 and 1702?

martincostello: Should prefix with `$(NoWarn);`, otherwise you'll overwrite any warnings configured to be ignored elsewhere.

JustSaying.UnitTests/JustSaying.UnitTests.csproj
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;CA2007;CA1051;CA1707;CA1034;CA1812;CA1307</NoWarn>
```
martincostello: As above comments.

JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
AnthonySteele: Hmm, I did that via the VS GUI, and those values were already there. I assumed that they were pre-existing, but it look more like it's a default generated when you write that field. Safe to remove though, so I have done so.

Directory.Build.props
AnthonySteele: I'm not such a fan of that abstraction, as it makes life harder for any build tools that might try to red and modify it.

martincostello: Up to you, but it's a pattern with MS precedent (kind of where I got the idea to use it myself): https://github.com/aspnet/HttpClientFactory/blob/master/build/dependencies.props

PR #437, opened 2018-11-16
Add deserialization of incoming sqs message attributes

dv00d00:
I will try to add the ability to allow client code to access incoming SQS messages attributes. There is a high demand for this feature in my and neighboring teams to read the `ApproximateReceiveCount`. Currently it is being populated if you set the `IMessageBackoffStrategy` on the subscription config, but is not accessible for the client code.

Would be great if it could appear in the 6.* branch

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -49,23 +49,6 @@ namespace JustSaying.AwsTools.MessageHandling
             {
                 typedMessage = _serialisationRegister.DeserializeMessage(message.Body);
 
-                typedMessage.MessageAttributes = message.MessageAttributes?.ToDictionary(
-                    source => source.Key,
-                    source =>
-                    {
-                        if (source.Value == null)
-                        {
-                            return null;
-                        }
-
-                        return new JustSaying.Models.MessageAttributeValue
-                        {
-                            StringValue = source.Value.StringValue,
-                            BinaryValue = source.Value.BinaryValue?.ToArray(),
-                            DataType = source.Value.DataType
-                        };
-                typedMessage = _serialisationRegister.DeserializeMessage(message.Body);
-
                 Models.MessageAttributeValue MapAttribute(MessageAttributeValue sourceAttribute) =>
                     new Models.MessageAttributeValue
                     {
@@ -73,7 +56,7 @@ namespace JustSaying.AwsTools.MessageHandling
                         BinaryValue = sourceAttribute.BinaryValue?.ToArray(),
                         DataType = sourceAttribute.DataType
                     };
-                
+
                 typedMessage.MessageAttributes = message.MessageAttributes?.ToDictionary(
                     source => source.Key,
                     source => source.Value == null ? null : MapAttribute(source.Value));
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Amazon.SQS;
 using Amazon.SQS.Model;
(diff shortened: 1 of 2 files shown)


dv00d00:
I'll surely add some, but there are several questions.

* Currently it will address only `ApproximateReceiveCount`, not sure if anything else would be able to get in.
* Before this change the only use case for the `MessageAttributes` was to specify something on publish, and there are different sets of incoming/outgoing attributes.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -47,6 +48,23 @@ public async Task DispatchMessage(SQSMessage message)
...
+                    {
+                        if (source.Value == null)
+                        {
+                            return null;
+                        }
+
+                        return new JustSaying.Models.MessageAttributeValue
+                        {
+                            StringValue = source.Value.StringValue,
+                            BinaryValue = source.Value.BinaryValue?.ToArray(),
```
martincostello: Correct that null goes to null, rather than null goes to empty array?

dv00d00: yep, null will be propagated here

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -47,6 +48,23 @@ public async Task DispatchMessage(SQSMessage message)
             try
             {
                 typedMessage = _serialisationRegister.DeserializeMessage(message.Body);
+
+                typedMessage.MessageAttributes = message.MessageAttributes?.ToDictionary(
+                    source => source.Key,
+                    source =>
+                    {
```
AnthonySteele: A bit big for an inline lambda - extract a private static method for this?

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -47,6 +48,23 @@ public async Task DispatchMessage(SQSMessage message)
...
+                            return null;
+                        }
+
+                        return new JustSaying.Models.MessageAttributeValue
+                        {
+                            StringValue = source.Value.StringValue,
+                            BinaryValue = source.Value.BinaryValue?.ToArray(),
+                            DataType = source.Value.DataType
+                        };
+                    });
```
slang25:
```suggestion
                typedMessage = _serialisationRegister.DeserializeMessage(message.Body);

                Models.MessageAttributeValue MapAttributes(MessageAttributeValue sourceAttribute) =>
                    new Models.MessageAttributeValue
                    {
                        StringValue = sourceAttribute.StringValue,
                        BinaryValue = sourceAttribute.BinaryValue?.ToArray(),
                        DataType = sourceAttribute.DataType
                    };
                
                typedMessage.MessageAttributes = message.MessageAttributes?.ToDictionary(
                    source => source.Key,
                    source => source.Value == null ? null : MapAttributes(source.Value));
```

slang25: I clearly haven't got the hang of this suggestion thing

dv00d00:
> You can request to get "All" attributes, or some by name by passing a list. We would probably want it configurable in the fluent api.

Maybe as a separate pr?

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
dv00d00: I smell F#, thx for the snippet 👍

dv00d00:
Ok `ApproximateReceiveCount` will be on the `Attributes` property not `MessageAttributes`.
`MessageAttributes` would contain attribute values user has defined. Still useful, but not really relevant to my needs.

slang25:
This comes back to a recurring theme with JustSaying, which is we sometimes want to grab things through the abstraction.

I don't like making the `Message` base class look like the aws types, because then it's not an abstraction, you are forever mirroring the aws sdk, and it dirties your domain. We've talked before about having "message envelopes", or something as a way to get the full aws message. I like the idea of being able to take a context object in your handler which could expose some of these thing to you, while keeping the concept of domain messages completely separate.

It might be worth tackling this problem sooner than later, otherwise we continue to slip into a mess.

My suggestion would be to add an interface like this:
```csharp
public interface IMessageContextAccessor
{
    MessageContext MessageContext { get; }
}
```
(totally stealing the design from `IHttpContextAccessor`)
which you'd optionally take into your constructor. Here we'd give you all the AWS SDK goodness, and the full message itself.
We'd need to make a breaking change to `IHandlerResolver`, but this is IMO better than breaking existing consuming handlers, or `Message`

AnthonySteele:
message envelopes _might_ happen some time in some form. Will keep you updated.

PR #438, opened 2018-11-18
.NET Core 2.2.0 support

martincostello:
* Update to .NET Core SDK `2.2.100`.
  * Update tests to target `netcoreapp2.2`.
  * Use the [new NuGet symbol package format](https://blog.nuget.org/20181116/Improved-debugging-experience-with-the-NuGet-org-symbol-server-and-snupkg.html "Improved package debugging experience with the NuGet.org symbol server").
  * Update `ReportGenerator` to latest release.


martincostello:
I thought it was just a different package (`.snupkg`) instead of the current additionally generated symbols file `symbols.nupkg`. I did link to the same doc in the description.

It's not otherwise changing anything else? There's no PDB files inside the existing `.nupkg` files from what I can see changing them to `.zip` and opening them with Windows Explorer.

slang25:
Ah cool, I didn't realised it was the same page 👍 

The current artifacts are a bit misleading, in that we don't actually do anything with the symbols.nupkg files. The `DebugType` is `embedded` which means they are portable and embedded into the assembly itself. So we have a russian doll sorta thing which goes nupkg -> assembly -> symbols -> source.

Embedding barely changes the size of the assembly as well, but some purists have convinced Microsoft it's a big enough issue not to be the long term solution. You can tell I'm a bit salty about the decision.

AnthonySteele:
So,merge this after SDK version `2.2.100` is officially out?

martincostello:
It should be safe to merge once the packages get to NuGet.org.

PR #439, opened 2018-11-19
Switch CA1001 on

slang25:
The disposable fields rule is quite valuable to leave enabled to prevent new occurrences of this kind.
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1001-types-that-own-disposable-fields-should-be-disposable?view=vs-2017


martincostello:
Would it be better to leave this until after #403? Then we don't need the suppressions.

slang25:
We still need it in `Throttled.cs`, so the ordering doesn't matter much. I created this cause I was expecting 403 to remove a suppression.

adammorr:
Want me to just add this to that PR?

slang25:
Given the whitespace issue, I think this will be easiest for all 👍

slang25:
The ironic thing is that it's `SqsNotificationListener` that holds an instance of `Throttled` anyway, so it still holds disposable fields

adammorr:
@slang25 https://github.com/justeat/JustSaying/pull/403/commits/6641a146e050ec8ef5d34ad281661598286a5098

Issue #422, opened 2018-11-07
Threading review of SqsNotificationListener and Throttled Message Processing Strategy


slang25:
This has come up in recent discussions, so here is a braindump so that we don't lose this.

Looking at `IMessageProcessingStrategy`, it has the following members:

```csharp
int MaxWorkers { get; }
int AvailableWorkers { get; }
void StartWorker(Func<Task> action);
Task WaitForAvailableWorkers();
```

Imagine we have the default `Throttled` implementation, with 10 max workers, and 6 workers in-flight (invoking handlers).

When `SqsNotificationListener` wants to get messages it goes through the following process:

1. Is `AvailableWorkers` > 0? (if not `await` `WaitForAvailableWorkers`)
   Here, yes, we have 4 available. We can continue.
2. Long poll SQS for a maximum of 4 messages. We will either receive 4 messages, or wait a maximum of 20 second to receive up to 4 messages.
3. Foreach message, call `StartWorker` with the handler (wrapped in a few bits and bobs).
4. Go to step 1 immediately (while the handlers are inflight).
1. (Step 1 again) Is `AvailableWorkers` > 0? No, now `await` `WaitForAvailableWorkers`

In isolation, this process is alright, and seems sensible. Where it gets tricky is when you share the `IMessageProcessingStrategy` and have multiple `SqsNotificationListener` going through this same workflow.
However, it is quite likely you will want to share an instance of `IMessageProcessingStrategy`, as you might want a concurrency level across multiple queues, or for your whole application.

There's a few problems with the `IMessageProcessingStrategy` interface as it stands when you have multiple listeners interacting with it:
- When all workers are busy, you will likely have multiple listeners all waiting at the `WaitForAvailableWorkers` step.
  They will all be released simultaneously and race.
  They will all read the `AvailableWorkers` value, and likely see the same number (lets say 1).
  Once they receive the message, they will all try to call `StartWorker`, the first succeeding.
  In an old version of JustSaying (early 4), this race condition would cause `Throttled` to lose count, which wouldn't end well.
  In recent version, subsequent attempts to call `StartWorker` will block until there are available workers. (Blocking threadpool threads is bad)
- There is a race condition where listeners are release from awaiting `WaitForAvailableWorkers`, then they check the `AvailableWorkers` count to know how many messages to request from SQS.
  When reading `AvailableWorkers`, this value could have changed and now might be 0, in which case an exception is thrown, logged and the loop continues.

[...]

PR #440, opened 2018-11-19
Remove blocking in IMessageProcessingStrategy

slang25:
... and tidied up the background launch of the ListenLoop.

This PR addresses the concern in #422

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -143,6 +143,10 @@ namespace JustSaying.AwsTools.MessageHandling
                     {
                         foreach (var message in sqsMessageResponse.Messages)
                         {
+                            if (ct.IsCancellationRequested)
+                            {
+                                return;
+                            }
                             await HandleMessage(message, ct).ConfigureAwait(false);
                         }
                     }
--- a/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -177,7 +177,7 @@ namespace JustSaying.AwsTools.MessageHandling
                     {
                         foreach (var message in sqsMessageResponse.Messages)
                         {
-                            HandleMessage(message, ct);
+                            await HandleMessage(message, ct).ConfigureAwait(false);
                         }
                     }
                 }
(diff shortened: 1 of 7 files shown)


slang25:
Following this I'll rebase #426

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -177,7 +143,7 @@ internal async Task ListenLoop(CancellationToken ct)
                     {
                         foreach (var message in sqsMessageResponse.Messages)
                         {
-                            HandleMessage(message, ct);
+                            await HandleMessage(message, ct).ConfigureAwait(false);
```
adammorr: Something that I think was missed with my previous PR but we should check the cancellation token before calling `HandleMessage(...)` in the message loop

AnthonySteele: So if the worker gets 2 messages off the queue, Handles 1 of them and then notices that `ct.IsCancellationRequested` is true (e.g. during scale down), it will not process the second one.  So then the message is retrieved but not deleted, and will soon time out and end up back on the queue for other worker to  take?

JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -1,85 +1,83 @@
...
+            try
+            {
+                await _semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                return;
+            }
+
+            _ = Task.Run(() => ReleaseOnCompleted(action));
```
AnthonySteele: Just to confirm:

It should still be the case that `StartWorker` and `HandleMessage` returning doesn't mean that the message processing has been completed, just that it has been started  (that's the `Task.Run`  here ) and so when we loop over a list of messages here, we're usually launching them in parallel. I think this is still true. Although the semaphore can introduce latency between retrieving a message and processing it.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
slang25: Exactly correct. The other option is to try to commit to whatever messages we've peeked at from the queue, not sure which is the best option. Your thoughts?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -94,53 +94,19 @@ public void Listen(CancellationToken cancellationToken)
...
-                .Unwrap()
-                .ContinueWith(t => LogTaskEndState(t, queueInfo, _log));
+            // Run task in background
+            // ListenLoop will cancel gracefully, so no need to pass cancellation token to Task.Run
+            _ = Task.Run(async () =>
+            {
+                await ListenLoop(cancellationToken).ConfigureAwait(false);
+                IsListening = false;
+                _log.LogInformation($"Stopped Listening - {queueInfo}");
+            });
```
shaynevanasperen: 💯 Nice!

PR #444, opened 2018-11-21
don't use floating Version numbers

AnthonySteele:

don't use floating Version numbers.
We were using them in _some_ places and not others 
which was causing issues when building in Visual Studio (errors about two analyser assembly versions)
Rather insist on a very recent version

### DIFF
--- a/JustSaying/JustSaying.csproj
+++ b/JustSaying/JustSaying.csproj
@@ -6,9 +6,9 @@
     <ProjectReference Include="..\JustSaying.Models\JustSaying.Models.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="3.3.25.3" />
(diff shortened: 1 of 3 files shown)


JustSaying/JustSaying.csproj
```diff
@@ -6,9 +6,9 @@
     <ProjectReference Include="..\JustSaying.Models\JustSaying.Models.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="3.3.25.3" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.3.1.11" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.3.3.19" />
+    <PackageReference Include="AWSSDK.Core" Version="3.3.29.12" />
```
martincostello: I thought we'd agreed to leave these ones as-is indefinitely?

AnthonySteele:
Issue that we ware trying to fix

justsayinganlysersbuildfail

PR #445, opened 2018-11-21
Fix build quirks around Sourcelink and symbols package

slang25:
We currently use embedded source for JustSaying, which I think is appropriate for this library. This PR removes the separate symbols package (which isn't used), and removes the sourcelink metadata.

Originally I felt that the addition of sourcelink while redundant was harmless, but it causes build issues for me due the native libgit2sharp dependencies on my mac. Removing it also avoids future maintainers (and ourselves) from confusing how it currently works.

I want to double check the artifacts before removing the WIP prefix.


martincostello:
Downside to removing the package is we lose the embedded Git SHA on the assembly attributes (right?).

That'd mean you _have the source_ but won't know _what version of the source_ it is.

Not fussed about removing the symbols package, but I think we'd still want the benefits of the embedded attributes (without needing to use home-rolled targets to do it).

martincostello:
Well it compiles OK for Mac in Travis (see #446).

slang25:
You raise a good point, we could do with adding a commit hash

slang25:
Borrowed from out previous efforts in [httpclient-interception](https://github.com/justeat/httpclient-interception/blob/6756febbd23879077bea037c24a5b7014c8b4fc9/Directory.Build.targets):
```csharp
[assembly: AssemblyMetadata("BuildTimestamp", "2018-11-21T17:59:38Z")]
[assembly: AssemblyMetadata("CommitHash", "e064df26f8accb26c3cceb6615aec4fecfe0bcd0")]
[assembly: AssemblyMetadata("CommitBranch", "symbols-tidy")]
```

shaynevanasperen:
@slang25 I'm curious, when is it _not appropriate_ to embed source? Why wouldn't everyone just do this instead of using sourcelink?

slang25:
@shaynevanasperen that's my view too, but there are vocal MS customers who have said it doesn't work for them. I think it should be on by default, and if you want rid of it, you have an optimizer that strips it, but I'm not in charge 😄 

Deploying to Xamarin, you might really want to squeeze the assembly size for example (again, an optimizer would solve it). But after that I'm really struggling

slang25:
More info: https://github.com/dotnet/sdk/issues/2679

AnthonySteele:
I have worked with `libgit2sharp` before and it can be tricky - it's a wrapper over native C code binaries for `libgit2`, and will have platform-specific dependencies, e.g. `libcurl3` or `libcurl4`.

martincostello:
This seems more of a Rider bug to me. It's working fine on the command line as shown by #446, so it's not as cut-and-dried as _"doesn't work on macOS"_.

slang25:
Agreed, I'm sure I could fine the conditions that rider runs under to replicate the issue, and there will be workarounds. But given its unnecessary, it's simplist to remove it.

martincostello:
Just seems a shame to remove the metadata linkage between the source and its location in GitHub, even if the source is embedded as well.

slang25:
I know of 2 ways to view the sourcelink metadata, one is the dotnet tool, the other is using dotPeek/Rider and drilling into:
 `Assembly` `->` `JustSaying` `->` `Metadata` `->` `Embedded Portable PDB Metadata` `->` `CustomDebugInformation` `->` `Source Link`
which contains:
```json
{"documents":{"C:\\projects\\justsaying\\*":"https://raw.githubusercontent.com/justeat/JustSaying/bea4806f03e39d7a64e77934146a3475e7762fe2/*"}}
```

I'm not sure that's any more discoverable than decompiling the source and looking at the assembly attributes, it requires more knowledge about how this stuff works.

Unless I'm missing an obvious way to view the sourcelink info?

slang25:
I'm just as happy to reduce this PR to removing the symbols package and apply a workaround like this:
https://github.com/enricosada/falanx/commit/d525c2ef6cc1d13fc405990864e6a254d06400b7

I'll do whichever we think is better.

martincostello:
I _think_ I'd prefer that we just ditched the symbols package, plus a workaround for your Rider debugging experience, rather than throw it away completely. Then we don't need the custom targets file.

PR #446, opened 2018-11-21
Test on macOS and Linux

martincostello:
CI should test for both macOS and Linux, not just Linux. Otherwise if we have build issues (see #445), we might not find them.

Depending on the outcome from the tests for the first commit, might need to tweak things slightly to not run the integration tests on macOS if the Docker image doesn't work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: required
 dist: trusty
 
+os:
+  - linux
+  - osx
+
 env:
   global:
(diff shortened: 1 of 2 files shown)


martincostello:
Yeah, Docker isn't installed...

slang25:
Great stuff. My particular issue only happens with a design time build in Rider (but is particularly nasty because I have to restart it to fix it), so it probably won't catch that one.

slang25:
image

PR #447, opened 2018-11-21
[WIP] Publish Envelope

AnthonySteele:

For comparison with #449 

This is a step towards the "envelopes" proposal that separates out message metadata from the body. In this PR it is only on the `Publish` path, as the Publish path is a simpler first step.

Earlier investigation has raised that the envelope metadata used to configure a publish is different to the the metadata received with the message at the other end of the process. So these can be approached separately.

There are two fields on the `Message` base class that are optional and are used only on `Publish`. they are `DelaySeconds` and `MessageAttributes`. They are important, but are used in a minority of cases.

These are removed from the `Message` base class, and are on the `PublishEnvelope` class instead.
We now are facing more overloads - with/without an Envelope, with/without the `CancellationToken` so the overloads are moved to extension methods in `MessagePublisherHelpers`.

So the signature for publish methods is in this case 

```csharp
Task PublishAsync(PublishEnvelope envelope, CancellationToken cancellationToken)
````
and the `envelope` always contains the message, along with other optional data.  There are overloads/extension methods to handle simpler cases and forward them to the most general case.

This should not affect the "wire format" of messages sent.

please review, naming etc is still up for discussion.

### CODE REVIEW

**Comment by martincostello** [comment]
My understanding of default interfaces was that they require runtime support, so aren't usable by .NET 4.8.

File: JustSaying/Messaging/IMessagePublisher.cs [hunk 1]
```diff
@@ -1,13 +1,27 @@
...
 {
     public interface IMessagePublisher
     {
-        Task PublishAsync(Message message);
+        Task PublishAsync(PublishEnvelope message, CancellationToken cancellationToken);
+    }
+
+    public static class MessagePublisherHelpers
+    {
+        public static async Task PublishAsync(this IMessagePublisher publisher, Message message)
```
slang25: Nice 👍

JustSaying/Messaging/IMessagePublisher.cs
```diff
@@ -1,13 +1,27 @@
...
 
 namespace JustSaying.Messaging
 {
     public interface IMessagePublisher
     {
-        Task PublishAsync(Message message);
+        Task PublishAsync(PublishEnvelope message, CancellationToken cancellationToken);
+    }
+
+    public static class MessagePublisherHelpers
```
slang25: Stealing tips from @martincostello: 
```csharp
[EditorBrowsable(EditorBrowsableState.Never)]
```

JustSaying.Models/Message.cs
```diff
@@ -3,6 +3,19 @@
 
 namespace JustSaying.Models
 {
+    public class PublishEnvelope
```
martincostello: Just checking, this is a _publish-only_ thing? So a subscriber would never interact with an envelope? If they do, I'd suggest `MessageEnvelope`.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -37,19 +37,16 @@ protected SnsTopicBase(IMessageSerialisationRegister serialisationRegister, ILog
...
             _snsWriteConfiguration = snsWriteConfiguration;
             _messageSubjectProvider = messageSubjectProvider;
         }
 
         public abstract Task<bool> ExistsAsync();
         
-        public Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
-        public async Task PublishAsync(Message message, CancellationToken cancellationToken)
+        public async Task PublishAsync(PublishEnvelope env, CancellationToken cancellationToken)
```
slang25: Perhaps optional CT? I can imagine a lot of people rely on the timeout and won't want to pass one in.

JustSaying.Models/Message.cs
```diff
@@ -3,6 +3,19 @@
...
+    {
+        public PublishEnvelope(Message message)
+        {
+            Message = message ?? throw new ArgumentNullException(nameof(message));
+        }
+
+        public Message Message { get; }
+
+        public int? DelaySeconds { get; set; }
```
martincostello: `TimeSpan` this as per the previous changes in an earlier PR?

JustSaying.Models/Message.cs
```diff
@@ -3,6 +3,19 @@
...
+        public IDictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
```
martincostello: Might be nice for the integrator if we initialized this by default so they can just start adding to it immediately, rather than having to initialize it themselves every place it's used.

JustSaying.IntegrationTests/JustSayingFluently/MultiRegion/WithSqsPointToPointSubscriber/WhenAFailoverRegionIsSetup.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using JustSaying.Messaging;
```
slang25: Can we put the extension method in the same namespace as the publish method so it's source code compatible?

JustSaying.UnitTests/AwsTools/MessageHandling/Sns/TopicByName/WhenPublishingAsync.cs
```diff
@@ -42,19 +43,21 @@ protected override Task Given()
...
+                            {
                                 StringValue = MessageAttributeValue,
                                 DataType = MessageAttributeDataType
                             }
                     }
                 }
-            });
+            };
+
+            await SystemUnderTest.PublishAsync(publishData, CancellationToken.None);
```
martincostello: An overload (extension method) that accepts the message itself and wraps it in an envelope for you would be nice.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
martincostello: I'd rather `envelope`.

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -26,19 +27,17 @@ public class SqsPublisher : SqsQueueByName, IMessagePublisher
             _serialisationRegister = serialisationRegister;
         }
 
-        public Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
-        public async Task PublishAsync(Message message, CancellationToken cancellationToken)
+        public async Task PublishAsync(PublishEnvelope env, CancellationToken cancellationToken)
```
martincostello: `envelope`.

JustSaying/JustSayingBus.cs
```diff
@@ -130,12 +130,10 @@ public void Start(CancellationToken cancellationToken = default)
             }
         }
 
-        public Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
-        public async Task PublishAsync(Message message, CancellationToken cancellationToken)
+        public async Task PublishAsync(PublishEnvelope env, CancellationToken cancellationToken)
```
martincostello: We should definitely use proper names on the public interfaces, so `envelope`.

JustSaying/JustSayingFluently.cs
```diff
@@ -158,25 +158,19 @@ public void StartListening(CancellationToken cancellationToken = default)
...
-        /// <param name="message"></param>
-        public virtual Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
         /// <summary>
         /// Publish a message to the stack, asynchronously.
         /// </summary>
         /// <param name="message"></param>
         /// <param name="cancellationToken"></param>
-        public virtual async Task PublishAsync(Message message, CancellationToken cancellationToken)
+        public virtual async Task PublishAsync(PublishEnvelope env, CancellationToken cancellationToken)
```
martincostello: `envelope`.

JustSaying/JustSayingFluently.cs
```diff
@@ -158,25 +158,19 @@ public void StartListening(CancellationToken cancellationToken = default)
...
-        /// <summary>
-        /// Publish a message to the stack, asynchronously.
-        /// </summary>
-        /// <param name="message"></param>
-        public virtual Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
         /// <summary>
         /// Publish a message to the stack, asynchronously.
         /// </summary>
         /// <param name="message"></param>
```
martincostello: `envelope`.

JustSaying/Messaging/IMessagePublisher.cs
martincostello: If these are all extensions, the type name should be `(I)MessagePublisherExtensions`.

slang25:
This looks good, however I don't think it warrants the symmetry of doing the same for handlers (I know I'm getting ahead of myself). I have a little POC locally to illustrate a competing idea, not saying it's right, just worth exploring a few options IMO.

JustSaying.IntegrationTests/JustSayingFluently/MultiRegion/WithSqsPointToPointSubscriber/WhenAFailoverRegionIsSetup.cs
AnthonySteele: I would like to make that easier. Right now this PR has a bout 20 files doing `using JustSaying.Messaging;` so that they can use these overloads/extensions. But it is in the same namespace as `IMessagePublisher.PublishAsync`

JustSaying/Messaging/IMessagePublisher.cs
```diff
@@ -1,13 +1,27 @@
...
+    {
+        public static async Task PublishAsync(this IMessagePublisher publisher, Message message)
+        {
+            await publisher.PublishAsync(new PublishEnvelope(message), CancellationToken.None)
```
martincostello: Could just call the other overload?

martincostello: These should `null` check on `publisher`.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
AnthonySteele: Covered by the extensions

martincostello: I was mulling this over in my review, but this might/could be catered for by extension method(s).

JustSaying.Models/Message.cs
AnthonySteele: I am in two minds of doing that, since in the majority of cases, it won't be used and creating the dictionary is therefor a wasted allocation.

JustSaying.Models/Message.cs
shaynevanasperen: I think it might cause confusion if the envelope on the receive side is named differently from the one on the publish side. Would they have different members?

slang25: They would likely have different members, such as "approximate receive count". So I think it makes sense to have different versions, but keep the pattern the same.

JustSaying.Models/Message.cs
martincostello: Suggest here as part of the bigger overhaul.

AnthonySteele:
There is another route to go here, and that is a seperate object, e.g.
```
public async Task PublishAsync(Message message, PubishMetadata metadata = null, CancellationToken ctx = CancellationToken.None)
```

We can go that route if people prefer it.

JustSaying.Models/Message.cs
AnthonySteele: They would have different members and/or implementation of  such, yes.

slang25:
I really like that option, it feels cleaner

AnthonySteele:
The real goal here is: remove Publish metadata from the `Message` base class, in preparation for thinking about similar with respect to Subscribing.

AnthonySteele:
The intent is: On the `Message` base class, remove metadata that is only relevant to a Publish and put that on a separate object.  Do as much of this as is possible without changing the "wire format". `int? DelaySeconds` is a good example of publish metadata, as it is an instruction on how to publish the message This is in preparation for the similar but larger task of doing same with respect to Subscribing.

In subscribing we particularly want to go from a "fat" message base class, to a way to get at the complete message metadata that AWS will supply, but that is a later step.

### Option "envelope"

Now we have 2 related option on publishing. The First is detailed in this PR, and using it would be like

```csharp
var envelope =  new PublishEnvelope(message);
envelope.DelaySeconds = 3;
await publisher.PublishAsync(envelope, CancellationToken.None)
```

Along with various overloads and helpers to make this easier for the default case where the envelope contains nothing but the message.

### Option "metadata"

The second option is a separate object for publish metadata instead of a wrapper

```csharp
var metadata = new PublishMetadata();
metadata.DelaySeconds = 3;
await publisher.PublishAsync(message, metadata, CancellationToken.None)
```

Along with various overloads and helpers to make this easier for the default case where the metadata contains nothing.

martincostello:
My initial thoughts are to prefer the envelope approach, as the other would seem at first glance to require more overloads to maintain to keep it like-for-like for consumption just using `Message` classes.

dv00d00:
Dump of thoughts:

In terms of publishing envelope we should consider if the message is published to SNS topic where those attributes would not make much sense (as far as I'm concerned) or to SQS queue where this makes much sense.
Does this mean we will have to distinguish SNS/SQS envelope?

JustSaying is designed to work with AWS SDK. Specifically with SQS/SNS combo.

We are abstracting underlying infrastructure, and will have to extend contracts each time AWS will introduce another feature in SQS or SNS. 

There is a direct transformation or map from Message to `SNSPublishRequest` or `SQSPublishRequest`, I was thinking maybe it would be a little bit more resilient to changes in AWS SDK if we offer some sort of user defined manipulation on SNS/SQS requests, something like middleware.

JustSaying/Messaging/MessagePublisherHelpers.cs
```diff
@@ -0,0 +1,29 @@
...
+using JustSaying.Models;
+
+namespace JustSaying.Messaging
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class MessagePublisherHelpers
+    {
+        public static Task PublishAsync(this IMessagePublisher publisher, Message message)
+        {
+            return PublishAsync(publisher, message, CancellationToken.None);
```
martincostello: Should call this like an extension method:

```csharp
return publisher.PublishAsync(message, CancellationToken.None);
```

JustSaying/Messaging/IMessagePublisher.cs
AnthonySteele: I agree. Nothing else to add, so I am resolving this conversation.

AnthonySteele:
Dmitry has a good point: Publishing happens one of 2 ways

* Publishing to SQS happens in `SqsPublisher` where it builds a `Amazon.SQS.Model.SendMessageRequest` which has `DelaySeconds` but no `MessageAttributes`

* What's more common is the publish to SNS, which happens in `SNSTopicBase`, where it builds a `Amazon.SimpleNotificationService.Model.PublishRequest` which has `MessageAttributes` but no `DelaySeconds`.

So the `PublishEnvelope` is a collection of these two properties,  the union of what is needed in 2 different cases. And actually it doesn't overlap.

Still, I'd rather section this problem into something that is clearly only about publishing, rather than have it on the base class of ever message sent or received.

What would the middleware look like? allow the user to supply optional `Action<PublishRequest>` and `Action<SendMessageRequest>` in order to transform them? or `Func<PublishEnvelope, PublishRequest>` to generate them?

Would this replace or add to the `DelaySeconds` or `MessageAttributes` properties?

dv00d00:
After some lurking in codebase found this:

This is what available for publishing:

```csharp
sns request

public string Message { get; set; } 
public Dictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
public string MessageStructure { get; set; } // not exposed
public string PhoneNumber { get; set; } // not exposed
public string Subject { get; set; }  // implicit
public string TargetArn { get; set; } // not exposed
public string TopicArn { get; set; } // implicit

sqs request

public int DelaySeconds { get; set; } 
public Dictionary<string, MessageAttributeValue> MessageAttributes { get; set; } // exposed but not used in sqs path
public string MessageBody { get; set; } 
public string MessageDeduplicationId { get; set; } // not exposed
public string MessageGroupId { get; set; } // not exposed
public string QueueUrl { get; set; } // implicit
```

Not sure that phone stuff is particularly interesting but `MessageGroupId`  and `MessageDeduplicationId ` are very interesting for fifo cases


As for middleware, (or maybe a hook ?) It would be usefull for it to have shape of 

* `Func<TSource, SNSPublishMessage, SNSPublishMessage,>`
* `Func<TSource, SQSPublishMessage, SQSPublishMessage,>`

What about adding some builder overloads when registering publishers to use one of these?

dv00d00:
Also, based on the HttpRequestMessage we could introduce some store for user state (Dictionary<string, object> ? ) which would not be transferred on the wire, but could be used during interception phase.

AnthonySteele:
Sample code for simple user transform in envelope, actioned by `SnsTopicBase`:

```csharp
var request= new PublishRequest
{
	TopicArn = Arn,
	Subject = messageType,
	Message = messageToSend,
	MessageAttributes = BuildMessageAttributes(envelope)
};

if (envelope.SnsTransformer != null)
{
	request = envelope.SnsTransformer.Invoke(request);
}

return request;```

File: JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs [hunk 15]
```diff
@@ -74,34 +75,48 @@ private PublishRequest BuildPublishRequest(PublishEnvelope envelope)
...
-                    {
-                        StringValue = source.Value.StringValue,
-                        BinaryValue = binaryValueStream,
-                        DataType = source.Value.DataType
-                    };
-                });
+            var messageAttributeValues = BuildMessageAttributes(envelope);
             
-            return new PublishRequest
+            var request= new PublishRequest
```
martincostello: Missing space before `=`.

JustSaying/Messaging/PublishEnvelope.cs
```diff
@@ -0,0 +1,20 @@
...
+        public PublishEnvelope(Message message)
+        {
+            Message = message ?? throw new ArgumentNullException(nameof(message));
+        }
+
+        public Message Message { get; }
+
+        public TimeSpan? Delay { get; set; }
+
+        public IDictionary<string, MessageAttributeValue> MessageAttributes { get; } = new Dictionary<string, MessageAttributeValue>();
```
slang25: `MessageAttributeValue` was a straight copy from AWS, now would be a good time to change it if we aren't completely happy with it.

currently its:
```csharp
public class MessageAttributeValue
{
    public IReadOnlyCollection<byte> BinaryValue { get; set; }

    public string StringValue { get; set; }

    public string DataType { get; set; }
}
```

JustSaying.Models/Message.cs
```diff
@@ -20,8 +19,6 @@ protected Message()
         public string Conversation { get; set; }
         public string ReceiptHandle { get; set; }
         public Uri QueueUri { get; set; }
-        public int? DelaySeconds { get; set; }
-        public IDictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
```
slang25: Perhaps continue to populate these and add an obsolete attribute, then remove? (that said, the fix is easy)

AnthonySteele: Not widely used, can be removed in a major version update as long as the fix is documented.

martincostello:
I've but WIP on this on the basis that we don't know which candidate is the winner yet between this and #449.

martincostello:
Some quick thoughts I had about this that I _think_ would be easier with envelopes, so I'm putting it here.

In some cases, attributes would be homogeneous to all messages, rather than _per_ message, right? For example, all messages of type `OrderCreated` come from `domain=order`, and there _might_ be no other attributes to add.

In these cases, we could introduce something centrally for publishing that could apply a "global" set of attributes for a message type that always get added prior to publishing onto SNS/SQS. This would then allow code to be updated at registration time to add the attributes, with the call sites remaining untouched and the extra attributes for the messages _just works_.

So a quick hacky-not-in-an-IDE example:

```csharp
config.AddMetadata<OrderCreated>("domain", "order")
      .AddMetadata<OrderCreated>("importance", "high")
      .AddMetadata<OrderCreated>("foo", "bar")
      .AddMetadata<OrderPaid>("domain", "payments");
```

This would then transparently add these three attributes to all `OrderCreated` messages that are published. Any metadata specified at publish time by the caller would overwrite and/or add to the properties:

```csharp
var envelope = new Envelope(new OrderCreated());

envelope.MessageAttributes.Add("id", "abc123");
envelope.MessageAttributes.Add("importance", "low");
envelope.MessageAttributes.Add("ordertype", "test");

await publisher.PublishAsync(envelope);
```

Which would yield a message with the attributes:

```
domain=order   // Global
importance=low // Was overridden at publish time
foo=bar        // Global
id=abc123      // Added at publish time
ordertype=test // Added at publish time
```

AnthonySteele:
I agree, some attributes will indeed be the same across a lot of messages, and it would be better to set them up at startup. I think touched on this somewhere. In a later PR maybe.

JustSaying/Messaging/PublishEnvelope.cs
```diff
@@ -15,6 +15,25 @@ public PublishEnvelope(Message message)
...
+        public IDictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
+
+        public void AddMessageAttribute(string key, MessageAttributeValue value)
+        {
+            if (MessageAttributes == null)
+            {
+                MessageAttributes = new Dictionary<string, MessageAttributeValue>();
+            }
+
+            MessageAttributes.Add(key, value);
```
martincostello: Suggest using `[]` here instead of `Add()`, as if the value got added twice you'd get an `ArgumentException` because the key was already there.

martincostello:
If we wanted to do C# attributes, I'd suggest that as an alternative that's available in-box for those so minded. As attributes can only really done done at compile-time, having them as the only way would preclude other approaches such as applying attributes from a config file, or ones that need to be different per-message.

shaynevanasperen:
Good point about the compile-time restriction! I wasn't implying that the attribute way should be the _only_ way though.

martincostello:
The consensus reached in the discussion we had over hangout this afternoon was to pursue the metadata option (#449) over envelopes.

PR #449, opened 2018-11-23
Publish metadata

AnthonySteele:

@martincostello  said:

> but I guess really would need to see a PR of approach 2 to compare and contrast

For comparison with #447. This is a variant - merge one or the other, not both.

In this fork, `PublishEnvelope` is replaced by `PublishMetadata`, which does _not_ contain the message. In the general case, both are supplied. So the signature for publish methods is in this case 

```csharp
Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
````
and the `metadata` can be null.  There are overloads/extension methods to handle simpler cases and forward them to the most general case.

This should not affect the "wire format" of messages sent.

I have no strong preference between these PRs, and both approaches work.

### CODE REVIEW

File: JustSaying/AwsTools/MessageHandling/SqsPublisher.cs [hunk 1]
```diff
@@ -26,11 +26,9 @@ public class SqsPublisher : SqsQueueByName, IMessagePublisher
             _serialisationRegister = serialisationRegister;
         }
 
-        public Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
-        public async Task PublishAsync(Message message, CancellationToken cancellationToken)
+        public async Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
```
martincostello: Some overloads as extensions for _with just a token_, _with just metadata_ and _just a token_ would make this less source-breaking to consume.

JustSaying/JustSayingBus.cs
```diff
@@ -130,12 +130,10 @@ public void Start(CancellationToken cancellationToken = default)
             }
         }
 
-        public Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
-        public async Task PublishAsync(Message message, CancellationToken cancellationToken)
+        public async Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
```
martincostello: Same regarding overloads as extensions.

JustSaying/JustSayingBus.cs
```diff
@@ -175,14 +173,20 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
             return publishersByTopic[topic];
         }
 
-        private async Task PublishAsync(IMessagePublisher publisher, Message message, int attemptCount = 0, CancellationToken cancellationToken = default)
+        private async Task PublishAsync(
+            IMessagePublisher publisher,
+            Message message,
+            PublishMetadata metadata,
+            int attemptCount = 0,
```
martincostello: Do any of the methods in this class ever call this method without specifying `attemptCount` or `cancellationToken`? If not, can make them non-optional so they don't accidentally get forgotten.

JustSaying/JustSayingFluently.cs
```diff
@@ -158,25 +159,19 @@ public void StartListening(CancellationToken cancellationToken = default)
...
-        /// <param name="message"></param>
-        public virtual Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
-
         /// <summary>
         /// Publish a message to the stack, asynchronously.
         /// </summary>
         /// <param name="message"></param>
         /// <param name="cancellationToken"></param>
-        public virtual async Task PublishAsync(Message message, CancellationToken cancellationToken)
+        public virtual async Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
```
martincostello: Extension overloads again.

JustSaying/Messaging/IMessagePublisher.cs
```diff
@@ -1,13 +1,11 @@
...
 using JustSaying.Models;
 
 namespace JustSaying.Messaging
 {
     public interface IMessagePublisher
     {
-        Task PublishAsync(Message message);
-
-        Task PublishAsync(Message message, CancellationToken cancellationToken);
+        Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken);
```
martincostello: Same. I imagine one set would cover all the cases where I'm making this comment.

JustSaying/Messaging/MessagePublisherExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class MessagePublisherExtensions
+    {
+        public static Task PublishAsync(this IMessagePublisher publisher, Message message)
+        {
+            return publisher.PublishAsync(message, CancellationToken.None);
+        }
+
+        public static async Task PublishAsync(this IMessagePublisher publisher,
```
martincostello: One for `(Message, CancellationToken)` too?

JustSaying/Messaging/PublishMetadata.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public TimeSpan? Delay { get; set; }
+
+        public IDictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
+
+        public void AddMessageAttribute(string key, MessageAttributeValue value)
+        {
+            if (MessageAttributes == null)
+            {
+                MessageAttributes = new Dictionary<string, MessageAttributeValue>();
```
martincostello: An appropriate `StringComparer` should be passed to the dictionary constructor to make the case sensitivity explicit.

JustSaying/Messaging/PublishMetadata.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace JustSaying.Messaging
+{
+    public class PublishMetadata
+    {
+        public TimeSpan? Delay { get; set; }
+
+        public IDictionary<string, MessageAttributeValue> MessageAttributes { get; set; }
+
+        public void AddMessageAttribute(string key, MessageAttributeValue value)
```
martincostello: Would be nice if this (and the other) method returned `PublishMetadata` so you can chain fluently in your setup.

```csharp
var metadata = new AddMessageAttribute()
    .AddMessageAttribute("foo", "bar")
    .AddMessageAttribute("qux", "zab");
```

martincostello:
I've but WIP on this on the basis that we don't know which candidate is the winner yet between this and #447.

JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
AnthonySteele: Those are present in `MessagePublisherExtensions`

JustSaying/Messaging/MessagePublisherExtensions.cs
AnthonySteele: It's there, line 17-27

martincostello: It was at the bottom of the PR when I was reviewing it, hence I didn't see as I was commenting.

JustSaying/Messaging/MessagePublisherExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using JustSaying.Models;
+
+namespace JustSaying.Messaging
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class MessagePublisherExtensions
+    {
+        public static Task PublishAsync(this IMessagePublisher publisher, Message message)
```
martincostello: Should have an overload for `(Message, CancellationToken)` too.

JustSaying/Messaging/PublishMetadata.cs
```diff
@@ -29,5 +30,15 @@ public PublishMetadata AddMessageAttribute(string key, string value)
                 DataType = "String"
             });
         }
+
+        public PublishMetadata AddMessageAttribute(string key, double value)
+        {
+            return AddMessageAttribute(key, new MessageAttributeValue
```
martincostello: 👍

AnthonySteele:
Right now I prefer #449 over #447, on the basis that

* Design-wise, not all of this data needs to be a "wrapper" for the message.  the `MessageAttribute` are are sent with the message, but the `Delay` is not, it is instructions on how to publish. It doesn't need to enclose the message.

* Practicality wise, moving from `PublishAsync(m, ctx);` to the new equivalent `PublishAsync(m, null, ctx);` seems more familiar than `PublishAsync(new PublishEnvelope(m), ctx);`

AnthonySteele:
Things that are out of scope of this PR, but can be added in a non-breaking way later:

* A global `IDictionary<string, MessageAttributeValue> MessageAttributes` or default `PublishMetadata` that is configured once at startup, and defaulted onto every message sent. Might be useful for setting common properties?
* Any global `Func<PublishMetadata,PublishMetadata>` or `Action<PublishMetadata>` or similar for manipulating requests in an arbitrary manner.  We don't have a candidate use for these and for existing simple things, just supplying the value is much simpler.
* Any `Func<SendMessageRequest, SendMessageRequest>` or `Action<SendMessageRequest>` etc on the `PublishMetadata`.  Again, possible in a non-breaking way, but we don't have a candidate use for these.

martincostello:
For item 1, that's something we could probably add to the new builder stuff once #431 gets to master.

PR #450, opened 2018-11-24
Remove AsyncLazy<T>

martincostello:
* Remove `AsyncLazy<T>`.
  * Remove unused using statement.


JustSaying/AwsTools/AsyncLazy.cs
```diff
@@ -1,10 +1,11 @@
 using System;
-using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 
 namespace JustSaying.AwsTools
 {
-    public class AsyncLazy<T> : Lazy<Task<T>>
+    // TODO Does this still meet async best-practice?
```
martincostello: The usage of `Task.Factory.StartNew()` prompted this question.

slang25: Good shout 👍 

This code originates from a time where `Task.Run` didn't exist ([see here](https://blogs.msdn.microsoft.com/pfxteam/2011/01/15/asynclazyt/)), it's correct in this implementation, but would be easier to reason with by using `Task.Run` here.

However as pointed out in the post where this code comes from, we could chose to not start a new task if we don't need the initial synchronous work in the delegate to be started on a different thread, and looking at the usage we don't need that, so I created a PR:
https://github.com/martincostello/JustSaying/pull/1

AnthonySteele:
I think that this class basically does nothing now, it looks like an alias for `Lazy<Task<T>>`

slang25:
Yeah, you could very well inline it, it's only used in one place

slang25:
https://github.com/martincostello/JustSaying/pull/2

PR #454, opened 2018-11-24
Improve MessageSerialisationRegister

martincostello:
* Add a null check for the `IMessageSubjectProvider`.
  * Use `TryGetValue()` on register to prevent conflicts/race-conditions.
  * Throw nicer exception if serialization fails due to a message type not being registered.
  * Use correct `StringComparison` (see #401).
  * Use proper type instead of `var`.
  * Add some `TODO`s about logging.

### DIFF
--- a/JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
+++ b/JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
@@ -28,23 +28,25 @@ namespace JustSaying.Messaging.MessageSerialisation
 
         public Message DeserializeMessage(string body)
         {
-            foreach (var formatter in _map)
+            foreach (var pair in _map)
             {
-                string messageSubject = formatter.Value.Serialiser.GetMessageSubject(body);
+                TypeSerialiser typeSerializer = pair.Value;
+                string messageSubject = typeSerializer.Serialiser.GetMessageSubject(body);
 
                 if (string.IsNullOrWhiteSpace(messageSubject))
                 {
                     continue;
                 }
 
-                Type matchedType = formatter.Value.Type;
+                Type matchedType = typeSerializer.Type;
 
                 if (!string.Equals(_messageSubjectProvider.GetSubjectForType(matchedType), messageSubject, StringComparison.OrdinalIgnoreCase))
                 {
                     continue;
                 }
 
-                return formatter.Value.Serialiser.Deserialise(body, matchedType);
+                IMessageSerialiser messageSerializer = typeSerializer.Serialiser;
+                return messageSerializer.Deserialise(body, matchedType);
             }
 
             // TODO Maybe we should log the body separately (at debug/trace?), rather than include it in the exception message. Then they're easier to filter.
@@ -56,13 +58,14 @@ namespace JustSaying.Messaging.MessageSerialisation
         {
             var messageType = message.GetType();
 
-            if (!_map.TryGetValue(messageType, out TypeSerialiser serializer))
+            if (!_map.TryGetValue(messageType, out TypeSerialiser typeSerializer))
             {
                 // TODO Log out what *is* registered at debug?
                 throw new MessageFormatNotSupportedException($"Failed to serialize message of type {messageType} because it is not registered for serialization.");
             }
 
-            return serializer.Serialiser.Serialise(message, serializeForSnsPublishing, _messageSubjectProvider.GetSubjectForType(messageType));
+            IMessageSerialiser messageSerializer = typeSerializer.Serialiser;
+            return messageSerializer.Serialise(message, serializeForSnsPublishing, _messageSubjectProvider.GetSubjectForType(messageType));
         }
     }
 }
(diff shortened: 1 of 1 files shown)


JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
```diff
@@ -1,57 +1,68 @@
...
-            return formatter.Serialiser.Serialise(message, serializeForSnsPublishing, _messageSubjectProvider.GetSubjectForType(messageType));
-        }
 
+            if (!_map.TryGetValue(messageType, out TypeSerialiser serializer))
+            {
+                // TODO Log out what *is* registered at debug?
+                throw new MessageFormatNotSupportedException($"Failed to serialize message of type {messageType} because it is not registered for serialization.");
+            }
+
+            return serializer.Serialiser.Serialise(message, serializeForSnsPublishing, _messageSubjectProvider.GetSubjectForType(messageType));
```
martincostello: This naming could probably be improved...

AnthonySteele:
Only thing that I'm not sure about is the avoidance of `var`. Var is fine and is local style to use it in cases where the declared type would be exactly the same as the RSH expression. The main time that this is not true is when you want the variable to be of a base or interface type and the RHS resolves to a subclass /interface implementation type.

JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
```diff
@@ -1,57 +1,71 @@
...
         }
 
         public void AddSerialiser<T>(IMessageSerialiser serialiser) where T : Message
         {
             var key = typeof(T);
-            if (!_map.ContainsKey(key))
+            if (!_map.TryGetValue(key, out TypeSerialiser serializer))
             {
-                _map.Add(key, new TypeSerialiser(typeof(T), serialiser));
+                _map[key] = new TypeSerialiser(typeof(T), serialiser);
```
AnthonySteele: Since _map_ is private, there is no reason not to type it as `ConcurrentDictionary<>`  which would allow you to use the atomic methods `TryAdd` / `GetOrAdd` / `AddOrUpdate ` here.

As it stands, there's not much difference between this and the original code in terms of concurrency.

martincostello: `ConcurrentDictionary<K, V>` conforms to `IDictionary<K, V>` under the hood when used as such, so you can use it "simpler" while not getting the possibility of a `NullReferenceException` if two adds collide and corrupt the internal state, which is what happens to `Dictionary<K, V>`.

martincostello:
I changed `var` because I had to actually go delving to find what the types actually were, it wasn't intuitively obvious just from reading the code with no IDE assistance and scrolling up the file.

Also, I prefer to _not_ use it for the keyword types (`bool`, `string`, `int` etc.). I use an EditorConfig rule to hint against using `var` in those cases in other projects.

AnthonySteele:
¯\_(ツ)_/¯
_De gustibus non est disputandum_

JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
AnthonySteele: So `ConcurrentDictionary`'s implementation of `_map[key] = value;` is resistant to concurrency issues that you can get with `Dictionary` ? Cool, I did not know that.

JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
```diff
@@ -1,57 +1,71 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using JustSaying.Models;
 
 namespace JustSaying.Messaging.MessageSerialisation
 {
     public class MessageSerialisationRegister : IMessageSerialisationRegister
     {
+        // TODO Inconsistent use of American and British English for "serialize"/"serialise"
```
slang25: I think we should standardise (or is it standardize 😉) on the american spelling, we're fighting with the ecosystem otherwise

JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
slang25: This has been one of those questions I've wondered, but never taken the time to read the docs on. TIL 👍

JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
martincostello: Personally I go with the _zees_ to match tooling defaults, but I'm happy to go with the _esses_ if that's the majority opinion. We just need to be consistent.

slang25: Same feelings here, if the APIs are getting a rehaul then count me as a vote for the *zees*, if not then these methods should match.

martincostello:
I'll do the spelling standardization/standardisation separately.

JustSaying/Messaging/MessageSerialisation/MessageSerialisationRegister.cs
shaynevanasperen: I vote for _zees_ for the reasons @slang25 mentioned.

martincostello: #457

PR #456, opened 2018-11-25
Improve extensibility for AWS clients

martincostello:
Port two commits from #431 that improve extensibility:
  1. Allow the SQS and SNS clients to be more easily configured by allowing extension points in `DefaultAwsClientFactory`.
  1. Add a constructor to `AwsClientFactoryProxy` that supports re-using the same single factory via `Lazy<T>`.

### DIFF
--- a/JustSaying/AwsTools/AwsClientFactoryProxy.cs
+++ b/JustSaying/AwsTools/AwsClientFactoryProxy.cs
@@ -16,6 +16,11 @@ namespace JustSaying.AwsTools
             _awsClientFactoryFunc = awsClientFactoryFunc;
         }
 
+        public AwsClientFactoryProxy(Lazy<IAwsClientFactory> factory)
+        {
+            _awsClientFactoryFunc = () => factory.Value;
+        }
+
         public IAwsClientFactory GetAwsClientFactory() => _awsClientFactoryFunc();
 
         public void SetAwsClientFactory(Func<IAwsClientFactory> func) => _awsClientFactoryFunc = func;
--- a/JustSaying/AwsTools/DefaultAwsClientFactory.cs
+++ b/JustSaying/AwsTools/DefaultAwsClientFactory.cs
@@ -1,4 +1,5 @@
-﻿using Amazon;
+using System;
+using Amazon;
 using Amazon.Runtime;
 using Amazon.SimpleNotificationService;
 using Amazon.SQS;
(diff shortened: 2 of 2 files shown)


JustSaying/AwsTools/AwsClientFactoryProxy.cs
```diff
@@ -16,8 +16,13 @@ public AwsClientFactoryProxy(Func<IAwsClientFactory> awsClientFactoryFunc)
             _awsClientFactoryFunc = awsClientFactoryFunc;
         }
 
+        public AwsClientFactoryProxy(Lazy<IAwsClientFactory> factory)
```
slang25: `() => factory.Value` is just a `Func<IAwsClientFactory`, will the other overload not suffice?

slang25: Instead of:
```csharp
new AwsClientFactoryProxy(new Lazy<IAwsClientFactory>(ClientFactoryBuilder.Build))
```
you'd do:
```csharp
var lazyFactory = new Lazy<IAwsClientFactory>(ClientFactoryBuilder.Build);
new AwsClientFactoryProxy(() => lazyFactory.Value)
```
I can see how it would be easy to new it up inside the delegate (which would be wrong), or eagerly get the `.Value` here though. It just struck me as odd, but I see how it moves people towards "the pit of success".

martincostello: I can back that commit out, but I thought it was neater than doing it at the call site. I'm hoping if #431 goes somewhere we can get rid of `(I)AwsClientFactoryProxy` anyway.

AnthonySteele:
Is the `ServiceUri` so that you can more easily plug in the mock / local version for tests?

martincostello:
Exactly - otherwise you have to create your own factory _just_ to change that one thing.

JustSaying/AwsTools/AwsClientFactoryProxy.cs
slang25: I think you should keep it as it solved a problem for you. Just throwing out an alternative:
```csharp
public AwsClientFactoryProxy(Func<IAwsClientFactory> factory, bool lazy = false)
{
    _awsClientFactoryFunc = lazy ? new Lazy(factory) : factory;
}
```

PR #461, opened 2018-11-27
Remove IMessageLock

### LINKED ISSUE #458: Remove IMessageLock
In JustSaying we have two interfaces, `IMessageLock` and `IMessageLockAsync`.

We provide an implementation of `IMessageLockAsync` which just delegates to an inner `IMessageLock`, but we don't provide a built-in implementation of `IMessageLock` at all.

To simplify things as everything's async now anyway, we should just provide `IMessageLockAsync` (and drop the `Async` suffix?), and leave any "it's not really async" implementation to anyone who wants to use the interface as JustSaying consumes it all in an async way anyway.

martincostello:
Remove the `IMessageLock` interface, just leaving `IMessageLockAsync`.

Resolves #458.


JustSaying.IntegrationTests/WhenRegisteringASqsSubscriber/MessageLockStore.cs
```diff
@@ -17,13 +18,18 @@ public MessageLockResponse TryAquireLockPermanently(string key)
                 _store.Add(key, 1);
             }
 
-            return new MessageLockResponse { DoIHaveExclusiveLock = canAquire };
+            var response = new MessageLockResponse { DoIHaveExclusiveLock = canAquire };
```
shaynevanasperen: Shouldn't this be setting the expiry properties? And why have we got two methods that do that same thing but are named differently? And if we are going to set expiry properties, then we should implement the expiry!

martincostello: No idea. I've just moved this code to conform to the interface change. And it's a test object.

shaynevanasperen: Ah, didn't see that it's a test fake. Nevermind then.

PR #465, opened 2018-12-03
Adopt MinVer

slang25:
This PR swaps out our (IMO good) versioning in favour of [MinVer](https://github.com/adamralph/minver).
After reading the [How it works](https://github.com/adamralph/minver#how-it-works) section, I'm convinced it's nailed it, and does everything we currently want in terms of versioning (even the`x.0.0.0` assembly version number).

I've used the `MinVerMajorMinor` property to mimic how we currently control the version number, but we could be worth dropping this and letting it discover it automatically. It would mean after releasing a `v7.0.0`, we'd create a `v8.0.0-alpha.0` tag, then any PR builds would follow on from that (if that makes sense). This would help us avoid situations where a release is made without a tag, as that would be the only way to manage the version number.


slang25:
We should also re-evaluate [this script](https://github.com/justeat/JustSaying/blob/eb92af9c63281cf1d7cfc165e29d46d28243940d/release.ps1)

AnthonySteele:
You say "re-evaluate" I say "delete" with "fire".

slang25:
Exactly!

Releases.md
```diff
@@ -0,0 +1,11 @@
+# Process for Releasing
+
+To create a new release, either push a tag via git, or go to the GitHub [releases tab](https://github.com/justeat/JustSaying/releases) and click `Draft a new release`.
```
martincostello: I think the UI is the only method we should use. I seem to remember that someone made a tag in a repo somewhere recently, and for whatever reason it didn't trigger all the things that make a release work.

slang25: I know that I have got it wrong a few times, I only just learned you could do it through the UI! I'm a big fan of that.

AnthonySteele: That would have been me not getting it to work property.

Releases.md
```diff
@@ -0,0 +1,13 @@
+# Process for Releasing
+
+First before creating a release, make sure that [the version number in `Directory.Build.Props`](https://github.com/justeat/JustSaying/blob/d30543fbfc3cf640835339efbe497466e230f220/Directory.Build.props#L22) matches the major and minor version you are looking to release.
```
slang25: We couldn't need to do this is we opt to rely on tags only. If we did that then there's no risk of them not being aligned.

martincostello:
Have you got this configured correctly? I just tried adding MinVer to a package of my own and it was generating "too many numbers" and the warning caused my build to fail.

Looking in the [AppVeyor logs](https://ci.appveyor.com/project/justeattech/justsaying/builds/20756819#L83) the same is happening here, but it still passes because warnings aren't errors in JustSaying:

```
C:\projects\justsaying\.dotnetcli\sdk\2.1.500\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(202,5): error NU5105: The package version '7.0.0-alpha.0.49+build.455' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients. 
```

Specifically I think it's the `alpha.0.49` part that's wrong, because it should be `alpha.49`?

slang25:
Looks like this is relevant https://github.com/adamralph/minver/issues/92

slang25:
We can change this behaviour, but it is expected. The `alpha.0` is the controlled part that will be inferred from tags. Do you think we should drop the number part?

adamralph:
> 
> 
> Have you got this configured correctly? I just tried adding MinVer to a package of my own and it was generating "too many numbers" and the warning caused my build to fail.
> 
> Looking in the [AppVeyor logs](https://ci.appveyor.com/project/justeattech/justsaying/builds/20756819#L83) the same is happening here, but it still passes because warnings aren't errors in JustSaying:
> 
> ```
> C:\projects\justsaying\.dotnetcli\sdk\2.1.500\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(202,5): error NU5105: The package version '7.0.0-alpha.0.49+build.455' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients. 
> ```

@martincostello You _will_ get NU1505 when using MinVer, because it _will_ produce SemVer 2.0 version numbers. The reason the build is failing is that you have configured your project to treat warnings as errors. As the warning says, you can safely ignore it if the package is not intended for older clients. "Older clients" in this case, means [anything older than around September 2017](https://github.com/adamralph/minver/issues/92#issuecomment-437839733). You can get rid of the warning like so: https://github.com/adamralph/bullseye/blob/794fbaee349ab4478921f4d5d51b2faaf28158fc/Bullseye/Bullseye.csproj#L8. If you _do_ care about those older (ancient?) clients, then it's probably best not to use MinVer.

> Specifically I think it's the `alpha.0.49` part that's wrong, because it should be `alpha.49`?

@martincostello the version is `alpha.0.49` because the current commit is 49 commits ahead of the last tag or root commit. The 49 is the _height_. This is explained in https://github.com/adamralph/minver#how-it-works

> Lets see what the outcome of this is: [adamralph/minver#169](https://github.com/adamralph/minver/pull/169)

@slang25 that PR will not remove NU1505. You will still get it when you build any commit which has non-zero height above the root commit or the last tag.

The reason the default pre-release identifiers are `alpha.0` is to guarantee that they are lower than the first tag you will create, which at it the lowest, will be `alpha.1`. In that way, your commits will have the following versions:

- `7.0.0-alpha.0.49` (49 commits above the last tag or root commit)
- `7.0.0-alpha.0.50`
- `7.0.0-alpha.0.51`
- ...
- `7.0.0-alpha.1` (tagged)
- `7.0.0-alpha.1.1`
- `7.0.0-alpha.1.2`
- ...
- `7.0.0-alpha.2` (tagged)
- `7.0.0-alpha.2.1`
- `7.0.0-alpha.2.2`
- ...
- `7.0.0-beta.1` (tagged)
- `7.0.0-beta.1.1`
- `7.0.0-beta.1.2`
- ...
- `7.0.0` (tagged)
- `7.0.0-alpha.0.1` (default pre-release identifiers)
- `7.0.0-alpha.0.2`
- ...and so on

Unless you are using a pre-release scheme where `alpha` doesn't make sense, I strongly recommend that you stick to the default `alpha.0` pre-release identifiers built into MinVer, in which case [adamralph/minver#169](https://github.com/adamralph/minver/pull/169) will have no value to you.

slang25:
Thanks @adamralph, I was aware it wouldn't get rid of NU1505, like you say that's a tradeoff between SemVer 2.0 and ye olde clients. We are used to releasing just `-alpha<some incrementing number>` or `-beta<some incrementing number>` so wanted to be able to control the prerelease identifier (as we didn't need an additional controlled number), but if there is good reasoning to adopt the default in MinVer then I'm up for that.

adamralph:
@slang25 even if you change the default pre-release identifiers, MinVer will still add the height to any commit which is not tagged. That behaviour cannot be changed. When you tag, you can tag with whatever version number you like. `7.0.0-alpha<some incrementing number>` will work just fine. [MinVer doesn't care what your pre-release versioning scheme is](https://github.com/adamralph/minver#can-i-use-my-own-pre-release-versioning-scheme). In that case, your versions will look something like this:

- `7.0.0-alpha.0.49` (49 commits above the last tag or root commit)
- `7.0.0-alpha.0.50`
- `7.0.0-alpha.0.51`
- ...
- `7.0.0-alpha001` (tagged)
- `7.0.0-alpha001.1`
- `7.0.0-alpha001.2`
- ...
- `7.0.0-alpha002` (tagged)
- `7.0.0-alpha002.1`
- `7.0.0-alpha002.2`
- ...
- `7.0.0-beta001` (tagged)
- `7.0.0-beta001.1`
- `7.0.0-beta001.2`
- ...
- `7.0.0` (tagged)
- `7.0.0-alpha.0.1` (default pre-release identifiers)
- `7.0.0-alpha.0.2`
- ...and so on

The assumption with the commits between the tags, is that they will not be released (or perhaps just to a specialised CI NuGet feed, e.g. on MyGet). If you don't care about them, then you also don't care about the `alpha.0` identifiers. That said, if you _do_ want to release the CI builds to a special NuGet feed, I'm not sure what the order of precedence will between `alpha.0` and `alpha001`. If `alpha.0` comes first, then you're fine, but if the other way round, then you do indeed need to change the default pre-release identifiers (to `alpha000` in the example above).

Alternatively, as you suggest above, you could immediately tag the commit following `7.0.0` with `7.0.0-alpha000` to "prime" the pre-release identifiers.

adamralph:
Although having said all of the above, I recommend just switching to SemVer 2.0. 🙂  I'm not sure it's worrying about those older clients anymore.

slang25:
Just to be clear, I was thinking this:

- `7.0.0-alpha.49` (49 commits above the last tag or root commit)
- `7.0.0-alpha.50`
- `7.0.0-alpha.51`
- ...
- `7.0.0-beta` (tagged)
- `7.0.0-beta.1`
- `7.0.0-beta.2`
- ...and so on

I'm thinking now we should just adopt the default.

adamralph:
@slang25 that proposal would work, so long as you don't mind either:

- releasing every commit, or
- having gaps in your alpha/beta/etc. release numbers

Directory.build.targets
```diff
@@ -0,0 +1,16 @@
+<Project>
+    <Target Name="AppVeyorPR" AfterTargets="MinVer" Condition="'$(APPVEYOR_PULL_REQUEST_NUMBER)' != ''" >
+        <PropertyGroup>
+            <PackageVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch)-PR.$(APPVEYOR_PULL_REQUEST_NUMBER)</PackageVersion>
```
martincostello: I _think_ I'd prefer `pr` to `PR`.

martincostello:
The only remaining concern I have about this is whether the SemVer 2.0 tags would create issues with some of our older components that are using older .NET tooling, which may prevent them from being able to consume the 7.0.0 release.

slang25:
It's a valid concern. 

I think there may already be a requirement for this library to be consumed by nuget 4.3 and up because it has a `netstandard2.0` target (not sure if it's required only if you are needing the target).

I'd be interested to know if there are any builds you are seeing using a version of nuget lower than 4.3 and that could could currently update to the current JustSaying 7.x

martincostello:
Without doing a full audit of everyone's TeamCity builds, I have no idea what proportion that would be. But my team owns three components that are still on JustSaying _4_.

Attempts were made to update them recently but the .NET Standard issues with .NET Framework 4.x have made that...fraught.

slang25:
Cool, I'll try to verify client compatibility

slang25:
It appears to be fine for `7.0.0-alpha.0.66`, and we'd only be pushing `7.0.0-alpha.0`  to nuget.org, which again is fine.

It's only a problem if you wanted to experiment with a non-released build with your CI nuget client where you couldn't update the client.

adamralph:
FWIW, to clarify a couple of things:

- You wouldn't push `alpha.0` to NuGet. The first tag and release would be `alpha.1`. The interim builds, which have the height attached, are designed to have an earlier version than the next release. That is why they are `alpha.0.{height}` rather than `alpha.1.{height}`
- The versions with and without height have the same client requirements. It's the dot after the alpha that makes it SemVer 2.0 rather than the two numbers.

slang25:
Thanks for the clarification @adamralph

Releases.md
adamralph: This property doesn't affect the release process. The release is purely driven by the tag you create. Updating this property is actually something you'd do _after_ releasing, in the case where you want to start work on a new major-minor range.

slang25: Thanks, I'll update this 👍

Releases.md
```diff
@@ -0,0 +1,43 @@
+## About Versioning
+
+JustSaying adopts [SemVer 2.0](https://semver.org/spec/v2.0.0.html), and uses [MinVer](https://github.com/adamralph/minver) to achieve this. [See here](https://github.com/adamralph/minver#how-it-works) for more information on how it works.
+
+## Process for Releasing
+
+First before creating a release, make sure that [the version number in `Directory.Build.Props`](https://github.com/justeat/JustSaying/blob/d30543fbfc3cf640835339efbe497466e230f220/Directory.Build.props#L22) matches the major and minor version you are looking to release.
```
adamralph:
```suggestion
```

This paragraph can be removed completely. There is no reason to change this version number before releasing.

Directory.Build.props
```diff
@@ -12,13 +11,16 @@
     <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.NetCore.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
     <PackageReference Include="Text.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
+    <PackageReference Include="MinVer" Version="1.0.0-beta.2" PrivateAssets="All" />
```
adamralph: This can be `1.0.0-beta.4` now.

Releases.md
```diff
@@ -0,0 +1,43 @@
...
+
+`v{MAJOR}.{MINOR}.{PATCH}`
+
+For example `v6.0.1`.
+
+Once the AppVeyor build has completed, you can deploy to nuget via the AppVeyor UI.
+
+## Example Release Pattern
+
+Here is what the versioning should typically look like, including what and when to tag a realease as outlined above.
```
adamralph:
```suggestion
Here is what the versioning should typically look like, including what and when to tag a release as outlined above.
```

Releases.md
```diff
@@ -0,0 +1,43 @@
...
+- 7.0.0-alpha.2 (tagged)
+- 7.0.0-alpha.2.1
+- 7.0.0-alpha.2.2
+- ...
+- 7.0.0-beta.1 (tagged)
+- 7.0.0-beta.1.1
+- 7.0.0-beta.1.2
+- ...
+- 7.0.0 (tagged)
+- 7.0.0-alpha.0.1 (default pre-release identifiers)
```
adamralph:
```suggestion
- 7.0.1-alpha.0.1 (default pre-release identifiers)
```

Releases.md
```diff
@@ -0,0 +1,43 @@
...
+- 7.0.0-alpha.0.2
```
adamralph:
```suggestion
- 7.0.1-alpha.0.2
```

slang25:
Thanks @adamralph. I've applied your suggestions, and am happy with the state of this PR.

@AnthonySteele @martincostello I find it hard to gauge in GitHub comments the consensus and feelings towards things like this, feel free to let me know if this is something you do or don't want to adopt 🙂

martincostello:
If I'm honest, I'm currently ambivalent to this either way, though that's mainly because when I first tried it I couldn't get it to work the way I wanted.

We have our weekly JustSaying meeting in 20 minutes so we can discuss there, then feedback here afterwards.

slang25:
Cool, say hello to everyone from me

adamralph:
> ...when I first tried it I couldn't get it to work the way I wanted.

@martincostello out of interest, what problems did you face?

martincostello:
@adamralph I think it was because I was getting numbers like `alpha.0.x` where I didn't want the zero, but couldn't seem to suppress it. Admittedly this was on a repo that at the time was brand new and had no tags, but I think because I couldn't work out to remove it I just stopped looking at it.

adamralph:
@martincostello the zero effectively means "there isn't yet an alpha". When you want to release your first alpha, you'll tag a commit with something like "1.0.0-alpha.1", indicating "alpha 1".

Directory.Build.props
```diff
@@ -12,13 +11,16 @@
     <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.NetCore.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
     <PackageReference Include="Text.Analyzers" Version="$(AnalyzersPackageVersion)" PrivateAssets="All" />
+    <PackageReference Include="MinVer" Version="1.0.0-beta.4" PrivateAssets="All" />
```
adamralph: This can now be 1.0.0-rc.1

brainmurphy:
@slang25 Have merged 👍 Can you submit a PR to update MinVer to latest?

PR #469, opened 2018-12-11
Remove IsTestProject from Directory.Build.props

slang25:
We already were using `IsTestProject` so I've removed the duplicate definitions following #468 (I prefer the less magic approach)


slang25:
Confirmed that `EmbedAllSources` is still taking effect even though the properties are defined at a project level and it is at directory level.

PR #470, opened 2018-12-11
Remove redundant IsTestProject

slang25:
`IsTestProject` is always defined for projects referencing `Microsoft.NET.Test.Sdk`
 `15.9.0` (just learnt with thanks to @shep1987). https://github.com/Microsoft/vstest/blob/76fd0f5b09ac3c4c3db0e283deb01b43f7f8a6ab/src/package/nuspec/Microsoft.NET.Test.Sdk.props#L24


JosephWoodward:
@slang25 Oh interesting, that's good to know.

PR #479, opened 2018-12-13
Add NuGet package for Microsoft DI

### LINKED ISSUE #474: NuGet package to bridge to Microsoft .NET Core DI
As described in #473, implement a bridge for DI between Microsoft's .NET Core implementation and JustSaying.

This was initially merged as part of #431, but needs extracting out into its own NuGet package.

I'll work on this one as I did the initial plumbing in the prototype.

martincostello:
Add a new NuGet package to provide glue between JustSaying and Microsoft.Extensions.DependencyInjection.

This PR essentially just moves some code already added by #431 into its own NuGet package.

Resolves #474.

### DIFF
--- a/Build.ps1
+++ b/Build.ps1
@@ -14,7 +14,8 @@ $sdkFile = Join-Path $solutionPath "global.json"
 
 $libraryProjects = @(
     (Join-Path $solutionPath "JustSaying\JustSaying.csproj"),
-    (Join-Path $solutionPath "JustSaying.Models\JustSaying.Models.csproj")
+    (Join-Path $solutionPath "JustSaying.Models\JustSaying.Models.csproj"),
+    (Join-Path $solutionPath "JustSaying.Extensions.DependencyInjection.Microsoft\JustSaying.Extensions.DependencyInjection.Microsoft.csproj")
 )
 
 $testProjects = @(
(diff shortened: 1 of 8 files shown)


slang25:
Very nice. Naming wise, I see a lot of other packages with just `<blah>.Extensions.DependencyInjection`, or if they include `Microsoft` it comes before the DI part. What's your thoughts on dropping `Microsoft`?

tonymobster:
Can this package be published? Even as a pre-release? I really like the new builder API and want to use that on a new POC i'm about to start.

martincostello:
It’s a holiday here in the U.K. where the team are based until Tuesday, but we could discuss shipping an alpha to NuGet next week if people like yourself are interested in early access to parts of v7.0.0.

tonymobster:
Sweet! That would be great. We're on holiday here until Tuesday too so next week is fine.

tonymobster:
Has there been a decision on this?

martincostello:
Instead of us publishing an alpha to NuGet.org, you can configure a custom feed with the NuGet in our AppVeyor CI instead: `https://ci.appveyor.com/nuget/justsaying-jhto5gatphuv`

Hope that works OK for you.

tonymobster:
Sure, that works for me. Thanks

PR #480, opened 2018-12-13
[WIP] Message context

AnthonySteele:
A start of an approach to the "message Envelopes" feature.
Do not merge, just please say if you think this is worth completing, or if it is misguided or you can see a better way.

This approach  broadens the handler contract again into  
`Task<bool> HandleAsync(T message, MessageContext context)`.
We have expanded the contact before for async, moving from `bool` to `Task<bool>` return, with a similar wrapper.

The now idea is that
* Simple consumers would use the existing `IHandlerAsync.Handle<T>` as before and work with the message deserialised as `T`.
* Consumers that need all the metadata would use `IHandlerWithMessageContext` instead, and get the additional `messageContext` object containing various other properties such as the queue url and the raw SQS message. 
* The queue url and anything that can be read from the raw SQS message would be deleted from the `Message` base class. if you need them, that's not a "simple consumer", that's a "consumer that needs the metadata".
* internal to  JustSaying, only the `IHandlerWithMessageContext` need be dealt with, as `IHandlerAsync` instances can be wrapped in the `HandlerAdapter` and appear as `IHandlerWithMessageContext`.
* extension methods over `MessageContext` could make it easy to read e.g. specific named metadata on the raw SQS Message. 
* Longer term, the simpler handler and wrapper might be deprecated and fall away, as with the synchronous handler.

The additional `MessageContext` object is a parallel to the `PublishMetadata` in #449. 

"Envelope" is also a possible design here, but has the additional implementation complexity that as the Envelope contains a generic message of some type `T: Message`, the envelope is also generic over `T: Message`; whereas the `MessageContext` is not a generic at all.


JustSaying/Messaging/MessageHandling/IHandlerWithContext.cs
```diff
@@ -0,0 +1,25 @@
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.MessageHandling
+{
+    public interface IHandlerWithContext<in T>
+    {
+        Task<bool> Handle(T message, MessageContext context);
```
martincostello: `HandleAsync()`? (as it's new)

JustSaying/Messaging/MessageHandling/IHandlerWithContext.cs
```diff
@@ -0,0 +1,25 @@
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.MessageHandling
+{
+    public interface IHandlerWithContext<in T>
```
martincostello: `IMessageHandler`, `IHandlerWithMessageContext`? `WithContext` seems a bit vague (though so does my first suggested alternative...).

JustSaying/Messaging/MessageHandling/IHandlerWithContext.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class HandlerAdapter<T> : IHandlerWithContext<T>
+    {
+        private readonly IHandlerAsync<T> _inner;
+
+        public HandlerAdapter(IHandlerAsync<T> inner)
+        {
+            _inner = inner;
+        }
+
+        public Task<bool> Handle(T message, MessageContext context)
```
martincostello: Referring back to David Fowler's recommendations, should this be `async` rather than just return the `Task`?

JustSaying/Messaging/MessageHandling/IHandlerWithContext.cs
```diff
@@ -0,0 +1,26 @@
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.MessageHandling
+{
+    public interface IHandlerWithContext<in T>
+    {
+        Task<bool> HandleAsync(T message, MessageContext context);
+    }
+
+    internal sealed class HandlerAdapter<T> : IHandlerWithContext<T>
```
martincostello: When/if merged, this class should be in its own file.

JustSaying/Messaging/MessageHandling/MessageContext.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class MessageContext
+    {
+        public MessageContext(SQSMessage message, Uri queueUri)
+        {
+            Message = message ?? throw new ArgumentNullException(nameof(message));
+            QueueUri = queueUri ?? throw new ArgumentNullException(nameof(queueUri));
+        }
+
+        public SQSMessage Message { get; }
```
martincostello: Does there need to be an SNS equivalent for this, or are they the same thing in this context?

AnthonySteele: That's not needed, we only read from SQS queues. 

The distinction exists only on publishing where you can publish directly to a SQS queue, or more usually, indirectly by a SNS endpoint that forwards to all attached queues,.

martincostello:
If others are happy with this, then before merging needs:
  * Tests.
  * Examples/docs.
  * `///` comments.

AnthonySteele:
.. and it needs to be wired in 
* when registering handlers
* when finding handler

AnthonySteele:
This plan effectively broadens the handler contract again, adding MessageContext

It is fairly simple to implement in JustSaying, and fairly simple to update consumers, but

* You always have the context, even when you don't have a use for it, or even know what to do with it.
* JustSaying always has to create the context for each message.

An alternate proposal is this:

Have an interface `IMessageContextAccessor` with a method `MessageContext GetContext()`

A Handler can constructor-inject this. And later call it to get the context as needed.

## Pros:

* handlers that don't care about context stay as they are.
* JustSaying won't need to create the context object unless it's actually got.

## Cons:

* simpler handlers stay simple, but complex things become harder. if a handler does want the message context, there are more non-obvious moving parts in order to get it
* Harder to implement inside JustSaying as `IMessageContextAccessor` would have to be transient in whichever container is in use. Initial investigation raised issue #484 for starters. More code needed to make it work.

slang25:
Great write-up @AnthonySteele.

> IMessageContextAccessor would have to be transient in whichever container is in use
It would be registered as a singleton (because it's a kinda a factory), and we'd do it for them in `AddJustSaying`.

There's also the pro that if your handler has some dependencies, you could push the `IMessageContextAccessor` detail into those types, which keeping your handler clean of the extra AWS specific concerns.

As an example, maybe I'd have a `IRetryHelper` (naming is hard 😳), which might want to know about the attempt number, or visibility timeout. You could have:
```csharp
class RetryHelper : IRetryHelper
{
    public RetryHelper(IMessageContextAccessor contextAccessor)
   {
       // later on do something with contextAccessor.GetContext().AttemptNumber (or whatever)
```

AnthonySteele:
> It would be registered as a singleton

That underscores that we can't do things this way unless we find out a good way to do that at all. Yet it makes sense to register `IMessageContextAccessor` as a singleton. However it's a singleton that has to return _the SQS message currently being processed_ and other transient state.

slang25:
So what I'm proposing is that we do exactly what aspnetcore does, which backs it by AsyncLocal. It's sort of an ambient context, but with a nice way to get hold of it.
I have 2/3rds of it done here and will hopefully share it before the weekend.

AnthonySteele:
[FYI, "exactly what aspnetcore does"](https://github.com/aspnet/HttpAbstractions/blob/master/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs)

slang25:
Thank you, I had lost that link. Sometimes code says a thousand words 😄

martincostello:
My thinking on the accessor would be something like one of the two ideas:

## Option A - Always present

In this example, the accessor is always in DI and things will break without it.

```csharp
// Any relevant pre-processing...

var accessor = resolver.GetInstance<IMyAccessorThing>();
accessor.Current = new AwsContextThing();

// All the stuff that processes things...

accessor.Current = null; // We're done with this message completely
```

## Option B - Accessor is optional

In this example, it's opt-in for if you need it (to save allocations etc.) like how `IHttpContextAccessor` is opt-in.

```csharp
// On setup
services.AddJustSaying(...)
             .AddJusSayingContextOrWhateverForAwsContextAccessStuff();

// Any relevant pre-processing...

var accessor = resolver.GetInstance<IMyAccessorThing>();

if (accessor != null)
{
    accessor.Current = new AwsContextThing();
}

// All the stuff that processes things...

if (accessor != null)
{
    accessor.Current = null; // We're done with this message completely
}
```

AnthonySteele:
What iw as getting at is that `IMessageContextAccessor` merely a wrapper, the real decision is to put data into the "async thread local storage" just before invoking the message handler or not. We can base this on performance data: if it's cheap then do it every time, if not then have some kind of switch that matches up with the availability of a  `IMessageContextAccessor`  as set up by some kind of opt-in.

Like assess to a current http context, this _might_ lead people into bad designs - e.g. with http context, it is better that the controller has the sole responsibility for dealing with http request and responses, and classes that it calls to do domain things do not share this concern. However "you can use it to make bad designs" is not a fatal flaw, it is a pitfall to be wary of.

AnthonySteele:
Hmm, good points, paring this down to the essentials, the DI aspect is unnecessary. If the data is in async local storage, then a static helper method will be enough to get it.

slang25:
Ah yes, I'm with you now 👍

slang25:
I like the optional thing, it seems like a nice pattern. The only downside I can think of is that if a user forgets the registration and is using `IMessageContextAccessor` then they'd get a `NullReferenceException` and we couldn't give a taylored message. Feels like a reasonable trade off though.

AnthonySteele:
Closed in favour of the option discussed in the comments

martincostello:
The null thing was why the code example had the null checks 😉

AnthonySteele:
You don't _need_  a `IMessageContextAccessor` interface to access thread-local static storage. To do so in a typesafe manner all you need is a couple of static methods. However @martincostello makes a good case that putting this behind an interface means it's much easier to mock out for test etc.

AnthonySteele:
See #486

PR #481, opened 2018-12-13
Migrate publish integration test

martincostello:
This PR migrates the first existing integration test to the new builder API (#431), and fixes a few bugs and other improvements made while migrating it.

  1. Fixes SQS validation being run before the queue/topic names were set.
  1. Add convenience methods for setting queue and topic names.
  1. Fix several incorrect or missing XML comments.
  1. Improve messaging if a publisher is not configured.
  1. Add convenience DI method that wraps usage of `IMessageBusConfigurationContributor`.

Relates to #380.


JustSaying.IntegrationTests/Fluent/Publishing/WhenAMessageIsPublishedToAQueue.cs
```diff
@@ -0,0 +1,53 @@
...
+            string content = Guid.NewGuid().ToString();
+
+            var message = new SimpleMessage()
+            {
+                Content = content
+            };
+
+            await WhenAsync(
+                services,
+                async (publisher, listener, cancellationToken) =>
```
AnthonySteele: Stylistic, but when you have a long complex lambda, I prefer to extract it as a method with a descriptive name. In this case it would be something like `PublishMessageAndWaitForReceipt` ?

JustSaying.IntegrationTests/Fluent/Publishing/WhenAMessageIsPublishedToATopic.cs
```diff
@@ -0,0 +1,53 @@
...
+            var message = new SimpleMessage()
+            {
+                Content = content
+            };
+
+            await WhenAsync(
+                services,
+                async (publisher, listener, cancellationToken) =>
+                {
+                    listener.Start(cancellationToken);
```
AnthonySteele: mthod as above

JustSaying.IntegrationTests/Fluent/Publishing/WhenAMessageIsPublishedToAQueue.cs
martincostello: That then doesn't really flow as a simple all-in-one test method you just read and see what it's doing, as then you've got the _Arrange_ in the test method, and the _Act_ and _Assert_ off somewhere else in another method that the test method calls.

Happy to refactor further, but I think there readability is more important over "efficient engineering"

AnthonySteele: I find the named method more readable.

martincostello: YMMV 🤷‍♂️

martincostello:
I'll merge this for now with no changes, but bear the comment in mind. I imagine the tests will iterate as JustBehave is slowly removed as different bits of duplication and repetition get flushed out.

PR #485, opened 2018-12-19
MessageMonitor values as timespan

AnthonySteele:

Duration values are typed as `TimeSpan` not `long` in methods on `IMessageMonitor`

In line with what we now do in Statsd, values that express a duration are more correctly typed as `TimeSpan` not `long` (implicitly milliseconds)
I noticed this while looking at the consumer that sends these to statsd, which _formerly_ required values as long milliseconds, but now [also accepts `TimeSpan` structs](https://github.com/justeat/JustEat.StatsD/blob/master/src/JustEat.StatsD/IStatsDPublisherExtensions.cs#L159). So we keep the  value as the richer `TimeSpan` type until the last responsible place.

### DIFF
--- a/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
+++ b/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
@@ -5,11 +5,11 @@ namespace JustSaying.Messaging.Monitoring
     public interface IMessageMonitor
     {
         void HandleException(Type messageType);
-        void HandleTime(long handleTimeMs);
+        void HandleTime(TimeSpan handleTime);
         void IssuePublishingMessage();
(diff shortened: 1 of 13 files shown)


JustSaying/Messaging/Monitoring/IMessageMonitor.cs
```diff
@@ -5,11 +5,11 @@ namespace JustSaying.Messaging.Monitoring
     public interface IMessageMonitor
     {
         void HandleException(Type messageType);
-        void HandleTime(long handleTimeMs);
+        void HandleTime(TimeSpan handleTime);
```
martincostello: Suggest the name `duration` for these parameters' names.

PR #486, opened 2018-12-20
message context store

AnthonySteele:

Message context store in async local.
Handlers will read this via `IMessageContextReader`.
Inside JustSaying it will be written via `IMessageContextAccessor`.
These are implemented by the same object.
But handlers should not be able to write it.


JustSaying/Messaging/MessageHandling/MessageContextReader.cs
```diff
@@ -0,0 +1,21 @@
+using System.Threading;
+
+namespace JustSaying.Messaging.MessageHandling
+{
+#pragma warning disable CA1822 // Member Read does not access instance data and can be marked as static
```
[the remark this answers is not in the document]
AnthonySteele: Good spot. Would have got to this in the next step anyway ;)

JustSaying/Messaging/MessageHandling/IMessageContextReader.cs
```diff
@@ -0,0 +1,7 @@
+namespace JustSaying.Messaging.MessageHandling
+{
+    public interface IMessageContextReader
+    {
+        MessageContext Read();
```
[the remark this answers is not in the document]
AnthonySteele: would it be `ReadMessageContext` or `ReadFromAsyncLocal` or ?

slang25:
Why not use a getter and setter?

JustSaying.UnitTests/AwsTools/MessageHandling/MessageContextReaderTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        [Fact]
+        public async Task DifferentThreadsHaveDifferentContexts()
+        {
+            var data1 = MakeTestMessageContext();
+            var data2 = MakeTestMessageContext();
+
+            var t1 = Task.Run(async () => await RunWithData(data1));
+            var t2 =  Task.Run(async () => await RunWithData(data2));
+
+            await Task.WhenAll(t1, t2);
```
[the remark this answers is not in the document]
AnthonySteele: There are asserts in `RunWithData`. if the assert fails, the thread throws, and the ` await Task.WhenAll` rethrows it.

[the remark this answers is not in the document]
martincostello: Maybe rename `RunWithData()` to make it clearer it's not just an "Act" method.

JustSaying/Messaging/MessageHandling/IMessageContextReader.cs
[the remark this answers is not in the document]
martincostello: If we're making this model `IHttpContextAccessor`, why isn't it just a simple property with a get and a set? As it stands we've got an interface that's a front to a static object (which doesn't fix the testability issues because it's still a shared static) which has an internal set method.

JustSaying/Messaging/MessageHandling/MessageContextReader.cs
```diff
@@ -2,9 +2,7 @@
 
 namespace JustSaying.Messaging.MessageHandling
 {
-#pragma warning disable CA1822 // Member Read does not access instance data and can be marked as static
-
-    public class MessageContextReader
+    public class MessageContextReader: IMessageContextReader
```
martincostello: Nit: space before `:`.

JustSaying/Messaging/MessageHandling/IMessageContextReader.cs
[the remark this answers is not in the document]
AnthonySteele: A mockable "Get" allows you to supply fake data to a handler (test not yet written)
Point taken that also adding a mockable "set" also allows you to inspect what's happening in integration / localstack scenarios.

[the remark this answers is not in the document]
shaynevanasperen: Also, they're doing some fancy object indirection trick there. Should we be doing the same?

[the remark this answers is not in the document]
AnthonySteele: I think @shaynevanasperen means `private class HttpContextHolder` [here](https://github.com/aspnet/HttpAbstractions/blob/master/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs). 
I don't know of any benefit of adding that yet?

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -71,12 +71,13 @@ public async Task DispatchMessage(SQSMessage message, CancellationToken cancella
 
             var handlingSucceeded = false;
             Exception lastException = null;
+            IMessageContextAccessor ctxAx = new MessageContextAccessor();
```
martincostello: This will need wiring up into `IServiceResolver` at a later point so it's actually injectable.

AnthonySteele:
The only downside of a read/write property is that the consumer (in a handler) is _never_ expected to call `set` and so perhaps should not be given a way to do that

JustSaying/Messaging/MessageHandling/IMessageContextReader.cs
```diff
@@ -0,0 +1,7 @@
+namespace JustSaying.Messaging.MessageHandling
+{
+    public interface IMessageContextReader
```
shaynevanasperen: Since it let's us read and write, this should be named something like `IMessageContextAccessor`.

martincostello:
The point of the setter to to create an abstraction to JustSaying setting the context itself.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
AnthonySteele: Commit wording is: temporary code in MessageDispatcher, intentionally ugly

JustSaying/Messaging/MessageHandling/IMessageContextAccessor.cs
```diff
@@ -1,5 +1,10 @@
 namespace JustSaying.Messaging.MessageHandling
 {
+    public interface IMessageContextReader
```
martincostello: Move to `IMessageContextReader.cs`

JustSaying/Messaging/MessageHandling/IMessageContextAccessor.cs
```diff
@@ -0,0 +1,7 @@
+namespace JustSaying.Messaging.MessageHandling
+{
+    public interface IMessageContextAccessor
```
slang25: If there are going to be 2 interfaces, then shall we mark this one be internal?

JustSaying/Messaging/MessageHandling/MessageContextAccessor.cs
```diff
@@ -0,0 +1,15 @@
+using System.Threading;
+
+namespace JustSaying.Messaging.MessageHandling
+{
+    public class MessageContextAccessor : IMessageContextReader, IMessageContextAccessor
```
slang25: `internal sealed`?

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -21,6 +22,7 @@ public class MessageDispatcher
         private readonly Action<Exception, SQSMessage> _onError;
         private readonly HandlerMap _handlerMap;
         private readonly IMessageBackoffStrategy _messageBackoffStrategy;
+        private readonly IMessageContextAccessor _messageContextWriter;
```
adammorr: Rename this to `_messageContextAccessor`

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -36,6 +36,7 @@ public class SqsNotificationListener : INotificationSubscriber
             IMessageSerializationRegister serializationRegister,
             IMessageMonitor messagingMonitor,
             ILoggerFactory loggerFactory,
+            IMessageContextAccessor messageContextAccessor,
```
slang25: This is probably because it's wip, but this class isn't being resolved by DI yet

JustSaying/Messaging/MessageHandling/MessageContextAccessor.cs
martincostello: It needs to be public for DI to be able to register it. Otherwise we'll be going down the `InternalsVisibleTo` route.

JustSaying/Messaging/MessageHandling/IMessageContextAccessor.cs
martincostello: Needs to be public for DI and utility in others' tests.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
AnthonySteele: yep

slang25: Who needs the whole LiveShare coding when you can get feedback commit by commit in real-time!

JustSaying.IntegrationTests/WhenRegisteringHandlersViaResolver/HandlerWithMessageContext.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using System.Threading.Tasks;
+using JustSaying.IntegrationTests.TestHandlers;
+using JustSaying.Messaging.MessageHandling;
+
+namespace JustSaying.IntegrationTests.WhenRegisteringHandlersViaResolver
+{
+    class HandlerWithMessageContext : IHandlerAsync<OrderPlaced>
```
martincostello: Explicit accessibility modifier.

JustSaying.IntegrationTests/WhenRegisteringHandlersViaResolver/HandlerWithMessageContext.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        private readonly IMessageContextReader _contextReader;
+
+        public HandlerWithMessageContext(
+            Future<OrderPlaced> future,
+            IMessageContextReader contextReader)
+        {
+            Future = future;
+            _contextReader = contextReader;
```
slang25: I was thinking the pattern we'd encourage would be to grab the context in the constructor. You get the same result either way, it keeps the code in the handler method a little bit tidier.
```csharp
_context = contextReader.MessageContext;
```

JustSaying.IntegrationTests/WhenRegisteringHandlersViaResolver/HandlerWithMessageContext.cs
```diff
@@ -0,0 +1,38 @@
...
+        public async Task<bool> Handle(OrderPlaced message)
+        {
+            var context = _contextReader.MessageContext;
+
+            if (context == null)
+            {
+                throw new InvalidOperationException("Message context was not found");
+            }
+
+            Console.WriteLine($"Message context found with queue {context.QueueUri}");
```
martincostello: This won't work in xunit tests. You need to use `ITestOutputHelper`.

JustSaying.IntegrationTests/WhenRegisteringHandlersViaResolver/WhenRegisteringAHandlerWithMessageContext.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Threading.Tasks;
+using JustSaying.IntegrationTests.TestHandlers;
+using JustSaying.TestingFramework;
+using Shouldly;
+using StructureMap;
+using Xunit;
+
+namespace JustSaying.IntegrationTests.WhenRegisteringHandlersViaResolver
+{
+    [Collection(GlobalSetup.CollectionName)]
```
martincostello: It would be nice if new tests used the new builder pattern so I've got less to migrate from the old world to the new...

JustSaying.TestingFramework/TestEnvironment.cs
```diff
@@ -46,7 +46,7 @@ public static Uri SimulatorUrl
         {
             get
             {
-                var awsEnv = Environment.GetEnvironmentVariable("AWS_SERVICE_URL");
+                var awsEnv = "http://localhost:4100";
```
martincostello: I guess you forgot to undo this.

JustSaying.Extensions.DependencyInjection.Microsoft/IServiceCollectionExtensions.cs
```diff
@@ -124,6 +124,8 @@ public static IServiceCollection AddJustSaying(this IServiceCollection services,
             services.TryAddSingleton<IAwsClientFactoryProxy>((p) => new AwsClientFactoryProxy(p.GetRequiredService<IAwsClientFactory>));
             services.TryAddSingleton<IMessagingConfig, MessagingConfig>();
             services.TryAddSingleton<IMessageMonitor, NullOpMessageMonitor>();
+            services.TryAddSingleton<IMessageContextAccessor, MessageContextAccessor>();
```
martincostello: This won't work how you think it will - I'll explain verbally when you're free.

JustSaying.IntegrationTests/Fluent/Subscribing/HandlerWithMessageContext.cs
```diff
@@ -0,0 +1,43 @@
...
+using JustSaying.IntegrationTests.TestHandlers;
+using JustSaying.Messaging.MessageHandling;
+using JustSaying.TestingFramework;
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.IntegrationTests.Fluent.Subscribing
+{
+    internal class HandlerWithMessageContext : IHandlerAsync<SimpleMessage>
+    {
+        private readonly ILogger<HandlerWithMessageContext> _outputHelper;
```
martincostello: Rename to `_logger`.

JustSaying.IntegrationTests/Fluent/Subscribing/HandlerWithMessageContext.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public Future<SimpleMessage> Future { get; }
+
+        public async Task<bool> Handle(SimpleMessage message)
+        {
+            var messageContext = _messageContextReader.MessageContext;
+
+            if (messageContext == null)
+            {
+                throw new InvalidOperationException("Message context was not ");
```
martincostello: _"... was not "_ what?

JustSaying.IntegrationTests/Fluent/Subscribing/WhenAHandlerUsesMessageContext.cs
```diff
@@ -0,0 +1,61 @@
...
+using JustSaying.Messaging;
+using JustSaying.Messaging.MessageHandling;
+using JustSaying.TestingFramework;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+using Xunit.Abstractions;
+
+namespace JustSaying.IntegrationTests.Fluent.Subscribing
+{
+    public class WhenAHandlerUsesMessageContext: IntegrationTestBase
```
martincostello: Nit: space before `:`.

JustSaying.IntegrationTests/Fluent/Subscribing/WhenAHandlerUsesMessageContext.cs
```diff
@@ -0,0 +1,61 @@
...
+            // Arrange
+            var future = new Future<SimpleMessage>();
+            var accessor = new RecordingMessageContextAccessor(new MessageContextAccessor());
+
+            var services = GivenJustSaying()
+                .ConfigureJustSaying((builder) =>
+                    builder.Publications((options) => options.WithQueue<SimpleMessage>(UniqueName)))
+                .ConfigureJustSaying((builder) =>
+                    builder.Subscriptions((options) => options.ForQueue<SimpleMessage>(UniqueName)))
+
```
martincostello: Delete this blank line?

martincostello:
Looks like the StructureMap DI package needs updating as well.

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -80,7 +80,15 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
...
-                await Task.WhenAny(action(publisher, listener, source.Token), Task.Delay(Timeout, source.Token)).ConfigureAwait(false);
+                var actionTask = action(publisher, listener, source.Token);
+
+                await Task.WhenAny(actionTask, Task.Delay(Timeout, source.Token))
+                    .ConfigureAwait(false);
+
+                if (!actionTask.IsCompletedSuccessfully)
+                {
+                    throw actionTask.Exception ?? new Exception("The action failed");
+                }
```
martincostello: The other thing to add would be `source.Token.ThrowIfCancelled()` (sic).

JustSaying.Extensions.DependencyInjection.Microsoft/IServiceCollectionExtensions.cs
```diff
@@ -124,8 +124,11 @@ public static IServiceCollection AddJustSaying(this IServiceCollection services,
             services.TryAddSingleton<IAwsClientFactoryProxy>((p) => new AwsClientFactoryProxy(p.GetRequiredService<IAwsClientFactory>));
             services.TryAddSingleton<IMessagingConfig, MessagingConfig>();
             services.TryAddSingleton<IMessageMonitor, NullOpMessageMonitor>();
-            services.TryAddSingleton<IMessageContextAccessor, MessageContextAccessor>();
-            services.TryAddSingleton<IMessageContextReader, MessageContextAccessor>();
+
+            var messageContextAccessor = new MessageContextAccessor();
```
martincostello: This isn't lazy, so IMHO the nicer way is to do it like this:

```csharp
services.TryAddSingleton<MessageContextAccessor>();
services.TryAddSingleton<IMessageContextAccessor>(serviceProvider => serviceProvider.GetRequiredService<MessageContextAccessor>());
services.TryAddSingleton<IMessageContextReader>(serviceProvider => serviceProvider.GetRequiredService<MessageContextAccessor>());
```

Example:

https://github.com/aspnet/Extensions/blob/557995ec322f1175d6d8a72a41713eec2d194871/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs#L36-L38

JustSaying/Fluent/ServicesBuilder.cs
```diff
@@ -152,5 +157,11 @@ public ServicesBuilder WithMessageSerializationFactory(Func<IMessageSerializatio
             MessageSerializationFactory = factory ?? throw new ArgumentNullException(nameof(factory));
             return this;
         }
+
+        public ServicesBuilder WithMessageContextAccessor(Func<IMessageContextAccessor> messageContextAccessor)
```
martincostello: `///` docs once tests passing please.

JustSaying.Extensions.DependencyInjection.StructureMap/JustSayingRegistry.cs
```diff
@@ -30,6 +31,10 @@ public JustSayingRegistry()
             For<IMessageSubjectProvider>().Use<GenericMessageSubjectProvider>().Singleton();
             For<IVerifyAmazonQueues>().Use<AmazonQueueCreator>().Singleton();
 
+            var messageContextAccessor = new MessageContextAccessor();
```
martincostello: Again, can you make it lazy?

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
slang25: If the `actionTask` errors, then the exception will bubble up during `await WhenAny`. It might be worth copying the example here:
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md
If you search for the text:
`This example cancels the timer if the operation succesfully completes.`

slang25:
```csharp
using (var source = new CancellationTokenSource(Timeout))
{
    var delayTask = Task.Delay(Timeout, source.Token);
    var resultTask = await Task.WhenAny(action(publisher, listener, source.Token), delayedTask).ConfigureAwait(false);
    
    if (resultTask == delayTask)
    {
        throw new OperationCanceledException();
    }
    else
    {
        cts.Cancel();
    }
}
```

martincostello: 1) Why not just call `source.Token.ThrowIfCancelled()` if it's the task that finished?
2) If it didn't timeout, do we need to even bother canceling it?

slang25: For 2, it just stops having timers lying about, but for tests it's probably not that important.
1, yeah that's fair.

```csharp
using (var source = new CancellationTokenSource(Timeout))
{
    var resultTask = await Task.WhenAny(action(publisher, listener, source.Token),
                                        Task.Delay(Timeout, source.Token)).ConfigureAwait(false);
    
    source.Token.ThrowIfCancelled();
    cts.Cancel();
    await resultTask;
}
```

martincostello: Doesn't the dispose block for the CTS take care of that?

slang25: I don't think it cancels, just doing some tests

AnthonySteele: `await resultTask` or `await actionTask` might not be safe - what if it's deadlocked or very slow. 
That's the point of the `delayTask`, right?

slang25: confirmed, It doesn't cancel

slang25: You wouldn't be awaiting a task that is slow or stuck because you'd be at that line only if the delay task completes, then the `source.Token.ThrowIfCancelled()` guards you from the case it was because of a timeout.

martincostello: TIL.

AnthonySteele: Move this discussion to #487

JustSaying/Fluent/ServicesBuilder.cs
```diff
@@ -158,9 +158,17 @@ public ServicesBuilder WithMessageSerializationFactory(Func<IMessageSerializatio
             return this;
         }
 
-        public ServicesBuilder WithMessageContextAccessor(Func<IMessageContextAccessor> messageContextAccessor)
+        /// <summary>
+        /// Specifies the <see cref="IMessageContextAccessor"/> to use.
+        /// </summary>
+        /// <param name="contextAccessor"></param>
```
martincostello: Just copy the text from another method where this parameter has some text, rather than leave it empty.

JustSaying/Messaging/MessageHandling/IMessageContextAccessor.cs
```diff
@@ -2,6 +2,9 @@ namespace JustSaying.Messaging.MessageHandling
 {
     public interface IMessageContextAccessor
     {
+        /// <summary>
+        /// Get or set the context metadata about the SQS message currently being processed
```
martincostello: Nit: full stop.

JustSaying/Messaging/MessageHandling/IMessageContextReader.cs
```diff
@@ -2,6 +2,9 @@ namespace JustSaying.Messaging.MessageHandling
 {
     public interface IMessageContextReader
     {
+        /// <summary>
+        /// Get the context metadata about the SQS message currently being processed
```
martincostello: Nit: full stop.

JustSaying/Messaging/MessageHandling/MessageContext.cs
```diff
@@ -11,7 +11,14 @@ public MessageContext(SQSMessage message, Uri queueUri)
             QueueUri = queueUri ?? throw new ArgumentNullException(nameof(queueUri));
         }
 
+        /// <summary>
+        /// The AWS SQS Message that is currently being processed
```
martincostello: Nit: Full stop.

JustSaying/Messaging/MessageHandling/IMessageContextReader.cs
martincostello: _"Gets..."_

JustSaying/Messaging/MessageHandling/IMessageContextAccessor.cs
martincostello: _"Gets or sets..."_

JustSaying/Messaging/MessageHandling/MessageContext.cs
martincostello: _"Gets..."_

JustSaying/Messaging/MessageHandling/MessageContext.cs
```diff
@@ -11,7 +11,14 @@ public MessageContext(SQSMessage message, Uri queueUri)
...
+        /// </summary>
         public SQSMessage Message { get; }
+
+        /// <summary>
+        /// The SQS Queue that the message was received on
```
martincostello: Nit: full stop.

martincostello: _"Gets..."_

slang25:
We've got couple of flakey integration tests, so I reran it until it worked \_(ツ)_/¯

martincostello:
I'll be doing a PR tomorrow to migrate more of the tests to the "new world", which might flush our some flakiness too.

AnthonySteele:
@slang25 I have also been clicking "Restart Job"

slang25:
😆 We've probably been killing each others run.

AnthonySteele:
OMG, 
we've been wanting a way to do this for at least 2 years
and this is the third (or so) different approach to it that has been prototyped.

PR #487, opened 2018-12-21
reliability in IntegrationTestBase

AnthonySteele:

Make `IntegrationTestBase.WhenAsync` fail when it should.

This PR turns the tests red, because they were _already_ failing, unnoticed.

### DIFF
--- a/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
+++ b/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
@@ -87,9 +87,9 @@ namespace JustSaying.IntegrationTests.Fluent
 
                 source.Token.ThrowIfCancellationRequested();
 
-                if (!actionTask.IsCompletedSuccessfully)
+                if (actionTask.IsFaulted)
                 {
-                    throw actionTask.Exception ?? new Exception("The action failed but did not have an exception");
+                    throw actionTask.Exception ?? new Exception("The action is faulted, but did not have an exception");
                 }
             }
         }
--- a/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
+++ b/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
@@ -80,7 +80,17 @@ namespace JustSaying.IntegrationTests.Fluent
 
             using (var source = new CancellationTokenSource(Timeout))
             {
-                await Task.WhenAny(action(publisher, listener, source.Token), Task.Delay(Timeout, source.Token)).ConfigureAwait(false);
+                var delayTask = Task.Delay(Timeout, source.Token);
+                var actionTask = action(publisher, listener, source.Token);
+
+                await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
+
+                source.Token.ThrowIfCancellationRequested();
+
+                if (!actionTask.IsCompletedSuccessfully)
+                {
+                    throw actionTask.Exception ?? new Exception("The action failed but did not have an exception");
+                }
             }
         }
     }
--- a/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
+++ b/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
@@ -89,7 +89,7 @@ namespace JustSaying.IntegrationTests.Fluent
 
                 if (actionTask.IsFaulted)
                 {
-                    throw actionTask.Exception ?? new Exception("The action is faulted, but did not have an exception");
+                    await actionTask;
                 }
             }
         }
--- a/JustSaying.IntegrationTests/Fluent/Subscribing/WhenAHandlerThrowsAnExceptionWithNoMonitor.cs
+++ b/JustSaying.IntegrationTests/Fluent/Subscribing/WhenAHandlerThrowsAnExceptionWithNoMonitor.cs
@@ -5,7 +5,6 @@ using JustSaying.Messaging;
 using JustSaying.Messaging.MessageHandling;
 using JustSaying.TestingFramework;
 using Microsoft.Extensions.DependencyInjection;
-using NSubstitute;
 using Shouldly;
 using Xunit.Abstractions;
 
@@ -18,7 +17,7 @@ namespace JustSaying.IntegrationTests.Fluent.Subscribing
         {
         }
 
-        [AwsFact]
+        [AwsFact(Skip = "WithErrorHandler on SqsReadConfigurationBuilder not working yet")]
         public async Task Then_The_Message_Is_Handled()
         {
             // Arrange
@@ -53,7 +52,6 @@ namespace JustSaying.IntegrationTests.Fluent.Subscribing
                     await handler.DoneSignal.Task;
 
                     // Assert
-                    await handler.Received().Handle(Arg.Any<SimpleMessage>());
                     handler.MessageReceived.ShouldNotBeNull();
                     handledException.ShouldBeTrue();
                 });
--- a/JustSaying.IntegrationTests/Fluent/Subscribing/WhenAHandlerThrowsAnExceptionWithAMonitor.cs
+++ b/JustSaying.IntegrationTests/Fluent/Subscribing/WhenAHandlerThrowsAnExceptionWithAMonitor.cs
@@ -45,7 +45,6 @@ namespace JustSaying.IntegrationTests.Fluent.Subscribing
                     await handler.DoneSignal.Task;
 
                     // Assert
-                    await handler.Received().Handle(Arg.Any<SimpleMessage>());
                     handler.MessageReceived.ShouldNotBeNull();
 
                     monitoring.Received().HandleException(Arg.Any<Type>());


JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -80,7 +80,17 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
...
             {
-                await Task.WhenAny(action(publisher, listener, source.Token), Task.Delay(Timeout, source.Token)).ConfigureAwait(false);
+                var delayTask = Task.Delay(Timeout, source.Token);
+                var actionTask = action(publisher, listener, source.Token);
+
+                await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
+
+                source.Token.ThrowIfCancellationRequested();
+
+                if (!actionTask.IsCompletedSuccessfully)
```
martincostello: If this is false because it hasn't finished, then the `throw` might fail. Maybe we need a separate branch in this block which throws with a "still running and hasn't failed" if it's still not done?

AnthonySteele: Then we can use `if (actionTask.IsFaulted)`

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -80,7 +80,17 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
...
+                var delayTask = Task.Delay(Timeout, source.Token);
+                var actionTask = action(publisher, listener, source.Token);
+
+                await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
+
+                source.Token.ThrowIfCancellationRequested();
+
+                if (actionTask.IsFaulted)
+                {
+                    throw actionTask.Exception ?? new Exception("The action is faulted, but did not have an exception");
```
slang25: Here it would be better to use `ExceptionDispatchInfo.Capture` & '.Throw()` so that the stacktrace is preserved. But it would be simpler to await the task to let the exception out.

AnthonySteele:
I don't see any reason to cancel the `source` - it gets disposed of when the `using` block ends, and is not forwarded to any other test code directly or via `delayTask`

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
slang25: I think this covers all bases:
```csharp
using (var source = new CancellationTokenSource(Timeout))
{
    var actionTask = action(publisher, listener, source.Token);
    
    await Task.WhenAny(actionTask, Task.Delay(Timeout, source.Token))
                    .ConfigureAwait(false);
    
    source.Token.ThrowIfCancelled();
    source.Cancel();
    await actionTask.ConfigureAwait(false);
}
```

CTS is disposed, and cancelled to ensure there's no left over timers.
If it times out, `OperationCanceledException` is thrown. 
The `"The action is faulted, but did not have an exception"` case isn't a real concern as `Exception` will be non-null in the case `IsFaulted` is true.

AnthonySteele: Interesting, I did not know about `ExceptionDispatchInfo.Capture` 

So what I'm seeing is that the stack trace from `WhenAsync` is on the `AggregateException`, and that wrapper contains the real exception with the correct stack trace. I'm happy with that.

slang25:
`delayTask` has a timer that continues to run. It's a very minor nit to be honest. If this were production code it might be worth doing.

AnthonySteele:
Putting `source.Cancel()` in actually breaks some tests

slang25:
That's interesting, I'd like to understand that. Feel free to leave it out, the benefits are negligible here.

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -80,7 +80,17 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
...
+                var delayTask = Task.Delay(Timeout, source.Token);
+                var actionTask = action(publisher, listener, source.Token);
+
+                await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
+
+                source.Token.ThrowIfCancellationRequested();
+
+                if (actionTask.IsFaulted)
+                {
+                    await actionTask;
```
slang25: You know in your heart of hearts this would look great on one line 😉

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
AnthonySteele: You're right about `new Exception(...)` being a nummer to make a warning go away. A faulted task has an exception.
So `source.Cancel()` breraks a few more tests, not doing that now.
I don't think `await actionTask` is safe in the general case (it might have hung) and you would be relying on having got past the timeout check at that point.

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -80,7 +80,17 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
 
             using (var source = new CancellationTokenSource(Timeout))
             {
-                await Task.WhenAny(action(publisher, listener, source.Token), Task.Delay(Timeout, source.Token)).ConfigureAwait(false);
+                var delayTask = Task.Delay(Timeout, source.Token);
```
shaynevanasperen: I'm just confused about the need for a delay task if that timeout is already encapsulated by the cancellation token?

AnthonySteele: You can't await a cancellation token, only a task

slang25: I think Shayne means why can't we trust action to respect the token. It could be blocked on something, it may not be checking the token as frequently as it should etc...

I hadn't realised we were using the same token for the timeout and passing it into the action. That might explain the weirdness when cancelling, my bad. If it was just for the timeout I think it would be ok.

shaynevanasperen: Why even have a delay task at all though? The cancellation token source should handle the timeout automatically, not so?

AnthonySteele: Ok, the fact that the same token also goes into the action would explain why cancelling it cancels running tasks and fails tests.
However I think it's the right thing to do, the alternative would be to give `CancellationToken.None` which never expires.

slang25: @shaynevanasperen my understanding is it's a defensive mistrust of the code under test. It means we fail fast.

AnthonySteele: In answer to the question "why do we wait for two task, one of which is a fixed time delay", we' don't trust CancellationToken to work in all dodgy code  / failure cases.

AnthonySteele:
There is a todo on the test `WhenAHandlerThrowsAnExceptionWithNoMonitor`
Which is failing because the `config.WithErrorHandler(ErrorHandler)` is put into a `SqsReadConfigurationBuilder` but is not reaching `JustSayingFluently.CreateSubscriptionListener`.

It needs to be fed into `new SqsNotificationListener(...)` but as far as I can see, has not yet been set up at that point.

This is out of scope here, this PR does what's intended by surfacing real errors and is also green.

JustSaying.IntegrationTests/Fluent/Subscribing/WhenAHandlerThrowsAnExceptionWithNoMonitor.cs
```diff
@@ -18,7 +17,7 @@ public WhenAHandlerThrowsAnExceptionWithNoMonitor(ITestOutputHelper outputHelper
         {
         }
 
-        [AwsFact]
+        [AwsFact(Skip = "WithErrorHandler on SqsReadConfigurationBuilder not working yet")]
```
martincostello: I'll look at fixing this tomorrow 👍

shaynevanasperen:
I find this to be quite confusing. The `source.Token.ThrowIfCancellationRequested();` line would throw from there even if the action task did that itself. Surely we'd want the stack trace to show it throwing from the action task rather than from within this test method? Either way, I think it needs some comments added for clarity.

Perhaps this is a bit more straightforward and easier to understand:
```cs
using (var source = new CancellationTokenSource(Timeout))
{
    // Create a delay task to ensure that our test never hangs indefinitely
    var delayTask = Task.Delay(Timeout.Add(Timeout), CancellationToken.None); // Don't use source.Token here, because we never intend to cancel the delay
    var actionTask = action(publisher, listener, source.Token);

    // If the actionTask completes first, it was successful
    await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);

    if (delayTask.IsCompleted)
    {
        // If we get here, the actionTask took too long
        throw new Exception("The action took too long");
    }

    // Either this will no-op (if the actionTask completed successfully) or it will throw (if the actionTask completed with a fault due to the timeout expiring)
    await actionTask;
}
```

martincostello:
Personally I'd just get the framework to throw what it would throw via the call designed for that purpose, rather than just throw `Exception`, the worst of all possible exception types to use.

The `await` at the end is probably fine as it'll return because it either already completed or will throw, but I don't care enough either way.

shaynevanasperen:
Perhaps all that's required then is to move the `source.Token.ThrowIfCancellationRequested();` line to the end?

martincostello:
It shouldn't be last, as otherwise if the test task deadlocked, the test wouldn't end with the timeout.

shaynevanasperen:
But if the test task deadlocked, then the delay task would finish first and because the delay task uses the same cancellation token, it would have finished due a to a timeout (although it might actually be a race condition because the non-faulted delay length is the same as the timeout).

The key thing is that we probably want to differentiate between the safety-net delay task (which only exists to prevent the test from hanging indefinitely) and the timeout passed to the test task.

PR #489, opened 2018-12-22
Replatform some integration tests

martincostello:
1. Migrate a number of integration tests to the new builder pattern.
  1. Fix error handler not being used due to it being set to late.
  1. Add methods to builder pattern to set server-side encryption.
  1. Add helper method to tests for testing simple queue/topic loopback configurations.
  1. Remove lambda to reduce overhead of default error handler.
  1. Some minor style refactoring.
  1. Separate source and tests into solution folders.

Relates to #380.


JustSaying/Fluent/TopicSubscriptionBuilder`1.cs
```diff
@@ -100,13 +100,14 @@ public TopicSubscriptionBuilder<T> WithReadConfiguration(Action<SqsReadConfigura
...
-                           .IntoQueue(TopicName)
-                           .WithMessageHandler<T>(resolver);
+                           .IntoQueue(TopicName);
 
             if (ConfigureReads != null)
             {
                 topic.ConfigureSubscriptionWith(ConfigureReads);
             }
+
+            topic.WithMessageHandler<T>(resolver);
```
AnthonySteele: I see the fix!

does the file name here really end with `1 ?

AnthonySteele: There are several files where the name ends in "`1.cs" 
Is that intentional? Change in a separate PR?

slang25: I've seen it in a few other repos and I like it cause you don't have to think about naming the file vs the type, sometimes there's a non-generic type that would clash.

martincostello: `IEnumerable.cs` vs. ``IEnumerable`1.cs`` for instance.

PR #490, opened 2018-12-23
Make timeout easier to reason with in IntegrationTestBase

slang25:
Following @shaynevanasperen comment, it isn't clear how the timeout task is behaving, is it finishing because it's duration is up, or is it cancelled because of the cancellation token being fired.

### DIFF
--- a/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
+++ b/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
@@ -99,7 +99,7 @@ namespace JustSaying.IntegrationTests.Fluent
 
                 if (resultTask == delayTask)
                 {
-                    throw new OperationCanceledException();
+                    throw new TimeoutException($"The bus related action took longer to execute than the timeout of {Timeout.TotalSeconds} seconds");
                 }
                 else
                 {
--- a/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
+++ b/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
@@ -90,26 +90,23 @@ namespace JustSaying.IntegrationTests.Fluent
             IMessagePublisher publisher = serviceProvider.GetRequiredService<IMessagePublisher>();
             IMessagingBus listener = serviceProvider.GetRequiredService<IMessagingBus>();
 
-            using (var source = new CancellationTokenSource(Timeout))
+            using (var cts = new CancellationTokenSource())
             {
-                try
+                var delayTask = Task.Delay(Timeout, cts.Token);
+                var actionTask = action(publisher, listener, cts.Token);
+
+                var resultTask = await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
+
+                if (resultTask == delayTask)
                 {
-                    var delayTask = Task.Delay(Timeout, source.Token);
-                    var actionTask = action(publisher, listener, source.Token);
-
-                    await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
-
-                    source.Token.ThrowIfCancellationRequested();
-
-                    if (actionTask.IsFaulted)
-                    {
-                        await actionTask;
-                    }
+                    throw new OperationCanceledException();
                 }
-                finally
+                else
                 {
-                    source.Cancel();
+                    cts.Cancel();
                 }
+                
+                await actionTask;
             }
         }
     }


JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -90,7 +90,7 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
             IMessagePublisher publisher = serviceProvider.GetRequiredService<IMessagePublisher>();
             IMessagingBus listener = serviceProvider.GetRequiredService<IMessagingBus>();
 
-            using (var source = new CancellationTokenSource(Timeout))
+            using (var source = new CancellationTokenSource())
```
AnthonySteele: This is not easy to reason about. Should it not cancel at some point? If not on a timeout upfront, then maybe after the `.WhenAny` - assuming that if `actionTask` goes wrong, then the `delayTask` will complete first?

Why do we do `Task.Delay(Timeout, source.Token)` rather than `Task.Delay(Timeout)` ? [The docs suggest](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=netframework-4.7.2#System_Threading_Tasks_Task_Delay_System_TimeSpan_System_Threading_CancellationToken_) that the first overload "observes" the token. But it won't be signalled, will it?

slang25: Thanks, I've messed this up for sure. As far as I can see, there are 2 patterns and we are straddling them. I'll push an update and ping you.

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -90,26 +90,23 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
...
-            using (var source = new CancellationTokenSource(Timeout))
+            using (var cts = new CancellationTokenSource())
             {
-                try
-                {
-                    var delayTask = Task.Delay(Timeout, source.Token);
-                    var actionTask = action(publisher, listener, source.Token);
-
-                    await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
+                var delayTask = Task.Delay(Timeout, cts.Token);
```
slang25: The much more trivial change would be to change this to:
```csharp
var delayTask = Task.Delay(-1, cts.Token);
```

slang25: However I find that very slightly more confusing because you won't be sure which of the 2 tasks from `WhenAny` was the one that finished first. Even with the code that follows it, it looks like a deliberate race condition.

slang25:
Thanks @AnthonySteele, updated.

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -90,26 +90,23 @@ protected async Task WhenAsync(IServiceCollection services, Func<IMessagePublish
...
-                    source.Token.ThrowIfCancellationRequested();
+                var resultTask = await Task.WhenAny(actionTask, delayTask).ConfigureAwait(false);
 
-                    if (actionTask.IsFaulted)
-                    {
-                        await actionTask;
-                    }
+                if (resultTask == delayTask)
+                {
+                    throw new OperationCanceledException();
```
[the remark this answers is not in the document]
slang25: This was by choice, only cause it was what was being done before implicitly.
I'll cook one up 👍

AnthonySteele:
Looks fine, but in this case I think it's important to "test the test" as this code has tripped us up by not failing when it should.

So, I'm happy so long as you can verify that it fails when it should, i.e. temporarily
- add a `await Task.Delay(longtime)` to a test action body, does it fail with timeout as expected?
- add a `thrown new Exception("some message)` to a test action body, does it fail with `some message` as expected?

AnthonySteele:
What does ` var delayTask = Task.Delay(Timeout, cts.Token);` give use over ` var delayTask = Task.Delay(Timeout);` ?

slang25:
The thing with `Task.Delay(Timeout, cts.Token)` when the cts has the same timeout is what happens first? The task completing, or the task cancelling?

So if you say that the cts has no timeout, then `Task.Delay(Timeout, cts.Token)` is very marginally better than `Task.Delay(Timeout)` because it doesn't leave the task timing down in the event the work finishes early (and you explicitly cancel the token). I'm happy to omit the token parameter here it if it makes it easier to read cause it's really such a minor nit. I try to default for correctness unless there's good reason not to, but I feel like this might be pedantic.

Another alternative is to let the cts timeout drive with `Task.Delay(-1, cts.Token)`, but then like I said you'll be cancelling your action task with the same token, so it feels like another race.

I'll run some tests with some sleeps in to make sure we get a good failure message 👍

slang25:
`await Task.Delay(25_000);`
image

`throw new Exception("some message");`
image

slang25:
🎵 Don't break my tests, my flakey breaky tests... 🎵

flakey-breaky-tests

PR #491, opened 2018-12-23
Switch to xenial on travis

slang25:
Curious to see if we see the same flakey tests with this distro, I understand the hosting setup is quite different than trusty. The tests work like a dream locally ¯\_(ツ)_/¯

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: xenial
+dist: bionic
 
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: bionic
+dist: xenial
 
 os:
   - linux


slang25:
Out of 6 CI runs:

- Unit Tests Passed 5/6. (`SimulatedListenLoop_WhenThrottlingDoesNotOccur_DoNotCallMessageMonitor` failed once for `[InlineData(50, 50)]`).
- Integration Tests Passed 6/6.

I've been seeing the integration tests passing 1/3 on the current ubuntu image, so this would hopefully put an end to frequently retrying PR builds.

I'm on the fence about sidestepping the issue, it would be nice to make the test code more robust, but I've had a quick look and can't see anything obvious, without recreating the issue it could be tricky.

martincostello:
Maybe we just ride it out until they’re all migrated over to the new API style?

AnthonySteele:
We seem to have an old setup if we're using "trusty".  See this release list: https://wiki.ubuntu.com/Releases travis can use the keywords "trusty" (Ubuntu 14.0), "xenial" (Ubuntu 16.0) or "bionic" (Ubuntu 18.0) to target them.

[I had similar issues elsewhere](https://github.com/NuKeeperDotNet/NuKeeper/pull/626#discussion_r244231110).

It really looks to me like "trusty" is not current, is not going to get new SDKs as promptly as the others, and we are better of with a recent version,  first choice "bionic".

slang25:
Look's like the latest travis supports is `xenial`, [see here](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system).

I agree with you @AnthonySteele, I don't think sitting it out makes much sense cause it's causing us a problem. We either intend to make the tests more robust, and/or update to a fresher build image.

I think given that there will likely be an overhaul of the tests, a good option is to upgrade now and not spend much time investigating the current tests.

AnthonySteele:
[Ubuntu 16.0, "Trusty" drops out of LTS in April 2019](https://wiki.ubuntu.com/Releases).  So I would put updating off of that in the general timeframe of the JustSaying 7.0 Milestone.

PR #492, opened 2018-12-24
Test the tests

AnthonySteele:

Automate on "test the tests" so that we don't get falsely  passing tests. Put the "WhenAsync" logic under test, verify that it
- succeeds when the action succeeds
- throws when the action is still running at the timeout
- throws if the action throws

### DIFF
--- a/JustSaying.IntegrationTests/Fluent/ActionRunnerTest.cs
+++ b/JustSaying.IntegrationTests/Fluent/ActionRunnerTest.cs
@@ -1,0 +1,49 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace JustSaying.IntegrationTests.Fluent
+{
+    public class ActionRunnerTest: IntegrationTestBase
+    {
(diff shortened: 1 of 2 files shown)


JustSaying.IntegrationTests/Fluent/ActionRunnerTest.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace JustSaying.IntegrationTests.Fluent
+{
+    public class ActionRunnerTest: IntegrationTestBase
```
martincostello: Space before `:` please.

JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Reflection.PortableExecutable;
```
martincostello: What’s this for?

AnthonySteele: PEBKAC, I think

Issue #415, opened 2018-11-04
Use structured logging syntax with ILogger


martincostello:
I noticed that the `SqsNotificationListener` class uses `ILogger` from .NET Standard, but using string interpolation rather than structured logging syntax, [for example](https://github.com/justeat/JustSaying/blob/0739fdf4d90d4e75d55a762ac4a029a434786de7/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs#L110):

```csharp
log.LogInformation($"[{task.Status}] Stopped Listening - {queueInfo}");
```

We should review the logging messages throughout the codebase and prefer structured-logging syntax when data is formatted into messages similar to the below example so that if a logging implementation such as Serilog is in use, logs can be filtered/searched by the `MessageTemplate` property as well as the `Message` property.

```csharp
log.LogInformation("[{TaskStatus}] Stopped Listening - {QueueInfo}", task.Status, queueInfo);
```

AnthonySteele:
If the Standard logger interface supports this syntax, then this would be great!

martincostello:
It's up to the actual logging implementation. Serilog and Application Insights do for instance, others might just end up formatting it internally anyway and not ever generate a `MessageTemplate`, in which case the end result is the same in the log as the way the code is now.

AnthonySteele:
I believe that NLog 4.5 (plain NLog without NLog.StructuredLogging) also [does structured logging in some form](https://github.com/NLog/NLog/wiki/How-to-use-structured-logging). It would be good to test it out by doing this.

AnthonySteele:
I think that this is fixed by #493 .

IMHO it needs beta testing with load and an ELK stack to be sure that everything is of production quality, but we're not ready for that just yet.

Close this issue now?

AnthonySteele:
They are all done, as far as I can see (and I checked)

PR #493, opened 2018-12-24
Review logging code and use "structured" format

AnthonySteele:

Review logging code and use the "structured" format and [Message Templates](https://messagetemplates.org/). 
Review wording and sentence structure. Try to make them readable, and use the same names throughout.

#415

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -60,15 +60,16 @@ namespace JustSaying.AwsTools.MessageHandling
             }
             catch (MessageFormatNotSupportedException ex)
             {
-                _logger.LogTrace("Could not handle SQS message {MessageId} {MessageBody}. No serializer found",
-                    message.MessageId, message.Body ?? string.Empty);
+                _logger.LogTrace("Could not handle message with Id '{MessageId}' because a deserialiser for the content is not configured. Message body: '{MessageBody}'.",
+                    message.MessageId, message.Body);
                 await DeleteMessageFromQueue(message.ReceiptHandle).ConfigureAwait(false);
                 _onError(ex, message);
                 return;
             }
             catch (Exception ex)
             {
-                _logger.LogError(0, ex, "Error deserializing message {MessageId}", message.MessageId);
+                _logger.LogError(0, ex, "Error deserializing message with Id '{MessageId}' and body '{MessageBody}'.",
+                    message.MessageId, message.Body);
                 _onError(ex, message);
                 return;
             }
@@ -92,8 +93,8 @@ namespace JustSaying.AwsTools.MessageHandling
             }
             catch (Exception ex)
             {
-                _logger.LogError(0, ex,
-                    "Error handling message {MessageId} {MessageBody}", message.MessageId, message.Body);
+                _logger.LogError(0, ex, "Error handling message with Id '{MessageId}' and body '{MessageBody}'.",
+                    message.MessageId, message.Body);
 
                 if (typedMessage != null)
                 {
@@ -129,7 +130,7 @@ namespace JustSaying.AwsTools.MessageHandling
             var handlerSucceeded = await handler(message).ConfigureAwait(false);
 
             watch.Stop();
-            _logger.LogTrace("Handled message of type {MessageType} in {TimeToHandle}. ",
+            _logger.LogTrace("Handled message of type {MessageType} in {TimeToHandle}.",
                 message.GetType(), watch.Elapsed);
             _messagingMonitor.HandleTime(watch.Elapsed);
 
@@ -151,7 +152,8 @@ namespace JustSaying.AwsTools.MessageHandling
         {
             if (TryGetApproxReceiveCount(message.Attributes, out int approxReceiveCount))
             {
-                var visibilityTimeoutSeconds = (int)_messageBackoffStrategy.GetBackoffDuration(typedMessage, approxReceiveCount, lastException).TotalSeconds;
+                var visibilityTimeout = _messageBackoffStrategy.GetBackoffDuration(typedMessage, approxReceiveCount, lastException);
+                var visibilityTimeoutSeconds = (int)visibilityTimeout.TotalSeconds;
 
                 try
                 {
@@ -166,7 +168,7 @@ namespace JustSaying.AwsTools.MessageHandling
                 }
                 catch (Exception ex)
                 {
-                    _logger.LogError(0, ex, "Failed to update message visibility timeout by {VisibilityTimeout} seconds", visibilityTimeoutSeconds);
+                    _logger.LogError(0, ex, "Failed to update message visibility timeout by {VisibilityTimeout} seconds.", visibilityTimeoutSeconds);
                     _onError(ex, message);
                 }
             }
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -24,7 +24,7 @@ namespace JustSaying.AwsTools.MessageHandling
         private readonly IMessageBackoffStrategy _messageBackoffStrategy;
         private readonly IMessageContextAccessor _messageContextAccessor;
 
-        private static ILogger _log;
+        private static ILogger _logger;
 
         public MessageDispatcher(
             SqsQueueBase queue,
@@ -41,7 +41,7 @@ namespace JustSaying.AwsTools.MessageHandling
             _messagingMonitor = messagingMonitor;
             _onError = onError;
             _handlerMap = handlerMap;
-            _log = loggerFactory.CreateLogger("JustSaying");
+            _logger = loggerFactory.CreateLogger("JustSaying");
             _messageBackoffStrategy = messageBackoffStrategy;
             _messageContextAccessor = messageContextAccessor;
         }
@@ -60,14 +60,15 @@ namespace JustSaying.AwsTools.MessageHandling
             }
             catch (MessageFormatNotSupportedException ex)
             {
-                _log.LogTrace($"Didn't handle message [{message.Body ?? string.Empty}]. No serializer setup");
+                _logger.LogTrace("Could not handle SQS message {MessageId} {MessageBody}. No serializer found",
+                    message.MessageId, message.Body ?? string.Empty);
                 await DeleteMessageFromQueue(message.ReceiptHandle).ConfigureAwait(false);
                 _onError(ex, message);
                 return;
             }
             catch (Exception ex)
             {
-                _log.LogError(0, ex, "Error deserializing message");
+                _logger.LogError(0, ex, "Error deserializing message {MessageId}", message.MessageId);
                 _onError(ex, message);
                 return;
             }
@@ -91,8 +92,8 @@ namespace JustSaying.AwsTools.MessageHandling
             }
             catch (Exception ex)
             {
-                var errorText = $"Error handling message [{message.Body}]";
-                _log.LogError(0, ex, errorText);
+                _logger.LogError(0, ex,
+                    "Error handling message {MessageId} {MessageBody}", message.MessageId, message.Body);
 
                 if (typedMessage != null)
                 {
@@ -128,7 +129,7 @@ namespace JustSaying.AwsTools.MessageHandling
             var handlerSucceeded = await handler(message).ConfigureAwait(false);
 
             watch.Stop();
-            _log.LogTrace($"Handled message - MessageType: {message.GetType()}");
+            _logger.LogTrace("Handled message {MessageType}", message.GetType().ToString());
             _messagingMonitor.HandleTime(watch.Elapsed);
 
             return handlerSucceeded;
@@ -164,7 +165,7 @@ namespace JustSaying.AwsTools.MessageHandling
                 }
                 catch (Exception ex)
                 {
-                    _log.LogError(0, ex, $"Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds");
+                    _logger.LogError(0, ex, "Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds", visibilityTimeoutSeconds);
                     _onError(ex, message);
                 }
             }
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -129,7 +129,8 @@ namespace JustSaying.AwsTools.MessageHandling
             var handlerSucceeded = await handler(message).ConfigureAwait(false);
 
             watch.Stop();
-            _logger.LogTrace("Handled message {MessageType}", message.GetType());
+            _logger.LogTrace("Handled message of type {MessageType} in {TimeToHandle}. ",
+                message.GetType(), watch.Elapsed);
             _messagingMonitor.HandleTime(watch.Elapsed);
 
             return handlerSucceeded;
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -60,7 +60,7 @@ namespace JustSaying.AwsTools.MessageHandling
             }
             catch (MessageFormatNotSupportedException ex)
             {
-                _logger.LogTrace("Could not handle message with Id '{MessageId}' because a deserialiser for the content is not configured. Message body: '{MessageBody}'.",
+                _logger.LogTrace("Could not handle message with Id '{MessageId}' because a deserializer for the content is not configured. Message body: '{MessageBody}'.",
                     message.MessageId, message.Body);
                 await DeleteMessageFromQueue(message.ReceiptHandle).ConfigureAwait(false);
                 _onError(ex, message);
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -165,7 +165,7 @@ namespace JustSaying.AwsTools.MessageHandling
                 }
                 catch (Exception ex)
                 {
-                    _logger.LogError(0, ex, "Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds", visibilityTimeoutSeconds);
+                    _logger.LogError(0, ex, "Failed to update message visibility timeout by {VisibilityTimeout} seconds", visibilityTimeoutSeconds);
                     _onError(ex, message);
                 }
             }
--- a/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -129,7 +129,7 @@ namespace JustSaying.AwsTools.MessageHandling
             var handlerSucceeded = await handler(message).ConfigureAwait(false);
 
             watch.Stop();
-            _logger.LogTrace("Handled message {MessageType}", message.GetType().ToString());
+            _logger.LogTrace("Handled message {MessageType}", message.GetType());
             _messagingMonitor.HandleTime(watch.Elapsed);
 
             return handlerSucceeded;
--- a/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
+++ b/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
@@ -56,7 +56,7 @@ namespace JustSaying.AwsTools.MessageHandling
             try
             {
                 var response = await Client.PublishAsync(request, cancellationToken).ConfigureAwait(false);
-                _eventLog.LogInformation("Published message with subject '{MessageSubject}' and content '{MessageBody}'",
+                _eventLog.LogInformation("Published message with subject '{MessageSubject}' and content '{MessageBody}'.",
                     request.Subject, request.Message);
 
                 if (MessageResponseLogger != null)
--- a/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
+++ b/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
@@ -56,7 +56,8 @@ namespace JustSaying.AwsTools.MessageHandling
             try
             {
                 var response = await Client.PublishAsync(request, cancellationToken).ConfigureAwait(false);
-                _eventLog.LogInformation($"Published message: '{request.Subject}' with content {request.Message}");
+                _eventLog.LogInformation("Published message: {requestSubject} with content {requestMessage}",
+                    request.Subject, request.Message);
 
                 if (MessageResponseLogger != null)
                 {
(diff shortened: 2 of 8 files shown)


JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -60,14 +60,15 @@ public async Task DispatchMessage(SQSMessage message, CancellationToken cancella
             }
             catch (MessageFormatNotSupportedException ex)
             {
-                _log.LogTrace($"Didn't handle message [{message.Body ?? string.Empty}]. No serializer setup");
+                _logger.LogTrace("Could not handle SQS message {MessageId} {MessageBody}. No serializer found",
```
martincostello: These sentences should read in a way  so when you’re reading the message with formatting applied to them they make sense without context: _”Could not handle message with Id “XXX” because a serializer for the content is not configured. Message body: “{“foo”: “bar”}””_ or similar, whose template would be something like: _”Could not handle message with Id {MessageId} because a serializer for the content is not configured. Message body: “{MessageBody}””_

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -128,7 +129,7 @@ private async Task<bool> CallMessageHandler(Message message)
             var handlerSucceeded = await handler(message).ConfigureAwait(false);
 
             watch.Stop();
-            _log.LogTrace($"Handled message - MessageType: {message.GetType()}");
+            _logger.LogTrace("Handled message {MessageType}", message.GetType().ToString());
```
martincostello: To string is redundant here, it will do that anyway I believe.

martincostello: _”Handled message of type {MessageType}.”_

martincostello: Could add the time too. _”Handled message of type {MessageType} in {TimeToHandle}.”_

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -164,7 +165,7 @@ private async Task UpdateMessageVisibilityTimeout(SQSMessage message, string rec
                 }
                 catch (Exception ex)
                 {
-                    _log.LogError(0, ex, $"Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds");
+                    _logger.LogError(0, ex, "Failed to update message visibility timeout by {visibilityTimeoutSeconds} seconds", visibilityTimeoutSeconds);
```
martincostello: I believe the general convention is to use PascalCase for the names of the properties.

martincostello:
I think @slang25 might be the best person to ask about that one.

AnthonySteele:
* Version of `Microsoft.Extensions.Logging.Abstractions` is best left for a separate PR.
* I want to see this one in action, the "structured logging" is verified when it gets messages to Kibana in the correct format. But that might have to wait.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -60,14 +60,16 @@ public async Task DispatchMessage(SQSMessage message, CancellationToken cancella
             }
             catch (MessageFormatNotSupportedException ex)
             {
-                _log.LogTrace($"Didn't handle message [{message.Body ?? string.Empty}]. No serializer setup");
+                _logger.LogTrace("Could not handle message with Id '{MessageId}' because a deserialiser for the content is not configured. Message body: '{MessageBody}'.",
```
martincostello: Question of consistency of whether we should use US English in the log messages because we use US English for the code identifiers, i.e. _deserializer_ vs. _deserialiser_.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -60,14 +60,16 @@ public async Task DispatchMessage(SQSMessage message, CancellationToken cancella
...
+                    message.MessageId, message.Body);
                 await DeleteMessageFromQueue(message.ReceiptHandle).ConfigureAwait(false);
                 _onError(ex, message);
                 return;
             }
             catch (Exception ex)
             {
-                _log.LogError(0, ex, "Error deserializing message");
+                _logger.LogError(0, ex, "Error deserializing message with Id '{MessageId}' and body '{MessageBody}'.",
```
martincostello: I _think_ there's overloads (unless that's something added in the 2.0.0 version) that allow you to not need to specify the `0` as the first argument.

martincostello: Again consistency, as here it's US English. I think we should use US English consistently.

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -72,9 +73,10 @@ public async Task PublishAsync(Message message, PublishMetadata metadata, Cancel
             catch (Exception ex)
             {
                 if (!ClientExceptionHandler(ex, message))
+                {
                     throw new PublishException(
-                        $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}",
-                        ex);
+                        $"Failed to publish message to SNS. TopicArn: {request.TopicArn} Subject: {request.Subject} Message: {request.Message}", ex);
```
martincostello: I think this should be `Topic ARN`

martincostello: Should have a separator (semi-colon?) between the elements in the second sentence.

```
Topic ARN: {request.TopicArn}; Subject: {request.Subject}; Message: {request.Message}.
```

JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -75,22 +74,23 @@ public async Task<bool> CreateAsync()
...
             {
-                var response = await Client.CreateTopicAsync(new CreateTopicRequest(TopicName)).ConfigureAwait(false);
+                var response = await Client.CreateTopicAsync(new CreateTopicRequest(TopicName))
+                    .ConfigureAwait(false);
 
                 if (!string.IsNullOrEmpty(response.TopicArn))
                 {
                     Arn = response.TopicArn;
-                    _log.LogInformation($"Created Topic: {TopicName} on Arn: {Arn}");
+                    _log.LogInformation("Created topic '{TopicName}' on Arn '{Arn}'.", TopicName, Arn);
```
martincostello: `ARN`.

JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -75,22 +74,23 @@ public async Task<bool> CreateAsync()
...
                     return true;
                 }
-                _log.LogInformation($"Failed to create Topic: {TopicName}");
+                _log.LogInformation("Failed to create topic '{TopicName}'.", TopicName);
```
martincostello: Should this be a warn, or is "failed" the wrong choice of words?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -98,21 +98,20 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
...
-            var queueInfo = $"Queue: {queue}, Region: {region}";
 
             // Run task in background
             // ListenLoop will cancel gracefully, so no need to pass cancellation token to Task.Run
             _ = Task.Run(async () =>
             {
                 await ListenLoop(cancellationToken).ConfigureAwait(false);
                 IsListening = false;
-                _log.LogInformation($"Stopped Listening - {queueInfo}");
+                _log.LogInformation("Stopped listening on queue '{QueueName}' in region '{Region}'.", queueName, region);
```
martincostello: I _think_ that structured logging already puts quotes around strings in log messages, so the usages of `'` can all be removed.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -98,21 +98,20 @@ public SqsNotificationListener WithMessageProcessingStrategy(IMessageProcessingS
 
         public void Listen(CancellationToken cancellationToken)
         {
-            var queue = _queue.QueueName;
+            var queueName = _queue.QueueName;
             var region = _queue.Region.SystemName;
```
martincostello: Rename to `regionName`?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -128,21 +127,23 @@ internal async Task ListenLoop(CancellationToken ct)
                     sqsMessageResponse = await GetMessagesFromSqsQueue(queueName, region, ct).ConfigureAwait(false);
                     var messageCount = sqsMessageResponse.Messages.Count;
 
-                    _log.LogTrace(
-                        $"Polled for messages - Queue: {queueName}, Region: {region}, MessageCount: {messageCount}");
+                    _log.LogTrace("Polled for messages on queue '{QueueName}' in region '{Region}', and got {MessageCount} messages.",
```
martincostello: _"received"_ rather than _"got"_?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -128,21 +127,23 @@ internal async Task ListenLoop(CancellationToken ct)
...
+                        queueName, region, messageCount);
                 }
                 catch (InvalidOperationException ex)
                 {
-                    _log.LogTrace(
-                        $"Could not determine number of messages to read from [{queueName}], region: [{region}]. Ex: {ex}");
+                    _log.LogTrace(0, ex, "Could not determine number of messages to read from queue '{QueueName}', in '{Region}'.",
```
martincostello: Comma not needed.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -128,21 +127,23 @@ internal async Task ListenLoop(CancellationToken ct)
...
+                          queueName, region);
                 }
                 catch (OperationCanceledException ex)
                 {
-                    _log.LogTrace($"Suspected no message in queue [{queueName}], region: [{region}]. Ex: {ex}");
+                    _log.LogTrace(0, ex, "Suspected no message on queue '{QueueName}', in region '{Region}'.",
```
martincostello: Comma not needed.

JustSaying/AwsTools/MessageHandling/SqsQueueByNameBase.cs
```diff
@@ -65,7 +65,7 @@ public virtual async Task<bool> CreateAsync(SqsBasicConfiguration queueConfig, i
                     await Client.SetQueueAttributesAsync(queueResponse.QueueUrl, GetCreateQueueAttributes(queueConfig)).ConfigureAwait(false);
                     await SetQueuePropertiesAsync().ConfigureAwait(false);
 
-                    _log.LogInformation($"Created Queue: {QueueName} on Arn: {Arn}");
+                    _log.LogInformation("Created queue '{QueueName}' on arn '{Arn}'.", QueueName, Arn);
```
martincostello: `ARN`

martincostello: Would it be more _with_ than _on_?

JustSaying/AwsTools/MessageHandling/SqsQueueByNameBase.cs
```diff
@@ -74,26 +74,28 @@ public virtual async Task<bool> CreateAsync(SqsBasicConfiguration queueConfig, i
                 if (ex.ErrorCode == "AWS.SimpleQueueService.QueueDeletedRecently")
                 {
                     // Ensure we wait for queue delete timeout to expire.
-                    _log.LogInformation($"Waiting to create Queue due to AWS time restriction - Queue: {QueueName}, AttemptCount: {attempt + 1}");
+                    _log.LogInformation("Waiting to create queue '{QueueName}', due to AWS time restriction. Attempt number {attemptCount}.",
```
martincostello: Maybe add into the message that it's going to wait a minute before trying again?

JustSaying/AwsTools/MessageHandling/SqsQueueByNameBase.cs
```diff
@@ -74,26 +74,28 @@ public virtual async Task<bool> CreateAsync(SqsBasicConfiguration queueConfig, i
...
-                    _log.LogError(0, (Exception) ex, $"Create Queue error: {QueueName}");
+                    _log.LogError(0, ex, "Error trying to create queue '{QueueName}'.", QueueName);
                     throw;
                 }
 
                 // If we're on a delete timeout, throw after 2 attempts.
                 if (attempt >= 2)
                 {
-                    _log.LogError(0, (Exception) ex, $"Create Queue error, max retries exceeded for delay - Queue: {QueueName}");
+                    _log.LogError(0, ex, "Error trying to create queue '{QueueName}'. Maximum retries exceeded for delay.", QueueName);
```
martincostello: Include the delay?

JustSaying/AwsTools/MessageHandling/SqsQueueByNameBase.cs
```diff
@@ -74,26 +74,28 @@ public virtual async Task<bool> CreateAsync(SqsBasicConfiguration queueConfig, i
...
                     throw;
                 }
             }
 
-            _log.LogInformation($"Failed to create Queue: {QueueName}");
+            _log.LogInformation("Failed to create queue '{QueueName}'.", QueueName);
```
martincostello: Again, _failed_ doesn't seem to suggest an informational message.

JustSaying/JustSayingBus.cs
```diff
@@ -167,22 +167,21 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
             {
                 activeRegion = Config.GetActiveRegion();
             }
-            _log.LogInformation($"Active region has been evaluated to {activeRegion}");
+            _log.LogInformation("Active region has been evaluated to '{Region}'.", activeRegion);
```
martincostello: Given how often this gets logged and rarely changes, maybe we should either:
  1. Only log if it _changes_ after the first log. Or:
  1. Log it at trace.

JustSaying/JustSayingBus.cs
```diff
@@ -167,22 +167,21 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
...
 
             if (!_publishersByRegionAndTopic.ContainsKey(activeRegion))
```
martincostello: Unrelated to logging perf improvement. Should use `TryGetValue()` here so that the call to the indexer on line 179 isn't needed.

JustSaying/JustSayingBus.cs
```diff
@@ -167,22 +167,21 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
...
-                var errorMessage = $"Error publishing message, no publishers registered for active region {activeRegion}.";
-                _log.LogError(errorMessage);
-                throw new InvalidOperationException(errorMessage);
+                _log.LogError("Error publishing message. No publishers registered for active region '{Region}'.", activeRegion);
+                throw new InvalidOperationException($"Error publishing message. No publishers registered for active region '{activeRegion}'.");
             }
 
             var topic = message.GetType().ToTopicName();
             var publishersByTopic = _publishersByRegionAndTopic[activeRegion];
             if (!publishersByTopic.ContainsKey(topic))
```
martincostello: Again, `TryGetValue()` so the indexer call on line 187 isn't needed.

JustSaying/JustSayingBus.cs
```diff
@@ -211,11 +210,13 @@ await publisher.PublishAsync(message, metadata, cancellationToken)
                 if (attemptCount >= Config.PublishFailureReAttempts)
                 {
                     Monitor.IssuePublishingMessage();
-                    _log.LogError(0, ex, $"Failed to publish message {message.GetType()}. Halting after attempt {attemptCount}");
+                    _log.LogError(0, ex, "Failed to publish a message of type '{MessageType}'. Halting after attempt number {AttemptCount}.",
```
martincostello: Would `{PublishAttemptCount}` be a better name?

JustSaying/JustSayingBus.cs
```diff
@@ -211,11 +210,13 @@ await publisher.PublishAsync(message, metadata, cancellationToken)
...
+                        message.GetType(), attemptCount);
                     throw;
                 }
 
-                _log.LogWarning(0, ex, $"Failed to publish message {message.GetType()}. Retrying after attempt {attemptCount} of {Config.PublishFailureReAttempts}");
+                _log.LogWarning(0, ex, "Failed to publish a message of type '{MessageType}'. Retrying after attempt number {AttemptCount} of {PublishFailureReAttempts}.",
```
martincostello: `Reattempt` instead of `ReAttempt`.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
AnthonySteele: a) I'll check the overloads, but it's almost certainly starting with a `0` to make it compile with version 1.0.

b) Agreed, standardize on US English for consistency :/

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
AnthonySteele: Can you verify before I remove them all?

JustSaying/JustSayingBus.cs
AnthonySteele: Probably true. A separate PR maybe?

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
martincostello: If I were to verify it it would be deleting the 0 in an IDE and seeing if it complies.

JustSaying/JustSayingBus.cs
martincostello: Only for option 1. If we just want to make it Moreno verbose might as well just change it here.

JustSaying/JustSayingBus.cs
```diff
@@ -189,6 +189,16 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
...
         }
 
+        private string GetActiveRegion()
+        {
+            if (Config.GetActiveRegion == null)
+            {
+                return Config.Regions.First();
+            }
+
+            return Config.GetActiveRegion();
```
slang25:
```csharp
=> Config.GetActiveRegion?.Invoke() ?? Config.Regions.First();
```

AnthonySteele: Hm, I could go either way on that, does anyone want to break the tie on if the terse version is more readable?

JustSaying/AwsTools/MessageHandling/SqsQueueByNameBase.cs
```diff
@@ -53,6 +53,8 @@ private static bool Matches(string queueUrl, string queueName)
             => queueUrl.Substring(queueUrl.LastIndexOf("/", StringComparison.Ordinal) + 1)
                 .Equals(queueName, StringComparison.OrdinalIgnoreCase);
 
+        private static TimeSpan CreateRetryDelay = TimeSpan.FromMinutes(1);
```
martincostello: This can be `readonly`.

JustSaying/AwsTools/MessageHandling/SqsQueueByNameBase.cs
```diff
@@ -90,12 +92,13 @@ public virtual async Task<bool> CreateAsync(SqsBasicConfiguration queueConfig, i
                 // If we're on a delete timeout, throw after 2 attempts.
                 if (attempt >= 2)
                 {
-                    _log.LogError(0, ex, "Error trying to create queue '{QueueName}'. Maximum retries exceeded for delay.", QueueName);
+                    _log.LogError(0, ex, "Error trying to create queue '{QueueName}'. Maximum retries exceeded for delay {Delay}.",
```
martincostello: Maybe we should include what the maximum number of retries is?

JustSaying/JustSayingBus.cs
```diff
@@ -158,33 +161,42 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
                 throw new InvalidOperationException(errorMessage);
             }
 
-            string activeRegion;
-            if (Config.GetActiveRegion == null)
+            var activeRegion = GetActiveRegion();
```
martincostello: I think `string` would be clearer here.

JustSaying/JustSayingBus.cs
```diff
@@ -158,33 +161,42 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
...
+            if (!string.Equals(_previousRegion, activeRegion, StringComparison.Ordinal))
             {
-                activeRegion = Config.Regions.First();
+                _log.LogInformation("Active region for publishing has been set to '{Region}', was '{PreviousRegion}'.",
```
martincostello: I think `{ActiveRegion}` would be a better name here to contrast with `{PreviousRegion}`.

JustSaying/JustSayingBus.cs
```diff
@@ -158,33 +161,42 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
                 throw new InvalidOperationException(errorMessage);
             }
 
-            string activeRegion;
-            if (Config.GetActiveRegion == null)
+            var activeRegion = GetActiveRegion();
+            if (!string.Equals(_previousRegion, activeRegion, StringComparison.Ordinal))
```
martincostello: Exclude `null` from the logging here? Is it useful to log on the first execution that it changed from `null`?

JustSaying/JustSayingBus.cs
```diff
@@ -158,33 +161,42 @@ private IMessagePublisher GetActivePublisherForMessage(Message message)
...
 
-            return publishersByTopic[topic];
+            return publisher;
+        }
+
+        private string GetActiveRegion()
+        {
+            if (Config.GetActiveRegion == null)
+            {
+                return Config.Regions.First();
```
martincostello: Is there code elsewhere that _guarantees_ this _always_ has at least one item in it? Otherwise the exception thrown would be very vague and generic.

JustSaying/JustSayingBus.cs
martincostello: In this case I prefer the original, particularly if some handling needs adding to check there's something in the region set.

JustSaying/JustSayingBus.cs
AnthonySteele: We use just `{Region}` in most places.

JustSaying/JustSayingBus.cs
AnthonySteele: It's useful to log what it starts off as. The other option is to log the `null->some` transition as a special message.

JustSaying/JustSayingBus.cs
AnthonySteele: _de gustibus non est disputandum_

JustSaying/JustSayingBus.cs
AnthonySteele: Yes, it's checked in `Validate.Validate`.

JustSaying/AwsTools/MessageHandling/SqsQueueByNameBase.cs
AnthonySteele: That literal `2` of course means that there will be 3 attempts, numbered 0,1,2

AnthonySteele: And it needs to be moved a few lines earlier if it is to do anything.

martincostello:
Please don't mark comments from review as resolved if the code hasn't actually been changed to action them. Otherwise I have to manually go through all the comments and un-collapse them to see the the replies.

AnthonySteele:
Finally approved! Further changes will have to happen in other PRs 😆

Issue #494, opened 2018-12-29
Integration Tests not working with latest goaws


slang25:
The latest version of goaws has been pushed to docker hub, it contains some really nice fixes (we might be able to enable more tests). However currently it has broken our tests, so we either need to pin the version or fix them.

See here:
https://travis-ci.org/justeat/JustSaying/jobs/473402360

I'll do some investigating.

slang25:
It seems to be picking up the `Local` environment config, not `Dev` ([see here](https://github.com/p4tin/goaws/blob/c0c3d207a7c5e8a03460e034f53bf71b91a3eaf3/app/conf/goaws.yaml#L3)), so we are getting queue urls with `...goaws.com/...` in.

martincostello:
I had this problem locally a week or so ago actually running it locally on my work laptop because Docker is messed up on it.

In my case I had to explicitly tell it were the `goaws.yaml` file was in WSL as it didn't seem to be picking the location up for itself, so then it was generating queue URIs like `http://:/foo`, and causing exceptions from `Uri.ctor()`.

martincostello:
#495 works around this by pinning the version, but we should look into what's wrong and fix it or raise an issue against goaws.

slang25:
Agreed, that was just to unblock us. I think it is finding the config but selecting the wrong env. I'll create an image with our own config and see I can get it working.  Probably tonight

martincostello:
@slang25 I guess this is done now?

slang25:
Thanks, fixed in #496

PR #495, opened 2019-01-01
Use "Logging.Abstractions" version 2 for all target frameworks

AnthonySteele:

Use `Microsoft.Extensions.Logging.Abstractions` version 2.0 for all target frameworks.

Although it may be more technically correct to use version 1.1 for the `net461` target
What we found when using this, was that other libs  (e.g. [NLog.Extensions.Logging](https://www.nuget.org/packages/NLog.Extensions.Logging/) or [Serilog.Extensions.Logging](https://www.nuget.org/packages/Serilog.Extensions.Logging/)) force the use of Logging.Abstractions version 2.0 anyway on `net461`. And reconciling this is additional hassle. It is simpler to always use  Logging.Abstractions version 2.0.

This [was mentioned in comments](https://github.com/justeat/JustSaying/pull/493#issuecomment-449739928) for #493 but decided for a separate PR.


martincostello:
Looks like the tests on Linux are falling foul of #494.

PR #496, opened 2019-01-02
Update goaws

slang25:
Bare with me, I'm doing this from a phone

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ addons:
     - libicu-dev
     - libssl-dev
     - libunwind8
+  hosts:
+    - goaws.com
 
 script:
   - ./build.sh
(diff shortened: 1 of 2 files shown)


AnthonySteele:
Still failing in IntegrationTests on travis.

slang25:
Yeah, I think the current change is needed (the loopback for goaws.com) and we'll probably need add an accountid to the config, and there may still be a change required to goaws to get it working again (possible missing route). See here:
https://github.com/p4tin/goaws/issues/187

p4tin:
I am working on a fix for goaws to make the router problem go away...  Stil doing some tests to make sure and updating the unit tests should be out with version 1.1.2 by tomorrow.

slang25:
Nice one @p4tin 😄

slang25:
Great job @p4tin, works great!

PR #499, opened 2019-01-15
Fewer exclusions in the integration test

AnthonySteele:

Removed a few project-wide code quality rule exclusions from  `JustSaying.IntegrationTests.csproj`
- some are only needed for one file that intentionally does edge-case things, so instead I do a `#pragma` there.
- some can be fixed, like the string comparison and nested message type classes.
- the rest are retained

### DIFF
--- a/JustSaying.IntegrationTests/FluentNotificationStackTestBase.cs
+++ b/JustSaying.IntegrationTests/FluentNotificationStackTestBase.cs
@@ -11,7 +11,6 @@ using NSubstitute;
 
 namespace JustSaying.IntegrationTests
 {
-#pragma warning disable CA1054
     public abstract class FluentNotificationStackTestBase : XAsyncBehaviourTest<JustSaying.JustSayingFluently>
     {
         protected RegionEndpoint Region => TestFixture.Region;
@@ -173,6 +172,8 @@ namespace JustSaying.IntegrationTests
             return (false, null);
         }
 
+// disable warning about "queueUrl" typed as string
+#pragma warning disable CA1054
         protected async Task<bool> IsQueueSubscribedToTopic(Topic topic, string queueUrl)
         {
             var request = new GetQueueAttributesRequest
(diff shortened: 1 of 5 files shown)


JustSaying.IntegrationTests/WhenRegisteringASqsSubscriber/WhenHandlingMultipleTopics.cs
```diff
@@ -34,13 +34,13 @@ public async Task SqsPolicyWithAWildcardIsApplied()
 
             policyJson.Statement.Count.ShouldBe(1,  $"Expecting 1 statement in Sqs policy but found {policyJson.Statement.Count}");
         }
+    }
 
-        public class TopicA : Message
-        {
-        }
+    public class TopicA : Message
```
martincostello: Meh. I’d prefer to suppress this and keep them nested. Otherwise they have scope everywhere else in the test suite.

JustSaying.IntegrationTests/FluentNotificationStackTestBase.cs
```diff
@@ -11,6 +11,7 @@
 
 namespace JustSaying.IntegrationTests
 {
+#pragma warning disable CA1054
```
martincostello: It would be better to put the pragmas immediately above the specific lines they apply to, and then turn them on again immediately after. Otherwise you're turning off the entire rule for all the code in the entire file.

```csharp
#pragma warning disable BadThing
// Code that isn't really bad
#pragma warning restore BadThing
```

Otherwise someone knows you're turning a rule off, but doesn't know specifically what specific code is causing the warning.

AnthonySteele: Is the "restore" needed if it comes at the bottom of the file?

martincostello: If the error is on a class being declared, then you'd do this:

```csharp
#pragma warning disable BadThing
public class Foo
#pragma warning restore BadThing
```

You wouldn't wrap the entire class, because it isn't every line that causes the warning, just the first.

PR #501, opened 2019-01-18
Use analysers version 2.6.3

AnthonySteele:

Use analysers version 2.6.3
Which came out 8 days ago ( https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/2.6.3 )
And fix the "writeable collection" warning
https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca2227-collection-properties-should-be-read-only?view=vs-2017

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -237,7 +237,7 @@ namespace JustSaying.AwsTools.MessageHandling
             return _messageProcessingStrategy.StartWorker(action, ct);
         }
 
-        public ICollection<ISubscriber> Subscribers { get; set; }
+        public ICollection<ISubscriber> Subscribers { get; }
 
         private static void DefaultErrorHandler(Exception exception, Amazon.SQS.Model.Message message)
         {
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
   <Import Project="version.props" />
   <PropertyGroup>
-    <AnalyzersPackageVersion>2.6.2</AnalyzersPackageVersion>
+    <AnalyzersPackageVersion>2.6.3</AnalyzersPackageVersion>
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysisRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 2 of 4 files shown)


JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -237,7 +237,7 @@ private Task HandleMessage(Amazon.SQS.Model.Message message, CancellationToken c
             return _messageProcessingStrategy.StartWorker(action, ct);
         }
 
-        public ICollection<ISubscriber> Subscribers { get; set; }
+        public ICollection<ISubscriber> Subscribers { get; }
```
martincostello: Do we _actually_ want to do this to all these types?

AnthonySteele: For `PublishMetadata` yes, I made that class recently and we can get by without a settor on the collection. 
For the other two interfaces 🤷‍♂️ if you wanted to keep it, shoulda put a test on it.

martincostello: Counter example: all "options" classes that use collections would violate this rule and then config binding would silently fail because the setter was no longe public.

Adding tests to test that an API surface for a simple POCO with a get and a set lets you get and set..._really_?

AnthonySteele: This is the interface `INotificationSubscriberInterrogation` - what does that really do?

martincostello: I believe it's something to allow introspection on "stuff" after the listeners are all set up.

AnthonySteele: The interface doesn't appear to be used anywhere, in internal code or on the public github https://github.com/search?q=INotificationSubscriberInterrogation&type=Code
It seems to only be used inside JustSaying

martincostello: Maybe instead we should just remove it instead?

AnthonySteele: I mean, the `get` is used via `JustSayingBus.AddNotificationSubscriber`. The set appears to be unused.

martincostello: Well as long as it's actually being initialized and populated with stuff then.

PR #502, opened 2019-01-23
remove JustBehave from BaseQueuePollingTest

AnthonySteele:

remove `JustBehave` from `BaseQueuePollingTest`.

This leaves 17 `XAsyncBehaviourTest` classes in the unit tests.


AnthonySteele:
Hm yep. That's now the JustBehave / xunit integration was doing it - running the whole thing in `InitializeAsync()`. 

However now that the middleman is gone, we can follow more conventional patterns, e.g. 

https://mderriey.com/2017/09/04/async-lifetime-with-xunit/

PR #507, opened 2019-01-24
Remove another JustBehave reference

AnthonySteele:

### DIFF
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/WhenMessageHandlingFails.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/WhenMessageHandlingFails.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Threading.Tasks;
 using Amazon.SQS.Model;
 using JustSaying.TestingFramework;
 using NSubstitute;
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/WhenMessageHandlingSucceeds.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/WhenMessageHandlingSucceeds.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Threading.Tasks;
 using Amazon.SQS.Model;
 using NSubstitute;
 using Xunit;
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/WhenMessageHandlingThrows.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/WhenMessageHandlingThrows.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Threading.Tasks;
 using Amazon.SQS.Model;
 using JustSaying.TestingFramework;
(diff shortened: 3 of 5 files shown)


martincostello:
Given these PRs are/will be basically all the same, why not just do one PR with a commit per removal to make it just one PR to review instead of 16?

AnthonySteele:
OK, leave this open and I will add to it.

AnthonySteele:
I have removed `JustBehave` from all the tests in the `JustSaying.UnitTests.AwsTools.MessageHandling.Sqs` folder. This forms a natural unit of work, so it might be worth review and merge before moving on.

This is a unit of work (and was quick to do) because I added a simple base class `WhenPublishingTestBase` which takes the place of `XAsyncBehaviourTest<SqsPublisher>` and supplies the basic mechanism that these tests use.

So let's see what we think of that before moving on and attempting similar in the remaining `10` references to `JustBehave` in unit tests.

JustSaying.UnitTests/AwsTools/MessageHandling/Sqs/WhenPublishingTestBase.cs
```diff
@@ -0,0 +1,39 @@
...
+using Amazon.SQS;
+using JustSaying.AwsTools.MessageHandling;
+using NSubstitute;
+using Xunit;
+
+namespace JustSaying.UnitTests.AwsTools.MessageHandling.Sqs
+{
+    public abstract class WhenPublishingTestBase : IAsyncLifetime
+    {
+        protected SqsPublisher SystemUnderTest;
```
martincostello: Make these properties, not fields.

AnthonySteele:
That is all of the `XAsyncBehaviourTest<T>` usages removed from unit tests.
There are still 9 usages of `XBehaviourTest<T>` in the unit tests.
I will look at these next.

AnthonySteele:
All clear in `JustSaying.UnitTests`.
There is no `JustBehave` reference to remove  from the unit tests, as it come in via `JustSaying.TestingFramework`, which both the nit tests and integration tests use.

JustSaying.UnitTests/XBehaviourTest.cs
```diff
@@ -0,0 +1,55 @@
...
+        protected void Execute()
+        {
+            Given();
+
+            try
+            {
+                SystemUnderTest = CreateSystemUnderTest();
+                When();
+            }
+            catch (Exception ex)
```
martincostello: Could simplify this (I think) to:

```csharp
catch (Exception ex) when (_recordExceptions)
{
    ThrownException = ex;
}
```

martincostello:
Unless you've already started, I'm just about to start porting the tests in the `WhenRegisteringHandlersViaResolver` folder in the integration tests.

JustSaying.UnitTests/XBehaviourTest.cs
AnthonySteele: Ah the infamous `when` keyword, as avoided in this PR!

AnthonySteele:
No, I haven't done anything with integration tests. Go ahead!

martincostello:
Changes made in #508.

AnthonySteele:
My preference is to approve and merge this one now, it's at a point where further work is a change of focus.

PR #510, opened 2019-01-27
Migrate more integration tests

martincostello:
* Migrate more integration tests to use the fluent API.
  * Add a method to `MessagingBusBuilder` to get a configured AWS client factory (rather than a default one from the `IServiceProvider`).
  * Add a convenience extension method to wrap multiple handlers as one handler.
  * Add overloads for specifying multiple regions as `RegionEndpoint` instances.
  * Change `InvalidOperationException` to `NotSupportedException` if no message handlers are configured (added in #508).
  * Add a convenience method to the base class for fluent integration tests that provides the `IServiceProvider` to the callee.

Relates to #380.


JustSaying.Extensions.DependencyInjection.Microsoft/IServiceCollectionExtensions.cs
```diff
@@ -199,6 +200,53 @@ public static IServiceCollection AddJustSaying(this IServiceCollection services,
...
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="services"/> or <paramref name="handlers"/> is <see langword="null"/>.
+        /// </exception>
+        /// <exception cref="ArgumentException">
+        /// <paramref name="handlers"/> contains no handlers.
+        /// </exception>
+        public static IServiceCollection AddJustSayingHandlers<TMessage>(
+            this IServiceCollection services,
+            IList<IHandlerAsync<TMessage>> handlers)
```
AnthonySteele: I would go with an `IEnumerable<>` here because it's the least powerful and most widely useful interface, also it's what the constructor of `ListHandler` takes as parameter.
If it's enumerated multiple times and you want to avoid that, then convert it internally  with `.ToList()`.
If that's not acceptable, then maybe `IReadOnlyCollection<>` - just not a mutable list.

JustSaying.IntegrationTests/Fluent/Subscribing/WhenHandlerIsDeclaredAsExactlyOnce.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    public class WhenHandlerIsDeclaredAsExactlyOnce : IntegrationTestBase
+    {
+        public WhenHandlerIsDeclaredAsExactlyOnce(ITestOutputHelper outputHelper)
+            : base(outputHelper)
+        {
+        }
+
+        [AwsFact]
+        public async Task Then_Both_Handlers_Receive_The_Message()
```
AnthonySteele: Is this method correctly named? IFrom the code it seems more like "then the handler is called once"

JustSaying.Extensions.DependencyInjection.Microsoft/IServiceCollectionExtensions.cs
martincostello: I originally had it as `IEnumerable<T>`, but then I changed it to `ICollection<T>` so I could see if it was empty without having to use a LINQ `Count()` on it. Then I changed it to `IList<T>` so I could use the indexer to do the optimisation.

I can change it to `IReadOnly*<T>` if the distinction is important.

JustSaying.IntegrationTests/Fluent/Subscribing/WhenHandlerIsDeclaredAsExactlyOnce.cs
martincostello: Copy-pasta.

JustSaying.Extensions.DependencyInjection.Microsoft/IServiceCollectionExtensions.cs
AnthonySteele: If you're not going to modify it, then the `IReadOnly*` indicates that intent.

Issue #506, opened 2019-01-23
Publish Retry logic review and docs


AnthonySteele:
**Is your feature request related to a problem? Please describe.**

Slow publishes and even time-outs when publishing do happen. That's an SQS issue, but the code has to deal with it. I get questions about how the publish retry logic works, and how best to set it up. I don't have clear answers about how all the parts interact.

There are publish retry settings inside the AWS SDK, and in JustSaying - do they stack? If there are 3 retries in each of those laters is that 9 tries total?
 
The caller can also pass in a `CancellationToken` to the `Publish()` to time it out after a set time (or whatever other logic they might use to invalidate the token). If they use this, should they then retry with e.g. [Polly](https://github.com/App-vNext/Polly)?

**Describe the solution you'd like**

Let's clarify and document.

Good docs, possibly even review and simplification of the relevant codes. 

And examples for what to do in cases where 
 * The message must be published if at all possible, even if it is late
 * Where the message is out of date and useless after a minute  (e.g. a location ping)

Which combination of timeout and retry is likely to give best results. e.g. given 15 second, is it better to try once with a 15 second timeout, or 3 times with a 5 second timeout each?

PR #504, opened 2019-01-23
remove XAsyncBehaviourTest in WhenThereAreExceptionsInMessageProcessing

AnthonySteele:
_Summarise the changes this Pull Request makes._

remove `XAsyncBehaviourTest` in `WhenThereAreExceptionsInMessageProcessing`

_Please include a reference to a GitHub issue if appropriate._

PR #511, opened 2019-01-29
Rename "When"

AnthonySteele:

In the unit tests, rename virtual methods called `When` to avoid `CA1716`.
private and not-virtual methods can be called "When", that's not a warning.

This is tidy-up after #504, #506, #507

### DIFF
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/BaseQueuePollingTest.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/BaseQueuePollingTest.cs
@@ -44,7 +44,7 @@ namespace JustSaying.UnitTests.AwsTools.MessageHandling.SqsNotificationListener
 
             SystemUnderTest = CreateSystemUnderTest();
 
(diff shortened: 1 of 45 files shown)


JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/BaseQueuePollingTest.cs
```diff
@@ -44,7 +44,7 @@ public virtual async Task InitializeAsync()
 
             SystemUnderTest = CreateSystemUnderTest();
 
-            await When().ConfigureAwait(false);
+            await WhenAction().ConfigureAwait(false);
```
adammorr: Nit: Do we want this to be called `WhenActionAsync` to align it with `DisposeAsync` / `InitialiseAsync`?

adammorr: Ah ok, well `WhenAsync` would work unless anyone else has any strong feelings

PR #512, opened 2019-01-31
Update integration test timeout logic.

### LINKED ISSUE #497: Update test timeout logic for integration tests?
I just stumbled across this while looking at the open PRs against AspNetCore:

https://github.com/aspnet/Extensions/blob/master/src/TestingUtils/Microsoft.AspNetCore.Testing/src/TaskExtensions.cs

Thoughts on using a similar approach rather than this?

https://github.com/justeat/JustSaying/blob/e79209608893e273256ff3630f09d0e1b02b5a40/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs#L98

martincostello:
Resolves #497.

### DIFF
--- a/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
+++ b/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using Amazon.Runtime;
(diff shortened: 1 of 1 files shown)


JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -110,11 +112,14 @@ protected async Task RunActionWithTimeout(Func<CancellationToken, Task> action)
 
                 if (resultTask == delayTask)
                 {
-                    throw new TimeoutException($"The bus related action took longer to execute than the timeout of {Timeout.TotalSeconds} seconds");
+                    throw new TimeoutException($"The tested action took longer than the timeout of {Timeout} to complete.");
+                }
+                else
+                {
+                    cts.Cancel();
```
adammorr: Out of interest, is there a reason we have the else condition when we throw on the first one being true?

martincostello: I didn't go to the talk, so I don't actually know. I was just copying the sample. Before I opened the code I didn't realise it was 99% the same as what was in the sample already anyway.

martincostello: No braces? 🤢

slang25: This is the hill I will die on 😆

PR #518, opened 2019-02-17
[WIP] Unit tests with the new fluent api

AnthonySteele:

Work towards deprecating the old fluent api and using the new one instead. Trying to port a unit test by building a base class `UnitTestBase` to replace `JustSayingFluentlyTestBase`.

I found that the `IVerifyAmazonQueues` instance needs to be injectable.

The tests in `WhenAddingASubscriptionHandler` that are commented out depend on verifying against a mock `IAmJustSaying Bus`. 

What equivalent should be mocked, if any? Do these tests still make sense? 


martincostello:
@AnthonySteele Are you going to come back to this PR? If not we can close it, then I'll do a PR to shift the code around into `src` and `tests` before we do any more.

Issue #471, opened 2018-12-13
Remove/deprecate CreateMeABus and JustSayingFluently*


martincostello:
Now that #431 has been merged which provides the basis for a new _"builder API"_ for using JustSaying, which should start work on slowly deprecating the existing [`CreatMeABus`](https://github.com/justeat/JustSaying/blob/master/JustSaying/CreateMeABus.cs) class in `master` so that by the time we release `7.0.0` it no longer exists, as well as removing the existing `JustSayingFluently*` classes.

martincostello:
Unless there are compelling reasons not to do so, the new API should have feature parity with the `6.x.x` releases in the new API surface.

Feature parity in this sense means that you can achieve a comparable result, but it does not necessarily need to be like-for-like in terms of it being a default behaviour or not, naming or location in the API.

gkinsman:
@martincostello for V7, can we perhaps scope this issue to removing CreateMeABus while leaving JustSayingFluently? The new fluent API depends on it so we'd need to pass those calls straight through to the JustSayingBus if we wanted to rip out fluently too. 

JustSayingStack bypasses it completely so we don't need to think about it much. We could also just make JustSayingFluently internal to hide it?

martincostello:
The original idea was to implement it on top of the old one just to get the new API in, and then after that hoist the implementation up (or move it) so that CMAB and JSF could be deleted.

If we want to not wait to do that for v7, I'd be happy to settle for making all of it internal so that it's effectively removed and then can be chipped-away at separately.

gkinsman:
That sounds fair - I'm tempted to spike out removing it in the case that it is just a matter of passing the calls through to JSB, but I suspect it might be a little trickier than that. If that fails I'll make it internal 👍

PR #521, opened 2019-02-24
Migrate tests

martincostello:
As part of #380 (and eventually #471), this PR migrates some of the "old style" integration tests to the "newer way" of doing our integration tests.

These tests still use the queue classes directly in an _integration-y_ way, but after this is merged I'll migrate them to setup via the fluent API so they're less "low-level".

The only "old" integration tests left now are the multi-region ones, which need a real AWS account to refactor them against as the emulator doesn't really support multi-region.


AnthonySteele:
Hard to review, since there are a lots of files and they don't match up - name, folder and contents have all changed so git can't tell which new one matches which old one.

martincostello:
All I can suggest is review one commit at a time - they're a rewrite, so they won't align, and the styles are fundamentally different anyway.

Alternatively, I can cherry pick each commit into a different PR each, but that's basically the same as looking at the diffs in each commit, except split over 6 PRs instead of one.

JustSaying.IntegrationTests/Fluent/AwsTools/WhenCreatingErrorQueue.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    public class WhenCreatingErrorQueue : IntegrationTestBase
+    {
+        public WhenCreatingErrorQueue(ITestOutputHelper outputHelper)
+            : base(outputHelper)
+        {
+        }
+
+        [AwsFact]
+        public async Task Then_The_Retention_Period_Stays_At_Maximum()
```
adammorr: This test name is a bit misleading as it looks like we're asserting on the retention change (14 days to 100 seconds)

martincostello: 39b923565c1d6e41d96b86e04b93150026669179

PR #525, opened 2019-02-24
Update NuGet packages

martincostello:
Update NuGet packages to their latest versions.

### DIFF
--- a/JustSaying.Tools/JustSaying.Tools.csproj
+++ b/JustSaying.Tools/JustSaying.Tools.csproj
@@ -1,6 +1,6 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
   <ItemGroup>
--- a/JustSaying.Tools/JustSaying.Tools.csproj
+++ b/JustSaying.Tools/JustSaying.Tools.csproj
@@ -1,6 +1,6 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


JustSaying.Tools/JustSaying.Tools.csproj
```diff
@@ -1,6 +1,6 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
slang25: @martincostello Locally I get a nuget warning about `Magnum` being restored using full framework, which turns into an error due to `TreatWarningsAsErrors`. I see that CI is passing and am a little confused.

slang25: Also, this project is using the async methods in the AWS SDK, switching TFMs will require some small code changes to build as far as I can see. This project is probably superseeded internally at JustEat with JustSayingWeb (or whatever these days).

martincostello: Blargh - I'll look at it later.

AnthonySteele: I'm not familiar with `Magnum` and it seems to be only used for commandline parsing in `JustSaying.Tools` and hasn't been updated in ages (much like `JustSaying.Tools` ).

Suggest that we stop using it, in favour of more  modern / better maintained approaches to same?

slang25: Agreed. I'm actually working on that now with the new command line package from Microsoft (experimental, but why not?). The only thing we lose is the interactive nature that's there currently (i.e. type "quit" to exit etc...), but I think it's a bit weird and not useful.

slang25: I think this would be a straight forwards port to the "DotnetMetal" model, see the sample here:
https://github.com/dotnet/command-line-api/blob/b20e71a404eaac6cde023706336723fff60f28ef/samples/DotnetMetal/CommandDefinitions.cs

AnthonySteele: Yep, [that's the successor that I meant](https://github.com/dotnet/command-line-api).

martincostello: For now I've just reverted this in 74e0fef167b118a5c7f73b4604d412d7bcebebe2 - we can update the TFM and fix all the associated issues another time.

martincostello:
Stand by for another push... 😄

AnthonySteele:
FYI, this arrived in the last 24 hours: https://www.nuget.org/packages/ReportGenerator/4.0.15

martincostello:
There seems to be a release every week at the moment, which is starting to get irritating tbh...

AnthonySteele:
It doesn't seem to be a big deal though: https://github.com/danielpalme/ReportGenerator/releases/tag/v4.0.15

martincostello:
I've updated ReportGenerator again, even though it's the third update to it just in this PR, and the second in less than 24 hours...

PR #532, opened 2019-03-28
Make it possible to build net461 targets on OSX/Linux by adding a net…

iancooper:
…fx.props file to instruct the build system where to find Mono.

On OSX etc. if we want to build the solution we need to tell the build system how to find Mono so as to satisfy the net461 dependency. By adding a netfx.props and linking from the project file, you can build on other platforms (works in Rider etc.)

The contents of neftx.props were written by Don Syme, not me btw

### DIFF
--- a/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
+++ b/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
@@ -1,4 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\netfx.props" />
   <PropertyGroup>
     <Description>JustSaying extensions for Microsoft.Extensions.DependencyInjection.</Description>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
+++ b/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="..\netfx.props" />
-  <PropertyGroup>
+ <Import Project="$(MSBuildThisFileDirectory)..\netfx.props" />
+   <PropertyGroup>
     <Description>JustSaying extensions for Microsoft.Extensions.DependencyInjection.</Description>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <RootNamespace>Microsoft.Extensions.DependencyInjection</RootNamespace>
--- a/JustSaying.Extensions.DependencyInjection.StructureMap/JustSaying.Extensions.DependencyInjection.StructureMap.csproj
+++ b/JustSaying.Extensions.DependencyInjection.StructureMap/JustSaying.Extensions.DependencyInjection.StructureMap.csproj
@@ -1,5 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="..\netfx.props" />
+ <Import Project="$(MSBuildThisFileDirectory)..\netfx.props" />
   <PropertyGroup>
     <Description>JustSaying extensions for StructureMap.</Description>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/JustSaying.Extensions.DependencyInjection.StructureMap/JustSaying.Extensions.DependencyInjection.StructureMap.csproj
+++ b/JustSaying.Extensions.DependencyInjection.StructureMap/JustSaying.Extensions.DependencyInjection.StructureMap.csproj
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\netfx.props" />
   <PropertyGroup>
     <Description>JustSaying extensions for StructureMap.</Description>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/JustSaying.Models/JustSaying.Models.csproj
+++ b/JustSaying.Models/JustSaying.Models.csproj
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\netfx.props" />
   <PropertyGroup>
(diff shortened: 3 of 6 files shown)


JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
```diff
@@ -1,4 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\netfx.props" />
```
martincostello: Could simplify by adding this to `Directory.Build.props`, then it's imported into everything automatically?

iancooper: OK, let me peek at that option

iancooper: Doesn't seem to work.

See: https://andrewlock.net/building-net-framework-asp-net-core-apps-on-linux-using-mono-and-the-net-cli/

martincostello: Which bit doesn't work?

iancooper: It doesn't build using the mono framework once you merge the property  group into the DirectoryBuild.Props. See the blog post, the poster there identified that they could not get merging to work. I found the same thing.

martincostello: Ah, right I see.

I thought you were just trying to get it to build the 461 target on not-Windows, rather than trying to get it to build with Mono natively. OK that makes sense then as it wouldn't know to look for the `Directory.Build.props` file to use it at all.

martincostello: Would these be better as `$(MSBuildThisFileDirectory)..\netfx.props` so it won't be affected by the current directory of the shell not being the solution root?

iancooper: Ah I wonder if the issue is MS Build versions for Mono

iancooper: Nah, it is running v15, so it ought to have that feeature. It looks like the global props are not inserted into the build early enough

slang25: This won't work due to the order msbuild processes the props stuff (it's all a bit black magic). You have to import it as initially done.

slang25: Also, could we take the approach I added to StatsD:
https://github.com/justeat/JustEat.StatsD/blob/master/ReferenceAssemblies.props

Microsoft are doing some work to make this approach a one-liner.

martincostello: I _knew_ you'd done something somewhere else. I was looking in httpclient-interception and couldn't find it 🤦‍♂️

martincostello: PR: https://github.com/justeat/JustEat.StatsD/pull/164

iancooper: @slang25 @martincostello So you went with the myget approach? I saw that here: https://stu.dev/how-to-target-net4x-on-macos/ and just opted for the other one as I was a little bit cautious of adding a package manager dependency in the file to build on an alternate platform.


Any preferences to how we do this?

iancooper: Happy to change that if we prefer this approach over the myget assemblies approach

martincostello: I'll defer to Stu, but I think I'd prefer whichever one is the smallest diff.

iancooper: LOL. How about I make your change to path, and then push this one as I think that the dependency on myget is a little worse than the 'look up mono if you are on Unix' approach. Then we can change it if we really don't like that, but I can move on to trying to add some value to JustSaying now, whilst we debate the approach :-)

slang25: I've been preferring the targeting pack from myget, only because it's closest to the solution that will land near the .NET Core 3 time frame.

However both will work, so the smallest diff is a good way to go, we could go with this as it is 👍

iancooper: @martincostello OK I updated the PR for that change

iancooper: @slang25 Yeah, it will need review when the .NET Core 3 solution arrives anyway. So...

iancooper: @slang25 Just realized I was using you blog post as the source on how to do it in the first place. LOL.

JosephWoodward: @slang25 
> I've been preferring the targeting pack from myget, only because it's closest to the solution that will land near the .NET Core 3 time frame.

Which solution are you referring to? What's happening near the .NET Core 3 time frame? (I must have missed the memo.)

JosephWoodward: Never mind, [found it](https://github.com/dotnet/designs/pull/33) :)

PR #535, opened 2019-04-06
Update the analysers

AnthonySteele:

Update the Roslyn code analysers
Remove `Text.Analyzers` as it is deprecated.

and deal with `CA1031`  - use specific exception types.
It looks like `AmazonServiceException` is a useful base class for exceptions from AWS

I'm happy to alter if people have different views on which base types should be caught where, and where to use the `#pragma` and a catch-all instead.

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
+++ b/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Amazon.Runtime;
 using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
 using JustSaying.AwsTools.QueueCreation;
@@ -73,7 +74,7 @@ namespace JustSaying.AwsTools.MessageHandling
                     MessageResponseLogger.Invoke(responseData, message);
                 }
             }
-            catch (Exception ex)
+            catch (AmazonServiceException ex)
             {
                 if (!ClientExceptionHandler(ex, message))
                 {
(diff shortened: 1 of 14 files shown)


JustSaying/AwsTools/MessageHandling/SqsPublisher.cs
```diff
@@ -53,7 +54,7 @@ public async Task PublishAsync(Message message, PublishMetadata metadata, Cancel
                     MessageResponseLogger.Invoke(responseData, message);
                 }
             }
-            catch (Exception ex)
+            catch (AmazonServiceException ex)
```
slang25: Perhaps the scope of the try should be reduced to only the aws call here. If the logger stuff throws then this is a change in behaviour, perhaps that's fine but it could be misleading in it's current state.

AnthonySteele: Good idea, done in 3 places

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -52,28 +53,12 @@ public abstract class SnsTopicBase : IMessagePublisher
         public async Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
         {
             var request = BuildPublishRequest(message, metadata);
-
+            PublishResponse response =null;
```
brainmurphy:
```suggestion
            PublishResponse response = null;
```

brainmurphy:
@AnthonySteele Looks fine, just one fussy spacing comment 😉

AnthonySteele:
@brainmurphy agreed with that spacing change. That is one that `dotnet format` can do for us now. But I would not try that while other PRs are in progress.

PR #542, opened 2019-05-04
Try new reference assemblies package

slang25:
As [discussed previously](https://github.com/justeat/JustSaying/pull/532#discussion_r270079045), this is a nice cross-plat way to acquire reference assemblies.


slang25:
I'm just checking if we should bother with the condition, doesn't matter afaict, but I want to do the simplest thing.
https://github.com/dotnet/designs/pull/33#issuecomment-489307873

AnthonySteele:
`1.0.0-preview.1` 
When does it come out of preview? Assuming that if it's around the .NET Core 3.0 time frame, that means 2019 Q3

slang25:
That is my understanding, there are some issues people are seeing, so they'll be still fixing bugs until then.

Issue #543, opened 2019-05-30
Short Circuit Retries in the case that a given message can't be handled


Yoztastic:
### Expected behaviour

In some circumstances a failure may occur that is not retry-able. It could be that the action of handling the message is sufficiently expensive that you would not wan't to process it 'configured retry' times before putting to the dead letter queue. Especially as it is non-transient error that will be guaranteed to fail every time.

### Actual behaviour

It seems currently that you can configure a IMessageBackoffStrategy but this can only affect the time it takes to re-queue the message. It would be useful to extend this to allow certain breeds of Exception to move the message directly to the Dead-Letter queue.

Is there something I missed?

Yoztastic:
As a minimal change to achieve this I was thinking add 
`public interface INonRetryable { }`
which could be implemented by any specifically non-retryable exceptions that are thrown in the handler

Then some minor changes to the MessageDispatcher like

![image](https://user-images.githubusercontent.com/5213476/58642075-70f13d80-82f4-11e9-909c-5127fb963a93.png)

should be benign for anything not extending `INonRetryable`.

There is a small chance that the message will be added to the error queue but subsequent delete fails which could lead to duplicates in the error queue. I don't think there is a compensating action as the ErrorQueue message can not be deleted without a receiptHandle.

PR #544, opened 2019-05-31
[WIP] Short Circuit Retries on NonRetryableException

Yoztastic:
_Changes this Pull Request makes._

This PR adds an interface INonRetryable that can be extended by any Message Handling Exceptions that should not result in any retries. 
If during handling the message a non transient error is detected the message will be moved to the error queue if it exists on the message dispatchers queue.
The message will be deleted from the default queue if there is no error queue or if the response to moving to the error queue is successful. 
If there is an error queue and the re-queue is not successful the message is allowed to either return to the default queue or be driven to the error queue as part of the re-drive behaviour. This is to try and and ensure that as far as possible messages are not dropped. 

There are tests to support the behaviour. 

_reference to a GitHub issue_
https://github.com/justeat/JustSaying/issues/543

Tests requiring AWS simulator URL not run locally.

### DIFF
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
@@ -264,7 +264,7 @@ namespace JustSaying.UnitTests.AwsTools.MessageHandling.MessageDispatcherTests
         {
         }
 
-        public NonRetryableException()
+        public NonRetryableException() : base("msg")
         {
         }
     }
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
@@ -1,11 +1,13 @@
 using System;
 using System.Globalization;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Amazon;
 using Amazon.Runtime;
 using Amazon.SQS;
 using Amazon.SQS.Model;
+using JustSaying.AwsTools;
 using JustSaying.AwsTools.MessageHandling;
 using JustSaying.Messaging.MessageHandling;
 using JustSaying.Messaging.MessageProcessingStrategies;
@@ -13,6 +15,7 @@ using JustSaying.Messaging.MessageSerialization;
 using JustSaying.Messaging.Monitoring;
 using JustSaying.TestingFramework;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Newtonsoft.Json;
 using NSubstitute;
 using NSubstitute.ExceptionExtensions;
@@ -24,9 +27,24 @@ namespace JustSaying.UnitTests.AwsTools.MessageHandling.MessageDispatcherTests
 {
     public class DummySqsQueue : SqsQueueBase
     {
+        public class DummyErrorQueue : ErrorQueue
+        {
+            public DummyErrorQueue(RegionEndpoint region, string sourceQueueName, IAmazonSQS client, ILoggerFactory loggerFactory, Uri uri) : base(region, sourceQueueName, client, loggerFactory)
+            {
+                Uri = uri;
+            }
+            public override Task<bool> ExistsAsync() => Task.FromResult(true);
+        }
+
         public DummySqsQueue(Uri uri, IAmazonSQS client) : base(RegionEndpoint.EUWest1, client)
         {
             Uri = uri;
+        }
+
+        public void SetErrorQueue()
+        {
+            var errorQueue = new DummyErrorQueue(Region, QueueName, Client, new NullLoggerFactory(), new Uri(this.Uri, "error"));
+            ErrorQueue = errorQueue;
         }
 
         public override Task<bool> ExistsAsync() => Task.FromResult(true);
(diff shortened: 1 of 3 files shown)


JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
```diff
@@ -167,4 +253,19 @@ public void ShouldLogException()
...
+    {
+        public NonRetryableException(string message) : base(message)
+        {
+        }
+
+        public NonRetryableException(string message, Exception innerException) : base(message, innerException)
+        {
+        }
+
+        public NonRetryableException()
```
martincostello: Should provide a default message for this exception by calling `: base("some message")`

JustSaying/AwsTools/MessageHandling/INonRetryable.cs
```diff
@@ -0,0 +1,6 @@
+namespace JustSaying.AwsTools.MessageHandling
+{
+#pragma warning disable CA1040
+    public interface INonRetryable { }
```
martincostello: Could we use an attribute rather than using marker interfaces (which is what this suppressed warning is trying to coerce the design towards)?

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -98,6 +99,10 @@ public async Task DispatchMessage(SQSMessage message, CancellationToken cancella
             catch (Exception ex)
 #pragma warning restore CA1031
             {
+                if (!handlingSucceeded && ex is INonRetryable)
+                {
+                    await MoveMessageToErrorQueue(message).ConfigureAwait(false);
+                }
```
martincostello: Nit: newline after this block.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -112,7 +117,7 @@ public async Task DispatchMessage(SQSMessage message, CancellationToken cancella
             }
             finally
             {
-                if (!handlingSucceeded && _messageBackoffStrategy != null)
+                if (!handlingSucceeded && _messageBackoffStrategy != null && !(lastException is INonRetryable))
```
martincostello: Maybe an alternative to this would be some sort of strategy that people could implement to determine whether something is retryable or not? That would allow exceptions for types people don't own (or throw) in their own code to be made non-retryable.

Using an interface and/or attributes to mark makes this less generally useful as it's all compile-time based to extend, rather than something that can be controlled at runtime.

For example:

```csharp
public interface IRetryStrategy
{
    bool CanRetry(Exception exception);
}

IRetryStrategy retryStrategy = ...;
if (!handlingSucceeded && _messageBackoffStrategy != null && !retryStrategy.CanRetry(lastException))
{
    // ...
}
```

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -141,6 +146,18 @@ private async Task<bool> CallMessageHandler(Message message)
 
             return handlerSucceeded;
         }
+        private async Task MoveMessageToErrorQueue(SQSMessage message)
```
martincostello: Nit: suffix method name with `Async`.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -186,4 +203,5 @@ private static bool TryGetApproxReceiveCount(IDictionary<string, string> attribu
             return attributes.TryGetValue(MessageSystemAttributeName.ApproximateReceiveCount, out string rawApproxReceiveCount) && int.TryParse(rawApproxReceiveCount, out approxReceiveCount);
         }
     }
+
```
martincostello: Nit: remove.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
Yoztastic: Did think about this and thought that this would be more invasive. It would require to deliver the strategy to the dispatcher and looking at other activity this would cross cut some other IoC and deprecation stories in play. 

I think that in Handler code it is not much effort to handle any non-owned exception and throw one that derives this interface this would normally be more explicit. 

Not sure I follow the distinction between compile and run-time specifically how this effects 'usefulness'. I guess it does deliver some benefits if the strategy gets more context than just the exception, but that does trade flexibly for complexity

JustSaying/AwsTools/MessageHandling/INonRetryable.cs
Yoztastic: This would be considerably more convoluted from a reflection perspective. 
I guess it would be a performance penalty only for the failed messages. But could make an impact in high rate noisy (i.e. high rate of retry) environment where handling the message is fast. There are many reasons why you may wish to avoid marker interfaces not sure this is one of them. 
Personally I think it is this or the strategy you mention below. Checking the types custom attributes for a specific attribute in this application feels a bit expensive and not knowing the life time or scope of the dispatcher means it would be hard and of diminishing returns to cache the cost of checking for attributes to improve performance.

JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
Yoztastic: Happy to do this but it is only for the Test.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
Yoztastic: Yep

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
Yoztastic: agree

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
Yoztastic: yep

Yoztastic:
Is the integration test failure down to this PR? seems odd that I can influence an issue with multiple regions without the tests specifically throwing and exception that derives from INonRetryable. Have you any docs on running the AWS simulator URL? So I can run locally?

Yoztastic:
Looks like the integration test was flaky. As it is 'green' now.
The coverage has passed but it has highlighted some partial coverage that I would prefer to cover with more specific tests.
I will wait to see how discussions regarding the preferred architecture progress before doing this.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
martincostello: So to take an extreme example to illustrate the point, if I wanted to not retry on `OutOfMemoryException`, that exception type lives in the framework, so I can't change it to add an attribute or interface, and it's thrown by the runtime, so I can't change how and where it's thrown either.

In the case of using the strategy instead of an interface/attribute, it's possible for an integrator to easily specify this behaviour in their implementation of `IRetryStrategy`.

Yoztastic: Possibly. And I realise this is an extreme case but, if you can't handle `OutOfMemoryException` in order to throw your own exception, it stands a chance that '**all bets are off**' so your `IRetryStrategy` would not get a look in either. Also I would not condone short circuit for such a Exception as the instance that is OOM is toast but other instances will processes this message. 

In fact I think that it is clearer to handle exceptions in the Handler explicitly and re-throw a purposed error rather than bubble up system exceptions and decide elsewhere what may have caused them and whether or not to defeat the re-drive policy. 

if 

> I can't change how and where it's thrown either

then you should probably let the SQS redrive do its thing. 

So even if the `IRetryStrategy` existed I would still be catching exceptions and re-throwing as specific exception. Then handling the specific exception in the implementation of `IRetryStrategy`. 

I'm not denigrating the idea of a factory, but it would required changes to many classes to deliver it to the `MessageDespatcher` that would involve changes `IMayWantOptionalSettings`, `SqsNotificationListener`,`JustSayingFluently` etc and it looked like there is WIP on these in other tickets. 

I get that once published the Interface would always have to work so if we find a use case later that is not supported it will be a PITA. But it also adds more complexity to the consumer. Normally I would say 'take the first bullet' and see if a real use case presents that a Boolean retry or not does not satisfy, the strategy could be added later if more complex scenarios present. But I understand it is very much your call so I will go with `IRetryStrategy` if you require. Might take a little longer to get back on this as I guess it will fan out the points of change.

martincostello: I just picked OOM as an example of something built in you can't change, not the OOM condition itself.

Regarding the number of changes, we're currently midway doing lots of refactoring and redesign for v7, so wouldn't necessarily be a problem on that front as we're already making breaking changes and this PR wouldn't be part of any 6.x release.

Yoztastic: Yep sure, I appreciate OOM was just an example. 

The point I was trying to elaborate was I don't think it would be a good idea to handle **any** system errors anyway. There is a lot of opportunity to think that we do know where the exception came from and assume that makes the message non-retryable. So IMHO the strategy would be more complex to the consumer and be likely to lead to inadvertent dropping of messages to the error queue as opposed to explicitly having to consciously emit a specific exception in the handler in response a specific exception at a known place. 

So perhaps a compromise would be to go with the `IRetryStrategy` but provide a default that does what I have now?

martincostello: Sorry for the delayed response, yes I think a default implementation that does that would be a reasonable compromise.

martincostello:
Resolves #543.

martincostello:
Circling back around to this, it looks like the comments _seem_ to have agreed on extending the marker interface to be a way of providing a callback to see if an exception type can be retried. That then removes the need to use attributes, but allows things to be more configurable via the user-provider strategy implementation.

Do you intend to come back to this PR to update it @Yoztastic?

Yoztastic:
I would like to come back to this. But it is hard to schedule it in. I found a way round this lack of functionality using the ErrorHandler to move these errors to the Error queue but would be better to follow the approach discussed here. Since it have been a while don't mind if this get closed, as I think what we now propose is significantly different from what I have here so no loss to start again when I get the chance.

PR #546, opened 2019-06-20
Improve throttled processing strategy

martincostello:
_**TL;DR** In high throughput scenarios, the `Throttled` class' implementation causes `Task` backlogs that consume CPU and memory that cannot be recovered when message rates drop off._

This PR attempts to improve the behaviour of the `Throttled` implementation to use the async methods on the `SemaphoreSlim` class rather than wrap wait handles in tasks and manually queue to the thread pool.

Also makes changes to consider the fact that code using the processing strategy can race and think there are available workers but then there not be any available when it attempts to process the message.

This last point in particular needs some thought and discussion, as the implementation seems quite baked into that fact, so seems to just be asking for race conditions in retrospect.

The implementation does not appear to take into account that reads of the "free workers" value is non-atomic with respect to the time it takes to request messages from SQS and start to process them.

## Details

These changes stem from a production issue where a worker that processes a single SQS queue was migrated from .NET Framework 4.7.2 to .NET Core 2.2.5. This change was canary'd on a single EC2 instance for ~24 hours on a queue that has a _particularly_ high message throughput rate.

Below are the CPU and memory consumption for a 24 hour period where the code using JustSaying 6.0.1 is in use:

image

As can be seen, there is a slow memory leak in the service, as well as an increasing CPU usage from approximately 5pm which climbs up to ~50% by 9pm which never recovers even when messages being delivered drops off to near zero, as shown below.

image

There was otherwise no obvious signs of the worker being in distress, apart from the memory and CPU. The mean time to process messages in the canary was actually more consistent and messages where being handled faster than the .NET Framework version:

image

Capturing a process dump from the canary showed that there were _hundreds_ of tasks scheduled to be run with various stacks and start points, suggesting the thread pool was reaching its limits.

Some sniffing around in the code lead to looking at the `Throttled` class for performance bottlenecks, as the application is small and only consumes one queue with a single handler and makes a HTTP call to an external service based on the contents of the message, so is quite simple and lightweight.

Digging into the history leads to some refactoring in #440, but that was merged to master after 6.0.1 was released to NuGet, with no significant changes in the rest of the commit history (as far back as 5th April 2017, it's hard to look before that due to renaming).

My working hunch is that the processing rate improved to a degree that the throttling became the bottleneck, and tasks to process messages were backing up due to the indefinite wait for the semaphore here:

https://github.com/justeat/JustSaying/blob/88f241c8a6bcb5768e2d57fae4aef906aacad53a/JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs#L26

The attached refactoring is based on changes made directly to the application yesterday which was tested overnight last night to test the hypothesis (as well as increase the throttle factor from 8 to 16), which showed the same throughput improvements but without the CPU or memory issues. In fact, the CPU used is now less that the .NET Framework version in general.

Below are the comparison charts to the above with the fix. The CPU is the lowest orange line (look near 8pm) and the memory is the lowest yellow line, showing that while slightly more memory is used, there's no leak.

image

image

### DIFF
--- a/JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
+++ b/JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using JustSaying.Messaging.MessageProcessingStrategies;
 using JustSaying.Messaging.Monitoring;
+using Microsoft.Extensions.Logging;
 using NSubstitute;
 using Shouldly;
 using Xunit;
@@ -26,25 +27,27 @@ namespace JustSaying.UnitTests.Messaging.MessageProcessingStrategies
 
     public class MessageLoopTests
     {
-        private const int MinTaskDuration = 100;
+        private static readonly TimeSpan MinTaskDuration = TimeSpan.FromMilliseconds(100);
         private const int TaskDurationVariance = 20;
 
-        private const int ConcurrencyLevel = 20;
         private const int MaxAmazonBatchSize = 10;
+        private static readonly TimeSpan StartTimeout = Timeout.InfiniteTimeSpan;
 
         [Theory]
-        [InlineData(1)]
-        [InlineData(2)]
-        [InlineData(10)]
-        [InlineData(20)]
-        public async Task SimulatedListenLoop_ProcessedAllMessages(int numberOfMessagesToProcess)
+        [InlineData(1, 20)]
+        [InlineData(2, 20)]
+        [InlineData(10, 20)]
+        [InlineData(20, 20)]
+        public async Task SimulatedListenLoop_ProcessedAllMessages(
+            int numberOfMessagesToProcess,
+            int maxWorkers)
         {
             var fakeMonitor = Substitute.For<IMessageMonitor>();
-            var messageProcessingStrategy = new Throttled(ConcurrencyLevel, fakeMonitor);
+            var logger = Substitute.For<ILogger>();
+            var messageProcessingStrategy = new Throttled(maxWorkers, StartTimeout, fakeMonitor, logger);
             var counter = new ThreadSafeCounter();
 
-            var watch = new Stopwatch();
-            watch.Start();
+            var watch = Stopwatch.StartNew();
 
             var actions = BuildFakeIncomingMessages(numberOfMessagesToProcess, counter);
             await ListenLoopExecuted(actions, messageProcessingStrategy);
@@ -69,12 +72,30 @@ namespace JustSaying.UnitTests.Messaging.MessageProcessingStrategies
             messageCount.ShouldBeGreaterThan(capacity, "To cause throttling, message count must be over capacity");
 
             var fakeMonitor = Substitute.For<IMessageMonitor>();
-            var messageProcessingStrategy = new Throttled(capacity, fakeMonitor);
+            var logger = Substitute.For<ILogger>();
+            var messageProcessingStrategy = new Throttled(capacity, TimeSpan.FromTicks(1), fakeMonitor, logger);
             var counter = new ThreadSafeCounter();
+            var tcs = new TaskCompletionSource<bool>();
 
-            var actions = BuildFakeIncomingMessages(messageCount, counter);
+            for (int i = 0; i < capacity; i++)
+            {
+                (await messageProcessingStrategy.StartWorkerAsync(
+                    async () => await tcs.Task,
+                    CancellationToken.None)).ShouldBeTrue();
+            }
 
-            await ListenLoopExecuted(actions, messageProcessingStrategy);
+            messageProcessingStrategy.AvailableWorkers.ShouldBe(0);
+
+            for (int i = 0; i < messageCount - capacity; i++)
+            {
+                (await messageProcessingStrategy.StartWorkerAsync(() => Task.CompletedTask, CancellationToken.None)).ShouldBeFalse();
+            }
+
+            messageProcessingStrategy.AvailableWorkers.ShouldBe(0);
+
+            tcs.SetResult(true);
+
+            (await messageProcessingStrategy.WaitForAvailableWorkerAsync()).ShouldBeGreaterThan(0);
 
             fakeMonitor.Received().IncrementThrottlingStatistic();
             fakeMonitor.Received().HandleThrottlingTime(Arg.Any<TimeSpan>());
@@ -85,7 +106,10 @@ namespace JustSaying.UnitTests.Messaging.MessageProcessingStrategies
         [InlineData(1, 2)]
         [InlineData(2, 2)]
         [InlineData(5, 10)]
+        [InlineData(9, 10)]
+        [InlineData(10, 10)]
         [InlineData(10, 50)]
+        [InlineData(49, 50)]
         [InlineData(50, 50)]
         public async Task SimulatedListenLoop_WhenThrottlingDoesNotOccur_DoNotCallMessageMonitor(int messageCount, int capacity)
         {
@@ -93,44 +117,54 @@ namespace JustSaying.UnitTests.Messaging.MessageProcessingStrategies
                 "To avoid throttling, message count must be not be over capacity");
 
             var fakeMonitor = Substitute.For<IMessageMonitor>();
-            var messageProcessingStrategy = new Throttled(capacity, fakeMonitor);
+            var logger = Substitute.For<ILogger>();
+            var messageProcessingStrategy = new Throttled(capacity, Timeout.InfiniteTimeSpan, fakeMonitor, logger);
             var counter = new ThreadSafeCounter();
 
             var actions = BuildFakeIncomingMessages(messageCount, counter);
 
             await ListenLoopExecuted(actions, messageProcessingStrategy);
 
-            fakeMonitor.DidNotReceive().IncrementThrottlingStatistic();
+            if (messageCount == capacity)
+            {
+                // Throttles once in call to WaitForAvailableWorkerAsync
+                fakeMonitor.Received(1).IncrementThrottlingStatistic();
+            }
+            else
+            {
+                fakeMonitor.DidNotReceive().IncrementThrottlingStatistic();
+            }
         }
 
-        private static async Task ListenLoopExecuted(Queue<Func<Task>> actions,
+        private static async Task ListenLoopExecuted(
+            Queue<Func<Task>> actions,
             IMessageProcessingStrategy messageProcessingStrategy)
         {
             var initalActionCount = (double)actions.Count;
-            var timeoutSeconds = MinTaskDuration + (initalActionCount / 100);
-            var timeout = TimeSpan.FromSeconds(timeoutSeconds);
+            var timeout = MinTaskDuration + TimeSpan.FromMilliseconds(initalActionCount / 100);
             var stopwatch = Stopwatch.StartNew();
 
             while (actions.Any())
             {
                 var batch = GetFromFakeSnsQueue(actions, messageProcessingStrategy.AvailableWorkers);
 
-                foreach (var action in batch)
-                {
-                    await messageProcessingStrategy.StartWorker(action, CancellationToken.None);
-                }
-
-                if (!actions.Any())
+                if (batch.Count < 1)
                 {
                     break;
                 }
 
+                foreach (var action in batch)
+                {
+                    (await messageProcessingStrategy.StartWorkerAsync(action, CancellationToken.None)).ShouldBeTrue();
+                }
+
                 messageProcessingStrategy.AvailableWorkers.ShouldBeGreaterThanOrEqualTo(0);
-                await messageProcessingStrategy.WaitForAvailableWorkers();
+
+                (await messageProcessingStrategy.WaitForAvailableWorkerAsync()).ShouldBeGreaterThan(0);
                 messageProcessingStrategy.AvailableWorkers.ShouldBeGreaterThan(0);
 
-                stopwatch.Elapsed.ShouldBeLessThanOrEqualTo(timeout,
-                    $"ListenLoopExecuted took longer than timeout of {timeoutSeconds}s, with {actions.Count} of {initalActionCount} messages remaining");
+                stopwatch.Elapsed.ShouldBeLessThanOrEqualTo((timeout * 3),
+                    $"ListenLoopExecuted took longer than timeout of {timeout}, with {actions.Count} of {initalActionCount} messages remaining.");
             }
         }
 
@@ -154,7 +188,7 @@ namespace JustSaying.UnitTests.Messaging.MessageProcessingStrategies
             var actions = new Queue<Func<Task>>();
             for (var i = 0; i != numberOfMessagesToCreate; i++)
             {
-                var duration = MinTaskDuration + random.Next(TaskDurationVariance);
+                var duration = MinTaskDuration + TimeSpan.FromMilliseconds(random.Next(TaskDurationVariance));
 
                 var action = new Func<Task>(async () =>
                     {
--- a/JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
+++ b/JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
@@ -38,13 +38,53 @@ namespace JustSaying.UnitTests.Messaging.MessageProcessingStrategies
         [InlineData(2, 20)]
         [InlineData(10, 20)]
         [InlineData(20, 20)]
-        public async Task SimulatedListenLoop_ProcessedAllMessages(
+        public async Task SimulatedListenLoop_ProcessedAllMessages_With_Thread_Pool(
             int numberOfMessagesToProcess,
-            int maxWorkers)
+            int concurrency)
         {
-            var fakeMonitor = Substitute.For<IMessageMonitor>();
-            var logger = Substitute.For<ILogger>();
-            var messageProcessingStrategy = new Throttled(maxWorkers, StartTimeout, fakeMonitor, logger);
+            var options = new ThrottledOptions()
+            {
+                MaxConcurrency = concurrency,
+                Logger = Substitute.For<ILogger>(),
+                MessageMonitor = Substitute.For<IMessageMonitor>(),
+                StartTimeout = StartTimeout,
+                UseThreadPool = true,
+            };
+
+            var messageProcessingStrategy = new Throttled(options);
+            var counter = new ThreadSafeCounter();
+
+            var stopwatch = Stopwatch.StartNew();
+
+            var actions = BuildFakeIncomingMessages(numberOfMessagesToProcess, counter);
+            await ListenLoopExecuted(actions, messageProcessingStrategy);
+
+            stopwatch.Stop();
+
+            await Task.Delay(2000);
+
+            counter.Count.ShouldBe(numberOfMessagesToProcess);
+        }
+
+        [Theory]
+        [InlineData(1, 20)]
+        [InlineData(2, 20)]
+        [InlineData(10, 20)]
+        [InlineData(20, 20)]
+        public async Task SimulatedListenLoop_ProcessedAllMessages_Without_Thread_Pool(
+            int numberOfMessagesToProcess,
+            int concurrency)
+        {
+            var options = new ThrottledOptions()
+            {
+                MaxConcurrency = concurrency,
+                Logger = Substitute.For<ILogger>(),
+                MessageMonitor = Substitute.For<IMessageMonitor>(),
+                StartTimeout = StartTimeout,
+                UseThreadPool = false,
+            };
+
+            var messageProcessingStrategy = new Throttled(options);
             var counter = new ThreadSafeCounter();
 
             var watch = Stopwatch.StartNew();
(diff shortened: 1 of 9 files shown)


JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
```diff
@@ -85,52 +106,65 @@ public async Task SimulatedListenLoop_WhenThrottlingOccurs_CallsMessageMonitor(i
...
+            var logger = Substitute.For<ILogger>();
+            var messageProcessingStrategy = new Throttled(capacity, Timeout.InfiniteTimeSpan, fakeMonitor, logger);
             var counter = new ThreadSafeCounter();
 
             var actions = BuildFakeIncomingMessages(messageCount, counter);
 
             await ListenLoopExecuted(actions, messageProcessingStrategy);
 
-            fakeMonitor.DidNotReceive().IncrementThrottlingStatistic();
+            if (messageCount == capacity)
```
maurofranchi: Can you please separate the tests where the `messageCount == capacity` into another test method so you can avoid the `if` and make what you want to assert more clear?

JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
```diff
@@ -85,52 +106,65 @@ public async Task SimulatedListenLoop_WhenThrottlingOccurs_CallsMessageMonitor(i
...
 
                 messageProcessingStrategy.AvailableWorkers.ShouldBeGreaterThanOrEqualTo(0);
-                await messageProcessingStrategy.WaitForAvailableWorkers();
+
+                (await messageProcessingStrategy.WaitForAvailableWorkerAsync()).ShouldBeGreaterThan(0);
                 messageProcessingStrategy.AvailableWorkers.ShouldBeGreaterThan(0);
 
-                stopwatch.Elapsed.ShouldBeLessThanOrEqualTo(timeout,
-                    $"ListenLoopExecuted took longer than timeout of {timeoutSeconds}s, with {actions.Count} of {initalActionCount} messages remaining");
+                stopwatch.Elapsed.ShouldBeLessThanOrEqualTo((timeout * 3),
```
maurofranchi: where does the `* 3` come from?

JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
martincostello: Behaviour needs discussion overall, for now I just wanted it passing.

JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
martincostello: Threading means things don't take exact amounts of time and pending an agreement on exactly how this should work, this was a big enough number to get the time within enough of some bounds to get the test passing pending a proper discussion on how this should work because the overall approach to the message processing seems a bit flawed to me.

maurofranchi: Ok I suspected it was just to make it as big as possible to make it pass. Will wait for the refactoring (eventually) after the discussion.

iancooper:
Intriguingly I just saw a talk from Daniel Marbach of Particular (NServiceBus) about writing an async message handler, where he called out the need to apply back pressure, and used a SemaphoreSlim to control the number of in-flight requests.
The video will be up shortly I suspect, though I don't think that you will get much beyond confirmation of approach.


    On Thursday, 20 June 2019, 18:55:39 CEST, Martin Costello <notifications@github.com> wrote:  
 
 
TL;DR In high throughput scenarios, the Throttled class' implementation causes Task backlogs that consume CPU and memory that cannot be recovered when message rates drop off.

This PR attempts to improve the behaviour of the Throttled implementation to use the async methods on the SemaphoreSlim class rather than wrap wait handles in tasks and manually queue to the thread pool.

Also makes changes to consider the fact that code using the processing strategy can race and think there are available workers but then there not be any available when it attempts to process the message.

This last point in particular needs some thought and discussion, as the implementation seems quite baked into that fact, so seems to just be asking for race conditions in retrospect.

The implementation does not appear to take into account that reads of the "free workers" value is non-atomic with respect to the time it takes to request messages from SQS and start to process them.

Details

These changes stem from a production issue where a worker that processes a single SQS queue was migrated from .NET Framework 4.7.2 to .NET Core 2.2.5. This change was canary'd on a single EC2 instance for ~24 hours on a queue that has a particularly high message throughput rate.

Below are the CPU and memory consumption for a 24 hour period where the code using JustSaying 6.0.1 is in use:



As can be seen, there is a slow memory leak in the service, as well as an increasing CPU usage from approximately 5pm which climbs up to ~50% by 9pm which never recovers even when messages being delivered drops off to near zero, as shown below.



There was otherwise no obvious signs of the worker being in distress, apart from the memory and CPU. The mean time to process messages in the canary was actually more consistent and messages where being handled faster than the .NET Framework version:



Capturing a process dump from the canary showed that there were hundreds of tasks scheduled to be run with various stacks and start points, suggesting the thread pool was reaching its limits.

Some sniffing around in the code lead to looking at the Throttled class for performance bottlenecks, as the application is small and only consumes one queue with a single handler and makes a HTTP call to an external service based on the contents of the message, so is quite simple and lightweight.

Digging into the history leads to some refactoring in #440, but that was merged to master after 6.0.1 was released to NuGet, with no significant changes in the rest of the commit history (as far back as 5th April 2017, it's hard to look before that due to renaming).

My working hunch is that the processing rate improved to a degree that the throttling became the bottleneck, and tasks to process messages were backing up due to the indefinite wait for the semaphore here:


  
 JustSaying/JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs 
 
 Line 26 in 88f241c 
   
|  |  await _semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);  |

 
The attached refactoring is based on changes made directly to the application yesterday which was tested overnight last night to test the hypothesis (as well as increase the throttle factor from 8 to 16), which showed the same throughput improvements but without the CPU or memory issues. In fact, the CPU used is now less that the .NET Framework version in general.

Below are the comparison charts to the above with the fix. The CPU is the lowest orange line (look near 8pm) and the memory is the lowest yellow line, showing that while slightly more memory is used, there's no leak.





You can view, comment on, or merge this pull request online at:

  https://github.com/justeat/JustSaying/pull/546

Commit Summary
   
   - Improve async behaviour in Throttled
   - Remove commented-out code

File Changes
   
   -  M JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/Support/ThrowingBeforeMessageProcessingStrategy.cs (10) 
   -  M JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/Support/ThrowingDuringMessageProcessingStrategy.cs (11) 
   -  M JustSaying.UnitTests/AwsTools/MessageHandling/SqsNotificationListener/WhenExactlyOnceIsAppliedToHandlerWithoutExplicitTimeout.cs (1) 
   -  M JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs (86) 
   -  M JustSaying.UnitTests/Messaging/MessageProcessingStrategies/ThrottledTests.cs (35) 
   -  M JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs (91) 
   -  M JustSaying/Messaging/MessageHandling/ExactlyOnceAttribute.cs (4) 
   -  M JustSaying/Messaging/MessageProcessingStrategies/DefaultThrottledThroughput.cs (15) 
   -  M JustSaying/Messaging/MessageProcessingStrategies/IMessageProcessingStrategy.cs (36) 
   -  M JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs (160) 

Patch Links:
   
   - https://github.com/justeat/JustSaying/pull/546.patch
   - https://github.com/justeat/JustSaying/pull/546.diff

slang25:
This is some really nice analysis 😄 
Just to add a bit of context and history, I did a quick writeup on `IMessageProcessingStrategy`, the interface is basically broken because it pretty much guarantees there will be a race condition with how it will be consumed (calling `WaitForAvailableWorkers` then getting the number available). The more you can get away from the existing shape of this interface the better!
https://github.com/justeat/JustSaying/issues/422#issue-378204771

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -216,29 +245,31 @@ private async Task<ReceiveMessageResponse> GetMessagesFromSqsQueue(string queueN
...
         }
 
-        private async Task<int> GetNumberOfMessagesToReadFromSqs()
+        private async Task<int> GetNumberOfMessagesToReadFromSqsAsync()
         {
-            var numberOfMessagesToReadFromSqs = Math.Min(_messageProcessingStrategy.AvailableWorkers, MessageConstants.MaxAmazonMessageCap);
+            // TODO This needs reviewing as there can be available workers here now, yet by the time
+            // they're actually used, there could be none. This would then create a backlog of messages
+            // to process waiting on the message throttle because all the workers are busy.
+            int numberOfMessagesToReadFromSqs = Math.Min(_messageProcessingStrategy.AvailableWorkers, MessageConstants.MaxAmazonMessageCap);
```
slang25: Any chance you can kill the `AvailableWorkers` property and always rely on `WaitForAvailableWorkerAsync` here. It could sidestep a bunch of bugs in any custom implementations.

slang25:
Another thing to make you aware of (you may have already read it) is this great read from Mac Gravell on the topic of `SemaphoreSlim` and `WaitAsync`:
https://blog.marcgravell.com/2019/02/fun-with-spiral-of-death.html
and this upcoming change to corefx:
https://github.com/dotnet/corefx/commit/ecba811b1438517ac90a957e2cfe4cef64a13861

JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -3,81 +3,157 @@
...
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <inheritdoc />
+        public virtual async Task<bool> StartWorkerAsync(Func<Task> action, CancellationToken cancellationToken)
         {
             try
             {
```
slang25: This reads so much nicer now 👍

martincostello:
Thanks @slang25 - sounds like I did understand what was going on properly then, and that it is setting itself up for a race condition.

As 7's going to be a breaking change anyway, sounds like this is something I can try and fix properly here as part of this PR then, rather than fudge it and come back to it later.

Removing it and changing the "worker" design sounds like a sensible approach to make this nicer then, in some way.

Once thing I'd considered is instead of:

  1. Get X messages
  1. Process X messages in parallel
  1. Wait for free workers
  1. Repeat

We could change it to:
  1. Create Y workers
  1. Get Z messages
  1. Each worker Y processes Z messages sequentially
  1. Repeat

The idea being it's easier to have many things doing something serially, than one thing trying to manage many parallel things.

Thoughts?

I'll also check out that blog post and the related PR later on.

martincostello:
@slang25 Sounds like the app was experiencing a variant of the resource exhaustion thing Mark Gravell’s post, as the old code was using a TCS too, except we were still processing messages so the instance wasn’t totally dead.

slang25:
The TCS code I added when adding the semaphore implementation, I wanted an async way of waiting for the semaphore value to change and then to peek at the count.
At the time I considered doing a `WaitAsync` and `Release` but worried that a change to the semaphore in between could happen, tripping things up a bit. At the time this change was to mitigate against a bug that would lose count of workers, potentially creating an entire platform outage at peak, so I was mega-nervous, and favouring correctness over perf (even though this whole interface is broken).

I've given this some more thought, and I can't think of a way this would get tripped up, so I should have gone with this approach first time around.

On your suggestion about changing the processing pattern, I think what you're suggesting is a better approach. What can happen now with a shared throttled strategy of let's say 10 workers, under queue back pressure, is that every other message fetch tries to get 1, and because of concurrency they are still fetching too many, so if you have multiple queues the fetch counts can look like:

Q1 - ... 1, 9, 1, 1, 8, ...
Q2 - ... 1, 1, 1, 6, 1, ...
Q3 - ... 1, 1, 1, 1, 1, ...

or something as random as that.

Part of me wants to smooth out these numbers by tracking a rolling average of the rate of messages being processed, this would potentially introduce some (bounded) back pressure into the service (and out SQS) which is less desirable, but if kept minimal is manageable (messages could timeout easier for example). With this suggestion, the listenloop would have a steady cadence, this can also help reduce latency during quiet periods, rather than wait 20 seconds to receive a message batch containing 1 message, you could ask for 1 message and get it as soon as it's in the queue. However, all this comes at the cost of yet more complexity, which is where I always drop the idea. I think your suggestion wins in terms of simplicity and saneness.

If TFMs were no barrier, what you are suggesting would be really nice exposed as an `IAsyncEnumerable<Message>`, people would be able to use Ix.NET to throttle or do whatever. I've persuaded Steve Gordon it's a great fit for his high-perf SQS client (see here: https://github.com/highperfcloud/AwsSqsClient/blob/e13c65862cee61523ba986cd900421691b925f77/src/Aws.Sqs.Core/LightweightMessageReader.cs#L11), this would all fit together so nicely (perhaps better suited to a new project).

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -216,29 +245,31 @@ private async Task<ReceiveMessageResponse> GetMessagesFromSqsQueue(string queueN
...
-
-                numberOfMessagesToReadFromSqs = Math.Min(_messageProcessingStrategy.AvailableWorkers, MessageConstants.MaxAmazonMessageCap);
+                int availableWorkers = await _messageProcessingStrategy.WaitForAvailableWorkerAsync().ConfigureAwait(false);
+                numberOfMessagesToReadFromSqs = Math.Min(availableWorkers, MessageConstants.MaxAmazonMessageCap);
             }
 
             if (numberOfMessagesToReadFromSqs == 0)
             {
-                throw new InvalidOperationException("Cannot determine numberOfMessagesToReadFromSqs");
+                throw new InvalidOperationException("");
```
slang25: Does there need to be a useful message? `numberOfMessagesToReadFromSqs` can still be `0` here.

iancooper:
> We could change it to:
> 
>     1. Create Y workers
> 
>     2. Get Z messages
> 
>     3. Each worker Y processes Z messages sequentially
> 
>     4. Repeat

FWIW Brighter uses the model of multiple performers, where each thread is a message pump i.e.

1. Create one or more threads. Each thread does:
   (i) Get Message
   (ii) Translate Message
  (iii) Dispatch Message

And so works serially, with parallelism and explicit option. We use long-running threads not thread pool threads, as we have a long running loop, which is an anti-pattern for the thread pool which expects thread to do short jobs and optimizes for that AFAIK.

We are bit more old school, and the loop reads from an in-memory queue, into which we can insert a quit message so that we exit after the next job is taken, instead of a cancellation token, and the queue itself can act as a shared buffer on some transports so you can fill that buffer and have multiple performers read the same pool of work (it won't work on RMQ for example, where there is a thread affinity with a channel and the delivery tag that you ack with, but does for something like SQS which is just using HTTP and doesn't have open channels etc).

Our model reflects the need to preserve ordering which requires a single peformer and let's you manage that, and is overall simpler than trying to write back pressure on top of the pool, something that is possible but having done it a few times not something I really like over using explicit threads, as they are long running.

The biggest problem I need to fix though is that we have on thread for the pump and handler, which makes it easy, you only get the next item when you finish work, but to support an async message pump (for scenarios where you don't need ordering). 

We want to support thread affinity on a call back and we can support that using our own queue to implement a sychronization context so that your callback becomes another item to execute via the loop, but that's not finished and has some tricky edge cases.

Anyway, mentioned just to compare notes on approaches, but we steered away from the thread pool right from the beginning due to less that positive experiences for this kind of scenario.

slang25:
Having a synchronous message pump makes a lot of sense, certainly better than blocking on thread pool threads as you say.

Here the pump is per queue and asynchronous, which saves us a thread per queue when idle. I'd certainly trade that for better-performing code. This PR hopefully gives us both.

Out of curiosity, are you saying that your message handler async callbacks are run on one thread? What do you want/need that for?

iancooper:
> Having a synchronous message pump makes a lot of sense, certainly better than blocking on thread pool threads as you say.
> 
> Here the pump is per queue and asynchronous, which saves us a thread per queue when idle. I'd certainly trade that for better-performing code. This PR hopefully gives us both (or at least to some extent).
> 
> Out of curiosity, are you saying that your message handler async callbacks are run on one thread? What do you want/need that for?

it's more "can be" we let you set configureawait to true so that you can have thread affinity on your callback. We have a number of users that want that, mostly for legacy reasons, and even HttpClient is not quite as thread safe as most of might like. In addition, in some cases, you might not want to be pushed back onto the thread pool for your callbacks which is where the ASP.NET went in the end, as there is no synchronisation context in ASP.NET Core. I'm not entirely sure that dropping back into the thread pool is what you always want to do, as it comes with its own gotchas.

But for the moment it is parked because it is hard to get right, and the fact that the ASP.NET team quit on making it work was not encouraging...

martincostello:
Thanks for the input everyone - when I get the chance this week I'll come back to this PR and rework it some more based on the above discussion to change to something like this for the pump:

```csharp
while (!cancellationToken.IsCancellationRequested)
{
    var messages = await GetMessagesAsync();

    foreach (var message in messages)
    {
        if (cancellationToken.IsCancellationRequested)
        {
            break;
        }

        await DoMessageStuffAsync(message);
    }
}
```

Then there just needs to be something opt-in to let you do that in parallel on several threads, which would let you do more in a worker that only processes one queue so that the app is making efficient usage of its resources (assuming multiple cores).

JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -3,81 +3,157 @@
...
-            if (_semaphore.CurrentCount != 0)
-                return;
-
-            _messageMonitor.IncrementThrottlingStatistic();
+            if (_semaphore.CurrentCount == 0)
+            {
+                _messageMonitor.IncrementThrottlingStatistic();
 
-            var watch = Stopwatch.StartNew();
-            await AsTask(_semaphore.AvailableWaitHandle).ConfigureAwait(false);
```
slang25: Ironically you only need to dispose a `SemphoreSlim` when you access the `AvailableWaitHandle` property, but `IDisposable` is a welcome addition here anyway.

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
martincostello: Done.

martincostello:
I think this is done and ready for review now.

slang25:
This is looking great, my only question is when would you want to set `UseThreadPool` to false.

To optionally have just the first synchronous part of your async method run as part of the message processing doesn't make sense to me, is it intended for a certain scenario?

martincostello:
I thought it didn’t queue anything else up further down the chain, so it would be _actually_ handling the message before the await completed. Basically turning the strategy into “Do message, do message, do message, ..., get more messages, ...” rather than “get messages, despatch messages, get more messages, ...”.

slang25:
Sorry, I shouldn't try reviewing code on my phone 😳. I missed the await of `RunActionAsync`, this makes total sense. I'll have a proper looker this evening.

JustSaying/Messaging/MessageProcessingStrategies/ThrottledOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Gets or sets the maximum number of messages to process concurrently.
+        /// </summary>
+        public int MaxConcurrency { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to process messages by dispatching
+        /// them to the thread pool or whether to process messages before returning.
+        /// </summary>
+        public bool UseThreadPool { get; set; }
```
slang25: How about:

```csharp
public ProcessingMode ProcessingMode { get; set; }

enum  ProcessingMode
{
    Sequential,
    Parallel
}
```
 to make it clear how this changes the behaviour?

martincostello: Actually that’s nicer as it’s more explicit. Will change.

martincostello: Although...that does mean an enum specific to a specific options class of a specific implementation of an interface... 🤔

martincostello: Compromised by renaming the `bool` and inverting it to mean `true` to do sequential processing (the new opt-in behaviour): fef2976c26a24c1b9c1b26a47fc02e29728a1a6e

slang25: Yeah, I went through the same thought process, I think it's worth it here. A rename would also suffice, I thought that `UseThreadPool` might not convey the behaviour (assuming people don't read the doc comments).

martincostello: I think I prefer the rename as we're not adding a whole enum with only two possible values that then we'd have to error handle any other possible value someone could fudge in.

JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
martincostello: Updated the test to not validate the timings anymore.

JustSaying.UnitTests/Messaging/MessageProcessingStrategies/MessageLoopTests.cs
martincostello: Changed to separate the two cases in later commits.

JustSaying/Messaging/MessageProcessingStrategies/Throttled.cs
```diff
@@ -3,81 +3,203 @@
...
+        /// </summary>
+        ~Throttled()
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
+        /// Gets the available number of workers.
+        /// </summary>
+        public int AvailableWorkers => _semaphore.CurrentCount;
```
slang25: Is this intentionally left in for monitoring purposes?

martincostello: I'm using it in the tests as well.

martincostello:
Thanks @slang25 😄 

Will rebase onto #550 and then merged once I've gotten another ✔️

martincostello:
Rebased.

PR #547, opened 2019-06-22
[WIP] Switch to coverlet

martincostello:
Switch to coverlet for code coverage from OpenCover.

coverlet is supported on non-Windows platforms, as well as being easier to configure.


martincostello:
Well, that's the theory...

martincostello:
For reasons I don't understand, some of the messages are being delivered more than once and breaking some of the test expectations 😕

martincostello:
All I can think of is the instrumentation slows down the tests and exposes some timing issues.

martincostello:
Bah - still timing issues in the tests making them fail.

martincostello:
Replaced by #570.

PR #549, opened 2019-06-30
Cancellable handler

martincostello:
I had an idea this morning based on the changes in #546, so I thought I'd see how far I could get with it and submit a PR for discussion.

In the event that the issue that lead to #546 occurring and tasks to process messages back up, it's possible that a task to process a message may be trying to process a message that was received so long ago that the message visibility timeout has expired and it is now a potential duplicate waiting to be processed.

If the handler had a `CancellationToken` that cancels after the _approximate_ message visibility timeout expires, JustSaying could enable two behaviours:

  1) It could self-cancel trying to do anything with the message before getting to the handler and abandoning it (potentially just extending the visibility instead, but I've not done that here).
  2) The handler, when called, could observe the cancellation token itself and make a decision about what to do with the message. It could return `false` to send it back to the queue (where it could have since been handled elsewhere anyway), or it could return `true` on the basis, if this has expired I'll just ignore it and pretend I've dealt with it.

This is implemented here by adding the `ICancellableHandlerAsync<T>` interface that derives from `IHandlerAsync<T>`, and then plumbing it in appropriately so it's used if implemented by the handler for a particular message and extending things like the exactly-once handler and stopwatch wrappers.

There's also some random bits of refactoring in here that happened as I was going along, I'll pull those out into a separate PR shortly.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,9 +5,9 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0-preview.2" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-19270-01" PrivateAssets="All" />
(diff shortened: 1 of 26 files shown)


martincostello:
Replaced by #571 as this PR appears to be broken.

PR #550, opened 2019-06-30
Fixes and refactors

martincostello:
Cherry-pick changes from #546 and #549 to:

  * Remove leftover code from previous refactors.
  * Fix integration tests failing if tasks transition so they're already completed.
  * Update various NuGet pacakges to their latest versions.
  * Change timeout constructor parameter on `ExactlyOnceHandler` to a `TimeSpan`. This makes the unit clearer and removes the need to create a new `TimeSpan` for every message invocation.
  * Compute the message lock key suffix in `ExactlyOnceHandler` once, rather than per message invocation, as it does not change for the lifetime of the handler.
  * Do a reflection call in `StopwatchHandler` to get the inner handler's type once per instance, rather than once per message.
  * Use a local function for the worker task, rather than explicitly creating a `Func<Task>`.
  * Make logger field `readonly` as its value is never reassigned.
  * Reduce the number of calls to `message.GetType()` in various places.
  * Change logger overloads to not specify an event Id of zero.
  * Include receipt handle in log message.
  * Ensure that the `MessageContext` is reset to `null` if an exception is thrown when updating the message visibility timeout.

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -237,8 +237,12 @@ namespace JustSaying.AwsTools.MessageHandling
 
         private Task HandleMessage(Amazon.SQS.Model.Message message, CancellationToken ct)
         {
-            var action = new Func<Task>(() => _messageDispatcher.DispatchMessage(message, ct));
-            return _messageProcessingStrategy.StartWorker(action, ct);
+            async Task DispatchAsync()
+            {
+                await _messageDispatcher.DispatchMessage(message, ct);
+            }
+
+            return _messageProcessingStrategy.StartWorker(DispatchAsync, ct);
         }
 
         public ICollection<ISubscriber> Subscribers { get; }
--- a/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -132,20 +132,29 @@ namespace JustSaying.AwsTools.MessageHandling
                 }
                 catch (InvalidOperationException ex)
                 {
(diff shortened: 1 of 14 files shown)


JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -237,8 +249,12 @@ private async Task<int> GetNumberOfMessagesToReadFromSqs()
 
         private Task HandleMessage(Amazon.SQS.Model.Message message, CancellationToken ct)
         {
-            var action = new Func<Task>(() => _messageDispatcher.DispatchMessage(message, ct));
-            return _messageProcessingStrategy.StartWorker(action, ct);
+            async Task DispatchAsync()
+            {
+                await _messageDispatcher.DispatchMessage(message, ct);
```
slang25: Minor - could elide the `await` here

martincostello: I figured to keep it to make debug more predictable (I’m sure there’s something in David Fowler’s async tips about that).

martincostello: [Found it](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task)

slang25: For my curiosity, is this still an allocations saving? i.e. Does a `Task` allocate more than a `Func`? (my gut says yes).

martincostello: I did this more for readability after the changes I did in #549 where I needed to make the action do more work and then it felt a bit bloated and bleurgh for a lambda.

Issue #551, opened 2019-07-03
Support for SNS SSE (Server-side Encryption)


martalexandra:
I need support for server-side encryption in SNS, so I'd like to add it.
I've been working on it, and should have a PR  to submit soon.

martincostello:
Is this similar to #363?

martalexandra:
#363 Added support for server-side encryption in the SQS queues, I need it in the SNS topic as well.
I'm using a similar approach to #363 to add this when creating a topic.

martalexandra:
PR #553 submitted

martincostello:
Resolved by #575.

PR #553, opened 2019-07-05
WIP - Issue-551 Add server-side encryption for sns topic

martalexandra:
Adds support for SNS server-side encryption by passing encryption option in config.
Follows a similar approach to #363

(https://github.com/justeat/JustSaying/issues/551)


JustSaying/JustSaying.csproj
```diff
@@ -6,9 +6,9 @@
     <ProjectReference Include="..\JustSaying.Models\JustSaying.Models.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="3.3.25.3" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.3.1.11" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.3.3.19" />
+    <PackageReference Include="AWSSDK.Core" Version="3.3.102.5" />
```
martincostello: Are these the _minimum_ versions that support SSE for SNS? We would want the version constraint to be as low as possible, rather than whatever the latest possible versions right now are.

martincostello:
I've retried the job in case it was a transient failure, but the AWS emulator we use for the Linux integration tests doesn't like these changes.

JustSaying/JustSaying.csproj
martalexandra: You're right, actually don't need to update since I changed it to use the SetTopicAttributesRequest and am not passing the attribute when creating the topic anymore, so I'm reverting to the previous versions

JustSaying/AwsTools/MessageHandling/SnsServerSideEncryption.cs
```diff
@@ -0,0 +1,15 @@
+using Newtonsoft.Json;
+
+namespace JustSaying.AwsTools.MessageHandling
+{
+    public class SnsServerSideEncryption
```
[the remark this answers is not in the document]
martalexandra: I thought about using the existing ServerSideEncryption but the KmsMasterKeyId property is being defaulted to the aws sqs default key in the constructor, which could break current usages if changed, so I created a different one for sns in the end

martalexandra:
Thanks for the review @martincostello. Regarding the tests, I've been trying to figure out the error, which from the error message seems related with credentials, though I find it strange it happens only after the topic is created, when trying to access the topic attributes. However, it's difficult to reach a conclusion, since I can't reproduce the error - I'm hitting aws when I run the tests and see it working properly. Do you have any further details about the error that can help me understand what's happening, that are not shown in the build logs? Or is there a way for me to set up the emulator so I can try to reproduce the error?

JustSaying/AwsTools/MessageHandling/SnsServerSideEncryption.cs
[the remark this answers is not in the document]
martincostello: This PR would go into version 7 which has already made breaking changes, plus the renaming of a constant in this PR is also a breaking change, so don't worry about that.

martincostello:
@martalexandra If you look at `build.sh` you'll see how the emulator Docker image is downloaded and run, you should be able to do that yourself locally.

JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -63,7 +63,12 @@ public static class JustSayingConstants
         /// <summary>
         /// Default ID of an AWS-managed customer master key (CMK) for Amazon SQS
         /// </summary>
-        public static string DefaultAttributeEncryptionKeyId => "alias/aws/sqs";
+        public static string DefaultSqsAttributeEncryptionKeyId => "alias/aws/sqs";
```
martincostello: This is the change that's breaking, but that's fine for v7.

JustSaying/JustSayingFluently.cs
```diff
@@ -115,6 +115,18 @@ public virtual INamingStrategy GetNamingStrategy()
             return this;
         }
 
+        private void CreatePublisher<T>(SnsTopicByName eventPublisher, SnsWriteConfiguration snsWriteConfig) where T : Message
```
martincostello: Is the `T` parameter actually used here?

JustSaying/AwsTools/MessageHandling/SnsServerSideEncryption.cs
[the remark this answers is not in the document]
martalexandra: Yes, that's true. Will change it then!

JustSaying/JustSayingFluently.cs
martalexandra: It's not, sorry, missed that

martalexandra:
@martincostello <https://github.com/martincostello> sorry for the delayed
response, I'm having issues trying to see my PR when logged in (issue
currently open with github), so I wasn't able to reply in github and just
now remembered I could try to reply through the email.
I've concluded the issue with my test is due to the emulator not
supporting GetTopicAttributesAsync
,and so I used the existent NotSimulatorFact for those tests.
However, that last commit is not reflecting in my PR and I've noticed that
the original repository now is showing 0 forks, and my repository says is
has been forked from https://github.com/antonjefcoate/JustSaying-1, which
also means I'm not able to raise it again in the
https://github.com/justeat/JustSaying repo. Is this due to any changes on
your side?
Thanks

martincostello:
Replaced by #575.

PR #548, opened 2019-06-23
Fake message receipt prototype

martincostello:
_This PR isn't intended to be merged, it's just for discussion._

A prototype of configuring JustSaying to receive messages from a fake in-memory SQS queue. This is an approach we're thinking about using in our team to increase the _"in-memory"_ test coverage of .NET Core-based workers.

It incorporates the changes from https://github.com/justeat/httpclient-interception/pull/108 with the AWS [`HttpClientFactory`](https://github.com/aws/aws-sdk-net/blob/189abfb1d51c5300e68c741c536f4144d5f2c4fd/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs#L52) abstract class to allow HTTP requests to SQS to be intercepted and served from in-memory resources.

The in-memory queue is managed by the `SqsMessageStore` class which handles serializing messages into the correct format required for them to accepted by JustSaying and the AWS SDK (for example, having the correct MD5 hash).

The `Can_Receive_Messages_From_Fake_Message_Pump` test in the PR then outputs something like this:

```csharp
[Fact]
public async Task Can_Receive_Messages_From_Fake_Message_Pump()
{
    // Arrange
    string queueName = Guid.NewGuid().ToString();
    string regionName = "eu-west-1";

    var credentials = new BasicAWSCredentials("accessKey", "secretKey");
    var interceptionOptions = new HttpClientInterceptorOptions().ThrowsOnMissingRegistration();

    var clientFactory = new InterceptingAwsClientFactory(credentials, interceptionOptions);
    var store = new SqsMessageStore(queueName, regionName, interceptionOptions);

    var services = new ServiceCollection()
        .AddLogging((p) => p.AddXUnit(OutputHelper).SetMinimumLevel(LogLevel.Information))
        .AddSingleton(OutputHelper)
        .AddJustSayingHandler<NumberMessage, NumberHandler>()
        .AddJustSaying(
            (builder) =>
            {
                builder.Messaging((config) => config.WithRegion(regionName))
                       .Client((client) => client.WithClientFactory(() => clientFactory))
                       .Subscriptions((subscription) => subscription.ForQueue<NumberMessage>(queueName));
            });

    IServiceProvider serviceProvider = services.BuildServiceProvider();
    IMessagingBus listener = serviceProvider.GetRequiredService<IMessagingBus>();

    var tcs = new TaskCompletionSource<bool>();
[...]

PR #555, opened 2019-07-11
Some documentation of the message JSON and the APIs which are used

brainmurphy:
Some documentation of the message JSON and the APIs which are used for publishing/subscribing. Useful for publishers/subscribers which are not using JustSaying but who wish to interoperate.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,126 @@ An example policy would look like;
 }
 ```
 
+## Message formats and AWS interactions
+
+JustSaying uses a number of Amazon Web Services APIs to transport messages between publishers and subscribers. Since these HTTP APIs can be used by applications which do not use JustSaying, it is entirely possible to have, for example, a Java application publishing a message which is subscribed to by an application using JustSaying. Equally, a C# application publishing messages using JustSaying, which are subscribed to by a Java application, is fully supported by AWS. In order to support this interoperability, it is important that the actual message formats are described. Since Amazon provide SDKs for many programming languages and frameworks, you are unlikely to interact with the APIs purely via HTTP, but knowing the structure of the JustSaying message JSON itself is useful for cross-language purposes.
+
+### Publishing to SNS
+
+JustSaying uses the SNS Publish API to send messages to SNS (documented [here](https://docs.aws.amazon.com/sns/latest/api/API_Publish.html)). The following parameters are set:
+
+#### TopicArn
+The ARN of the topic to which the message is being published.
+
+#### Subject
+JustSaying will set this to the type of the .NET message class - "OrderAccepted", for example.
+
+#### Message
+The actual message object, serialised to JSON. Because JustSaying messages must derive from the base Message class, the properties on that base class will be serialised into this JSON. On the base class, the `Id` property is set as a random GUID and the `TimeStamp` property is set to the current time (in UTC). The default serialisation settings will ignore null properties, so properties like `Conversation` will not appear in the JSON if they are left as the default null value. Enums are serialised as strings.
+
+In summary, suppose we have the following message class:
+
+````c#
+        public class OrderAccepted : Message
+        {
+            public OrderAccepted(int orderId)
+            {
+                OrderId = orderId;
+            }
+            public int OrderId { get; private set; }
+        }
+````
+
+And suppose I create a message object from that class:
+
+````c#
+        var orderAccepted = new OrderAccepted(1234)
+        {
+            RaisingComponent = "my publisher",
+        };
+````
+
+This will result in the following JSON being created and populated on that Message parameter to the SNS Publish API:
+
+````json
+{
+    "OrderId": 1234,
+    "Id": "e3f84a55-b677-43df-8c24-92c170fdd89f",
+    "TimeStamp": "2019-07-03T09:53:09.2956149Z",
+    "RaisingComponent": "my sns publisher"
+}
+````
+
+#### MessageAttributes
+Any message attributes that you have chosen to send along with the message.
+
+### Sending to SQS ("point to point")
+
+JustSaying uses the SQS SendMessage API to send messages to SQS (documented [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html)). The following parameters are set:
+
+#### QueueUrl
+The URL of the queue to which the message is being sent.
+
+#### DelaySeconds
+JustSaying will set this value to be that of the `Delay` property on the `PublishMetadata` class which is supplied as a parameter to the `PublishAsync` method.
+
+#### MessageBody
+Aa JSON object containing two properties, `Subject` and `Message`. `Subject`, as with the SNS publisher, will be the type of the .NET message class. `Message` will be the actual message object, serialised to JSON. The same serialisation applies as for the SNS publisher.
+
+As an example, if I create the same OrderAccepted object as before and send that to SQS, then the value of this parameter will be:
+
+````json
+{
+    "Subject":"OrderAccepted",
+    "Message":"{\"OrderId\":1234,\"Id\":\"e3f84a55-b677-43df-8c24-92c170fdd89f\",\"TimeStamp\":\"2019-07-03T09:53:09.2956149Z\",\"RaisingComponent\":\"my sqs publisher\"}"
+}
+````
+
+This structure is to ensure that messages that are delivered to SQS queues as a result of directly publishing to the queue have the same format as messages delivered to a queue as a result of being subscribed to an SNS topic to which a message was published. So, from the point of view of a subscriber, whether the message was originally delivered to SNS or SQS is not important.
+
+### Subscribing to SQS
+
+Whether the message was originally published to an SNS topic or sent directly to an SQS queue, the subscriber will receive it by reading from an SQS queue. To do this, JustSaying uses the SQS ReceiveMessage API to receive messages from SQS (documented [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html)). The following parameters are set in JustSaying's request to this API:
+
+#### QueueUrl
+The URL of the queue from which messages are to be received.
+
+#### MaxNumberOfMessages
+JustSaying will vary this value depending on the number of CPU cores available and throttling configuration. 
+
+#### WaitTimeSeconds
+JustSaying sets this to 20.
+
+#### AttributeNames
+JustSaying will only set ApproximateReceiveCount for this parameter.
+
+The response from the API is an object with a single `Messages` property (documented [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_Message.html)). The JSON format of each message therein is described [here](https://docs.aws.amazon.com/sns/latest/dg/sns-sqs-as-subscriber.html) and [here](https://docs.aws.amazon.com/sns/latest/dg/sns-message-and-json-formats.html#http-notification-json).
+
+As an example, the following response would be available for that same sample message if it had been published to an SNS topic and then received from an SQS queue subscribed to that topic:
+
+````json
+{
+  "ReceiveMessageResponse": {
+    "ReceiveMessageResult": {
+      "messages": [
+        {
+          "Attributes": null,
+          "Body": "{\n  \"Type\" : \"Notification\",\n  \"MessageId\" : \"redacted\",\n  \"TopicArn\" : \"arn:aws:sns:eu-west-1:redacted:mytopic\",\n  \"Subject\" : \"OrderAccepted\",\n  \"Message\" : \"{\\n    \\\"OrderId\\\": 1234,\\n    \\\"Id\\\": \\\"e3f84a55-b677-43df-8c24-92c170fdd89f\\\",\\n    \\\"TimeStamp\\\": \\\"2019-07-03T09:53:09.2956149Z\\\",\\n    \\\"RaisingComponent\\\": \\\"my sns publisher\\\"\\n}\\n\",\n  \"Timestamp\" : \"2019-07-10T12:38:58.830Z\",\n  \"SignatureVersion\" : \"1\",\n  \"Signature\" : \"redacted\",\n  \"SigningCertURL\" : \"redacted\",\n  \"UnsubscribeURL\" : \"redacted\"\n}",
+          "MD5OfBody": "redacted",
+          "MD5OfMessageAttributes": null,
+          "MessageAttributes": null,
+          "MessageId": "redacted",
+          "ReceiptHandle": "redacted"
+        }
+      ]
+    },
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+
+### Publishing to SNS
+
+JustSaying uses the SNS Publish API to send messages to SNS (documented [here](https://docs.aws.amazon.com/sns/latest/api/API_Publish.html)). The following parameters are set:
+
+#### TopicArn
+The ARN of the topic to which the message is being published.
+
+#### Subject
+JustSaying will set this to the type of the .NET message class - "OrderAccepted", for example.
```
martincostello: _"...to the type name of the .NET message class..."_?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+
+#### Message
+The actual message object, serialised to JSON. Because JustSaying messages must derive from the base Message class, the properties on that base class will be serialised into this JSON. On the base class, the `Id` property is set as a random GUID and the `TimeStamp` property is set to the current time (in UTC). The default serialisation settings will ignore null properties, so properties like `Conversation` will not appear in the JSON if they are left as the default null value. Enums are serialised as strings.
```
martincostello: "`Message`"?

martincostello: _"Enumerations are..."?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+
+In summary, suppose we have the following message class:
+
+````c#
```
martincostello: TIL `c#` works - I've been writing `csharp` forever!

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+        public class OrderAccepted : Message
+        {
+            public OrderAccepted(int orderId)
```
martincostello: Shall we just make the messages here super-simple and elide the constructors and assume they're just `new Foo()`?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+            {
+                OrderId = orderId;
+            }
+            public int OrderId { get; private set; }
```
martincostello: Similarly, just a plain `{ get; set; }`.

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+And suppose I create a message object from that class:
+
+````c#
+        var orderAccepted = new OrderAccepted(1234)
+        {
+            RaisingComponent = "my publisher",
+        };
+````
+
+This will result in the following JSON being created and populated on that Message parameter to the SNS Publish API:
```
martincostello: Shall we make a note here that the sent format isn't the same as the receive format in all cases (in case people only read this far)?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+JustSaying uses the SQS SendMessage API to send messages to SQS (documented [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html)). The following parameters are set:
+
+#### QueueUrl
+The URL of the queue to which the message is being sent.
+
+#### DelaySeconds
+JustSaying will set this value to be that of the `Delay` property on the `PublishMetadata` class which is supplied as a parameter to the `PublishAsync` method.
+
+#### MessageBody
+Aa JSON object containing two properties, `Subject` and `Message`. `Subject`, as with the SNS publisher, will be the type of the .NET message class. `Message` will be the actual message object, serialised to JSON. The same serialisation applies as for the SNS publisher.
```
martincostello: _"A JSON..."_

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+
+As an example, if I create the same OrderAccepted object as before and send that to SQS, then the value of this parameter will be:
+
+````json
+{
+    "Subject":"OrderAccepted",
```
martincostello: Nit: spaces after the `:`'s?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+
+### Subscribing to SQS
+
+Whether the message was originally published to an SNS topic or sent directly to an SQS queue, the subscriber will receive it by reading from an SQS queue. To do this, JustSaying uses the SQS ReceiveMessage API to receive messages from SQS (documented [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html)). The following parameters are set in JustSaying's request to this API:
+
+#### QueueUrl
+The URL of the queue from which messages are to be received.
+
+#### MaxNumberOfMessages
+JustSaying will vary this value depending on the number of CPU cores available and throttling configuration.
```
martincostello: Call out the `Throttle` (right name?) class here?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+
+````json
+{
+  "ReceiveMessageResponse": {
+    "ReceiveMessageResult": {
+      "messages": [
+        {
+          "Attributes": null,
+          "Body": "{\n  \"Type\" : \"Notification\",\n  \"MessageId\" : \"redacted\",\n  \"TopicArn\" : \"arn:aws:sns:eu-west-1:redacted:mytopic\",\n  \"Subject\" : \"OrderAccepted\",\n  \"Message\" : \"{\\n    \\\"OrderId\\\": 1234,\\n    \\\"Id\\\": \\\"e3f84a55-b677-43df-8c24-92c170fdd89f\\\",\\n    \\\"TimeStamp\\\": \\\"2019-07-03T09:53:09.2956149Z\\\",\\n    \\\"RaisingComponent\\\": \\\"my sns publisher\\\"\\n}\\n\",\n  \"Timestamp\" : \"2019-07-10T12:38:58.830Z\",\n  \"SignatureVersion\" : \"1\",\n  \"Signature\" : \"redacted\",\n  \"SigningCertURL\" : \"redacted\",\n  \"UnsubscribeURL\" : \"redacted\"\n}",
+          "MD5OfBody": "redacted",
```
martincostello: `"The base64-encoded MD5 hash of Body"`?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+          "MD5OfMessageAttributes": null,
+          "MessageAttributes": null,
+          "MessageId": "redacted",
```
martincostello: `"The message Id specified by the publisher"`?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+          "ReceiptHandle": "redacted"
```
martincostello: `"The AWS receipt handle"`?

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+        }
+      ]
+    },
+    "ResponseMetadata": {
+      "RequestId": "redacted"
```
martincostello: I just find `redacted` here an odd choice of word as they aren't secret, they're just superfluous. Maybe just `omitted`?

README.md
yetanotherchris: `OrderAcceptedMessage` for the type might be a little bit more explicit/obvious

README.md
slang25: It's a curse cause you find yourself doing it both ways and not sticking with one 😆

README.md
iancooper: Although we do this, there is no need to enforce this as a constraint. In fact, I suspect it would be better not to. I think we would be better to suggest that the Subject should either be an Imperative if it is a Command i.e. DoCalculation and an Event a Past Participle CalculationDone.

It would then be up to consuming frameworks to understand how to map this to a type that they could dispatch.

We could mention that by convention .NET clients just use the message class

README.md
iancooper: If we are trying to document a format, I would say that we expect to have an Id property and a TimeStamp property on a message, and that the C# library provides a base class Message that defaults the Guid to a random new Guid, timestamp to etc....

iancooper: Again this feels C# specific. We might say that we recommend that enumerated values are sent as strings to allow easier comprehension of the message and the C# library defaults to this

README.md
iancooper: Yeah, I think we need to identify this as the 'body' that we send, and then discuss the expectations around attributes etc after

README.md
iancooper: OK, definitely worth highlighting the point-to-point vs SNS constraint here, so that we are clear all our messages look the same whether via a dynamic recipient list router (SNS) or straight to the channel (SQS) Thought I would share some fancy terms ;-)

README.md
iancooper: 'elided' might be better

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+````
+
+This will result in the following JSON being created and populated on that Message parameter to the SNS Publish API:
+
+````json
+{
+    "OrderId": 1234,
+    "Id": "e3f84a55-b677-43df-8c24-92c170fdd89f",
+    "TimeStamp": "2019-07-03T09:53:09.2956149Z",
+    "RaisingComponent": "my sns publisher"
```
bradenschmidt: `"my publisher"` is what is used above

README.md
```diff
@@ -275,6 +275,126 @@ An example policy would look like;
...
+            public int OrderId { get; private set; }
+        }
+````
+
+And suppose I create a message object from that class:
+
+````c#
+        var orderAccepted = new OrderAccepted(1234)
+        {
+            RaisingComponent = "my publisher",
```
bradenschmidt: Is there any guidance around what `RaisingComponent` should actually be? `my publisher` is confuses me.

martincostello: Internally it's generally the name of a feature/application, such as `"orderapi"`.

README.md
brainmurphy: Well, if we don't describe what JustSaying (a C# library) does, then how do people not using JustSaying know how to work together? By saying that we serialise enumerations as strings when publishing, a Python/Java/whatever developer then knows what JSON they'll be getting when they subscribe.

Hmmmmm, instead of saying "this is what JustSaying does", is it better to invert it and say rephrase it as "this is what you should do to send a message to JustSaying / what you should expect when you receive a message from JustSaying"?

README.md
brainmurphy: If you (in Python, say) publish a message to SNS that you want JustSaying to be capable of subscribing to, then you need to know that JustSaying uses the type name here because AWS sends that string to SQS where it's visible from the ReceiveMessage API in Messages ➡️ Message ➡️ Body ➡️ Subject and the JustSaying subscriber then assumes that string is a type name for deserialisation purposes. The constraints for allowed values in the `Subject` property are much more permissive than those for C# identifiers, so if you don't follow the C# rules you can easily publish messages that JustSaying will be unable to subscribe to (without replacing the serialisation implementation (which you can do...)).

README.md
brainmurphy: Nope. Internally at Just Eat, we set it to the name of the app (the independently-deployable component) that's publishing the message.

README.md
brainmurphy: Yeah, that's fair. I'm reluctant to replicate the AWS docs, but if people expect "what goes into SNS == what comes out of SQS", then that's worth calling out...

README.md
brainmurphy: Not sure I follow... Subscribing is always via SQS, whether the message originally came via SNS or SQS, so `MaxNumberOfMessages` config is independent of point-to-point / SNS.

README.md
brainmurphy: "Ommitted" sounds OK to me, I'm totally not jumping on that "elide" bandwagon 🤣

I just wanted to remove the original values since they're just giant opaque strings that JustSaying has no influence over and makes no use of so it might look distracting/pointless. I'm in two minds about adding this JSON at all, since it's just copy/paste from the AWS docs and the only thing we care about is a subset of the `Body` property string. Is it useful?

martincostello: I think it's good to see it in raw format to set some context. I had to reverse engineer this for #548, so this would have been useful if it had already existed.

README.md
bradenschmidt: Given the confusion already it'd be nice to see a recommendation

README.md
brainmurphy: I've added a comment above:

> ... the JSON sent in this parameter becomes a subset of the JSON that will be delivered (by the SNS service) to an SQS subscriber ...

README.md
brainmurphy: I've added a comment about the restrictions on the Subject parameter:

> Since a JustSaying _subscriber_ relies on this string to know what type to deserialise the JSON in the Message parameter into, non-JustSaying _publishers_ should restrict this parameter to alphanumeric characters.

README.md
brainmurphy: Done.

README.md
brainmurphy: I've added a table that lists the the properties in the base Message class, which ones JustSaying sets (`Id` and `TimeStamp`) and their intended use (or as best I can guess based on the names!) Also noted that properties which are _not_ set don't get serialised.

README.md
brainmurphy: Done. All hail the giant JSON blob 🙏

README.md
brainmurphy: I added a table that lists the intended use of these fields (or at least what I assume it to be based on the name!) Also tidied up the examples to use "orderapi" for `RaisingComponent` which should make things clearer.

README.md
brainmurphy: Added a mention of the `Throttled` class.

README.md
brainmurphy: Think I've clarified this: I've added a table with the properties on that base class, noting that `Id` and `TimeStamp` will be there by default.

brainmurphy:
@iancooper @bradenschmidt Can you gentlemen have another look at this? I'd like to get this moved over to somewhere more sensible once we're happy that it's an improvement.

README.md
```diff
@@ -275,6 +271,142 @@ An example policy would look like;
 }
 ```
```
iancooper: I assume the following sample code relates to IAM policy, but there is no text to describe that.

README.md
```diff
@@ -275,6 +271,142 @@ An example policy would look like;
...
+|Version|string|The message version - "2", for example.|No|
+|SourceIp|string|The IP address of the machine which published the message.|No|
+|Tenant|string|In a multi-tenant architecture, a string to identify the tenant for which the message is applicable.|No|
+|Conversation|string|A string which can be used to correlate multiple messages (to identify messages belonging to a single operation/customer journey, for example).|No|
+
+Properties which are not populated will not appear in the message JSON. Enumerations are serialised as strings.
+
+In summary, suppose we have the following message class:
+````c#
+        public class OrderAccepted : Message
```
iancooper: Should we describe the contents of the Message base class here, for clarity, or just leave the properties discussed below

PR #556, opened 2019-07-15
Remove MSBuildAllProjects

slang25:
Following guidance from here:
http://www.panopticoncentral.net/2019/07/12/msbuildallprojects-considered-harmful/
Given we have pinned the SDK we can be sure we are using MSBuild 16+, so it's a safe change.


martincostello:
Thanks @stuart-lang

PR #561, opened 2019-08-07
Create listen loop Task as long running

martincostello:
Create the task for the listen loop as a long running task.

From the docs (_emphasis mine_):

> Specifies that a task will be a long-running, coarse-grained operation involving fewer, larger components than fine-grained systems. It provides a hint to the System.Threading.Tasks.TaskScheduler that oversubscription may be warranted.
> Oversubscription lets you create more threads than the available number of hardware threads. It also provides a hint to the task scheduler that an additional thread might be required for the task _so that it does not block the forward progress of other threads or work items on the local thread-pool queue_.

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -112,14 +112,17 @@ namespace JustSaying.AwsTools.MessageHandling
             var queueName = _queue.QueueName;
             var region = _queue.Region.SystemName;
 
-            // Run task in background
-            // ListenLoop will cancel gracefully, so no need to pass cancellation token to Task.Run
-            _ = Task.Run(async () =>
-            {
-                await ListenLoopAsync(cancellationToken).ConfigureAwait(false);
-                IsListening = false;
-                _log.LogInformation("Stopped listening on queue '{QueueName}' in region '{Region}'.", queueName, region);
-            });
+            // Run listen loop in the background
+            _ = Task.Factory.StartNew(
+                async () =>
+                {
+                    await ListenLoopAsync(cancellationToken).ConfigureAwait(false);
+                    IsListening = false;
+                    _log.LogInformation("Stopped listening on queue '{QueueName}' in region '{Region}'.", queueName, region);
+                },
+                cancellationToken,
+                TaskCreationOptions.LongRunning,
+                TaskScheduler.Default);
 
             IsListening = true;
             _log.LogInformation("Starting listening on queue '{QueueName}' in region '{Region}'.", queueName, region);


martincostello:
Will merge pending some manual testing I'm doing locally.

slang25:
I this actually isn't what you want to do, I'll dig out some previous discussion we had around this.

slang25:
Thread here:
https://github.com/justeat/JustSaying/pull/403#discussion_r230790423

TL;DR - The `StartNew` returning `Task` only represents the synchronous start of the async delegate, so here it would be a tiny part of the listen loop.

martincostello:
I'm taking two points away from the guidance if I've read it correctly:
  1. `Task.Run()` is bad - this PR fixes it by explicitly avoiding stealing from the thread pool by using `Task.Factory.StartNew(...)` instead.
  1. `TaskCreationOptions.LongRunning` doesn't help, so should be changed to `None`.

slang25:
I'm not sharing your understanding on first point. This area of code led me to write this: https://stu.dev/miscellaneous-csharp-async-tips/#replace-task-factory-startnew

For the listen loop we will be using the threadpool threads many times for tiny work, there's nothing that keeps a thread blocked for a long period of time (like the example in the post).

martincostello:
I thought the overall message loop was long-running and doing stuff like continuously querying the message loop. Based on that, [this point and the first few sentences](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/7e773b7ac2e266c1368113c63db5a8a3b2f5334d/AsyncGuidance.md#avoid-using-taskrun-for-long-running-work-that-blocks-the-thread) lead to the conclusion that `Task.Run()` is bad in this scenario.

slang25:
Agreed with your understanding here. Our message loop is asynchronous, so as soon as we hit this line:
https://github.com/justeat/JustSaying/blob/88f96637690f765e4a2c7cfaaee446d764f40729/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs#L138

we are done with the thread, so there's no long blocking, and what we are saying is we want to use a dedicated thread for that minuscule work. It's not a problem, it just doesn't make sense to do here.

slang25:
For another explanation, the third bulb in the guidance you are referring to is what is happening here:
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/7e773b7ac2e266c1368113c63db5a8a3b2f5334d/AsyncGuidance.md#avoid-using-taskrun-for-long-running-work-that-blocks-the-thread

martincostello:
So I get that the third lightbulb means we don't need `LongRunning`, but not so sure about the first. Does the fact we're awaiting mean that there's no different between `Task.Run(...)` and `Task.Factory.StartNew(...)`?

slang25:
That's my understanding, we'd only need `Task.Factory.StartNew(...)` if we wanted the `LongRunning` option, otherwise they are just behaving the same, and `StartNew` can only lead to confusion IMO.

If we encounter blocking code then we should look to eliminate it or get it off the theadpool with the technique here, but I don't think we have that problem.

martincostello:
I think `StartNew()` looks a bit more obvious in its intention that `Task.Run()`, so I think I've rather leave that like it's been changed to. I'll add a bunch of documentation and links to the above so that the next time something thinks about long running tasks they know it's redundant.

Then I just need to change the enum value to `None`.

slang25:
I'm not glued to my opinions, so would happily be persuaded on this.

`StartNew()` seems confusing to me, to make the point: a `cancellationToken` is passed in, what do you think will happen when cancellation is triggered?

Answer: probably nothing, unless it is triggered before after the synchronous part of `ListenLoopAsync` has finished. The `Task` returned from `StartNew()` doesn't represent the overall task, this will (and has) cause confusion, whereas `Task.Run` is day-to-day code that people understand how it behaves.

martincostello:
I would expect the task to not be started if hadn't already, but that's it.

It seems more explicit to me that you want something _new_ to start off and run, as conceptually we're doing something long-running, even though the `async`-ness means it isn't truly.

`Task.Run()` to me is more for short-lived stuff, which _technically_ this is, but conceptually isn't.

slang25:
For me 'Task.Run()` is for non-blocking work, regardless of the actual duration. `StartNew` is only for sync delegates in my mind and is not to be mixed with async-await:
https://stackoverflow.com/a/24777502/2225808

slang25:
Perhaps the answer is to assign the `Task` to a field with a name that suggests it's intended to run indefinitely?

martincostello:
Gotta love a confusing API surface 😄 

In that case I'll close this and go back to hunting my thread-starvation stuff elsewhere.

PR #563, opened 2019-08-13
Retention period should be 4 days.

brainmurphy:
This change to the retention period was lost at some point. Was flagged in the discussion for #419.

Retention period of 10 minutes is too short and likely to lead to lost mesages.

#419

### DIFF
--- a/JustSaying/AwsTools/JustSayingConstants.cs
+++ b/JustSaying/AwsTools/JustSayingConstants.cs
@@ -41,7 +41,7 @@ namespace JustSaying.AwsTools
         /// <summary>
         /// Default message retention period on a queue in seconds
         /// </summary>
-        public static int DEFAULT_RETENTION_PERIOD = 60 * 10;    //10 minutes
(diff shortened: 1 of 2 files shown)


JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -41,7 +41,7 @@ public static class JustSayingConstants
         /// <summary>
         /// Default message retention period on a queue in seconds
         /// </summary>
-        public static int DEFAULT_RETENTION_PERIOD = 60 * 10;    //10 minutes
+        public static int DEFAULT_RETENTION_PERIOD = 345600;    //4 days
```
AOOlivares: wouldn't be better to have something like **3600 * 24 * 4**?
The end result is the same but is easier to read.

brainmurphy: I want this to be exactly, precisely, 100% identical to what was committed (multiple times) before. All of these have been changed to the less-idiotic `TimeSpan` in v7, by the way.

PR #565, opened 2019-08-14
Retention period should be 4 days.

brainmurphy:
This change to the retention period was lost at some point. Was flagged in the discussion for #419.

Retention period of 10 minutes is too short and likely to lead to lost mesages.

#419


martincostello:
Branchapalooza!

Issue #291, opened 2017-01-10
No VisibilityTimeout Renewal


slang25:
When we take messages from an SQS queue, we do so with a `VisibilityTimeout` ([Docs](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html)) so that as we process the message, no other worker can see the inflight message, this acts as a peek lock. It's important because should anything bad happen while processing the message, we know that it will get processed again, there is no responsibility on the worker to "put it back". The default timeout is 30 seconds, and is configurable within JustSaying.

However we have now enforced our handlers to complete their work within this timeout, otherwise another worker will start processing the same message, and now will be doing concurrent work, and possibly duplicate work, as there is no hard requirement for consumers to make their processing idempotent (although SQS doesn't make any guarantees of exactly once delivery).

I suggest we add a visibility timeout renewal mechanism.

Here is an example of how it could work; once we take a message off the queue, using `Task.Run` passing in a CancellationToken, we use the same timeout the message has and await a `Task.Delay(messageTimeout,ct)`, then use `ChangeMessageVisibilityAsync` to renew the timeout for another duration, then loop until the CT is cancelled. The CT will be cancelled once the handler has completed. If that doesn't make sense I will follow up with a rough PR.

Pros:
- Safer defaults for people not wanting to think hard about configuration. aka "Pit of success"
- Better robustness when handlers are delayed abnormally.
- Visibility timeouts could be set more aggressively, whereas currently they need to be conservative as to not hit this scenario.

Cons:
- More complexity
- `Task.Run` or however you cut it, will be requiring another thread from the pool (per message), even if only very breifly.
- Must be careful not to introduce any new race conditions, memory leaks, or bugs in general.

With SQS FIFO queues on the way, and our current [IMessageLock](https://github.com/justeat/JustSaying/blob/ffe602c44422fbe3259f9a51ade947f36689c1a9/JustSaying.Messaging/MessageHandling/IMessageLock.cs) behavior which may offer some overlap, is this a sensible addition? What are peoples thoughts?

je-ewan:
surely expected behaviour for a worker queue is that is that if a message is not processed after 'timeout period' the job is assigned to another worker. 

It sounds like you would want to remove this timeout completely. which, for me, would introduce unexpected behaviour?

AnthonySteele:
[...]

PR #571, opened 2019-08-26
Cancellable handler

martincostello:
I had an idea this morning based on the changes in #546, so I thought I'd see how far I could get with it and submit a PR for discussion.

In the event that the issue that lead to #546 occurring and tasks to process messages back up, it's possible that a task to process a message may be trying to process a message that was received so long ago that the message visibility timeout has expired and it is now a potential duplicate waiting to be processed.

If the handler had a `CancellationToken` that cancels after the _approximate_ message visibility timeout expires, JustSaying could enable two behaviours:

  1) It could self-cancel trying to do anything with the message before getting to the handler and abandoning it (potentially just extending the visibility instead, but I've not done that here).
  2) The handler, when called, could observe the cancellation token itself and make a decision about what to do with the message. It could return `false` to send it back to the queue (where it could have since been handled elsewhere anyway), or it could return `true` on the basis, if this has expired I'll just ignore it and pretend I've dealt with it.

This is implemented here by adding the `ICancellableHandlerAsync<T>` interface that derives from `IHandlerAsync<T>`, and then plumbing it in appropriately so it's used if implemented by the handler for a particular message and extending things like the exactly-once handler and stopwatch wrappers.

Replaces #549.

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
+++ b/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
@@ -67,6 +67,8 @@ namespace JustSaying.AwsTools.MessageHandling
             }
         }
 
+        public Func<DateTimeOffset> Clock { get; set; } = () => DateTimeOffset.UtcNow;
+
         public string Queue => _queue.QueueName;
 
         // ToDo: This should not be here.
@@ -177,6 +179,13 @@ namespace JustSaying.AwsTools.MessageHandling
                     continue;
                 }
 
+                if (sqsMessageResponse == null || sqsMessageResponse.Messages.Count < 1)
+                {
+                    continue;
+                }
+
+                DateTimeOffset receivedAt = Clock();
+
                 try
                 {
                     foreach (var message in sqsMessageResponse.Messages)
@@ -186,7 +195,7 @@ namespace JustSaying.AwsTools.MessageHandling
                             return;
                         }
 
-                        if (!await TryHandleMessageAsync(message, ct).ConfigureAwait(false))
+                        if (!await TryHandleMessageAsync(message, receivedAt, ct).ConfigureAwait(false))
                         {
                             // No worker free to process any messages
                             _log.LogWarning(
@@ -283,11 +292,62 @@ namespace JustSaying.AwsTools.MessageHandling
             return messagesToRequest;
         }
 
-        private Task<bool> TryHandleMessageAsync(Amazon.SQS.Model.Message message, CancellationToken ct)
+        private Task<bool> TryHandleMessageAsync(Amazon.SQS.Model.Message message, DateTimeOffset receivedAt, CancellationToken ct)
         {
+            DateTimeOffset utcNow = Clock();
+            TimeSpan timeUntilMessageExpiry = utcNow - receivedAt;
+
+            if (timeUntilMessageExpiry < TimeSpan.Zero)
+            {
+                _log.LogWarning(
+                    "Message with Id {MessageId} was not handled because the message visibility timeout expired.",
+                    message.MessageId);
+
+                return Task.FromResult(false);
+            }
+
             async Task DispatchAsync()
             {
-                await _messageDispatcher.DispatchMessage(message, ct);
+                utcNow = Clock();
+                timeUntilMessageExpiry = utcNow - receivedAt;
+
+                if (timeUntilMessageExpiry <= TimeSpan.Zero)
+                {
+                    _log.LogWarning(
+                        "Message with Id {MessageId} was not handled because the message visibility timeout expired while waiting to dispatch it.",
+                        message.MessageId);
+
+                    return;
+                }
+
+                using (var timeoutSource = new CancellationTokenSource(timeUntilMessageExpiry))
+                using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(ct, timeoutSource.Token))
+                {
+                    if (linkedSource.IsCancellationRequested)
+                    {
+                        // Either the message has become stale or the message loop has been stopped
+                        return;
+                    }
+
+                    try
+                    {
+                        await _messageDispatcher.DispatchMessage(message, linkedSource.Token);
+                    }
+                    catch (OperationCanceledException ex) when (ex.CancellationToken == ct)
+                    {
+                        _log.LogWarning(
+                            ex,
+                            "Handling of message with Id {MessageId} was canceled because the message loop was canceled.",
+                            message.MessageId);
+                    }
+                    catch (OperationCanceledException ex) when (ex.CancellationToken == timeoutSource.Token)
+                    {
+                        _log.LogWarning(
+                            ex,
+                            "Handling of message with Id {MessageId} was canceled because the message visibility timeout expired.",
+                            message.MessageId);
+                    }
+                }
             }
 
             return _messageProcessingStrategy.StartWorkerAsync(DispatchAsync, ct);
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/HandlerMapTests.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/HandlerMapTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using JustSaying.AwsTools.MessageHandling;
 using JustSaying.Models;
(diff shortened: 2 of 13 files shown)


slang25:
Just to add another option, it could renew the message visibility timeout: https://github.com/justeat/JustSaying/issues/291
This is what [Rebus does](https://github.com/rebus-org/Rebus.AmazonSQS/blob/master/Rebus.AmazonSQS/AmazonSQS/AmazonSqsTransport.cs#L368), and kind of what I'd expect by default for this type of library.

martincostello:
To be honest I thought we already had that capability, so I've re-opened it. It's also something we'd probably want as opt-in?

However, that's not what I was trying to achieve here. This is more of a reaction to a backed-up local queue (see #546).

In the event of messages being sat on scheduled Tasks waiting to progress and not being processed, the chances are AWS has already considered the message "lost" and redelivered it somewhere else. By that point _if_ the task does eventually run, it's probably dealing with what is now a duplicate, and a waste of processing. By observing the fact the timeout has _probably_ elapsed, we can just not process the message at all and end the Task and release some resources back to the worker to try and make some headway through the queue.

In that scenario, renewing the message visibility timeout would probably make things worse as it would leave messages on a backed-up local queue even longer?

slang25:
I would expect an "auto-renew" to be on by default, but you'd certainly want control. This is something that the PE team needed a long time back for their deployment tool, so we exposed the `RecieptHandle` on the messages so they could do it themselves.

I happen to have the opposite view to you on this 😄, I think it more directly solves the problem of local queues holding stale messages. Cancelling sounds more haphazard as a general pattern here. If one worker has the message, it has a duty to process it, and if it doesn't it will eventually become visible to others.

I could see that cancelling before a handler gets the message would solve your problem, but adding cancellation to handlers feels wrong from an event-driven architecture perspective. If a hander is still working as (or soon before) a message visibility times out, it should let others know it's still working on it. It's "at least once" message delivery after all, cancellation doesn't belong here.

martincostello:
But here, the cancellation is _pre_-handler. Once control is passed to the handler, it's just as it is today, except you have to opt-in to implementing the new interface, and then have to decide to observe the token to do something different in your handler implementation.

This is just trying to improve the situation where the internals get stuck in a never-progressing backlog (which is what we had in the issue that lead to #546 as the `Task` queue was backing up) by never calling the handler in the first place.

If we wanted to do something to allow renewing the message visibility, I think that's something else that would need to be a first-class concept somewhere in the handlers to give you the hook to decide to renew/extend the visibility _after_ the handler has already seen the message to make a decision.

Maybe that could be morphed into a more detailed handler concept that did both separately, so for more advanced scenarios you'd get like this very-of-the-top-of-my-head thing:

```csharp
public interface ISuperDuperHandler<in T>
{
    Task<bool> HandleAsync(HandlerContext<T> context);
}

public abstract class HandlerContext<T>
    where T : Message
{
    public CancellationToken CancellationToken { get; protected set; }

    public T Message { get; protected set; }

    public MessageContext MessageContext { get; protected set; }

    public abstract async Task ExtendVisibilityAsync(TimeSpan value);
}

internal sealed class SuperDuperHandlerContext<T> : HandlerContext<T>
    where T : Message
{
    private _client;

    public SuperDuperHandlerContext(
        T message,
        MessageContext messageContext,
        SomeClient client)
    {
        Message = message;
        MessageContext = messageContext;
        _client = client;
    }

    public override async Task ExtendVisibilityAsync(TimeSpan value)
    {
        await _client.ExtendTheVisbilityAsync(Message.Id, value, CancellationToken);
    }
}
```

Then you could do stuff like this:

```csharp
public class MyHandler : ISuperDuperHandler<MyMessage>
{
    public async Task<bool> HandleAsync(HandlerContext<T> context)
    {
        if (context.CancellationToken.IsCancellationRequested)
        {
            // Too old or shutting down, decide to junk it and let someone else handle it
            return false;
        }

        if (context.Message.IsMessageHuge)
        {
            await context.ExtendVisibilityAsync(TimeSpan.FromHours(1));
        }

        await DoExpensiveStuffAsync(context.Message);

        return true;
    }
}
```

slang25:
Also want to add, I'm not saying that by default the behaviour is that messages are continually renewed in terms of visibility, there would be a configurable policy of how long / how many renews before giving up.

slang25:
Ah, I was (or maybe still am) confused 😄 

So if I understand you, `ICancellableHandlerAsync<T>` lets us build a pipeline that's cancellable, before calling the user provider handler.

If we want to let the user optionally know if the message had timed out, the cancellation token could lead to a little confusion, as the signature would imply that the ct is live, and you'd be expected to constantly check it and flow it for cooperative cancellation.

martincostello:
No, the interface would be to allow a developer to _opt-in_ to doing something different in the handler if the message is considered "stale" or if the message loop is shutting down.

To me, visibility locks is a different feature for people who need to process messages that _may_ take a long time to handle in the _handler_.

I don't think they're a solution to a backed-up internal queue that is operating with no context that could be potentially having messages lying around in memory that are dozens of minutes old.

Effectively this makes this PR two different, related, things in one:
  1. If a message has been waiting to be passed to a handler since receipt for longer than the visibility timeout, don't bother trying to process (effectively, abandon it as if the service had crashed).
  1.  Provides a separate opt-in mechanism that allows a handler to decide to do different logic if the message loop has been requested to stop or if the message visibility timeout has lapsed.

slang25:
> To me, visibility locks is a different feature for people who need to process messages that may take a long time to handle in the handler.

The way I see it, JustSaying grabs a handful of messages and will use the processing strategy to decide when to call the handlers, it's its duty to try to guarantee correctness (within reason), in this case by dropping the messages (FIFO as this change is), or with re-acquiring visibility locks. This wouldn't need much further complexity than what is done in this PR, there is still the same timer involved. (I sense I won't sell you in the short term)

I see though that dropping the messages is a good solution, and is the minimal change. I'd add a suggestion of being able to add (or subtract rather) a buffer to the timeout, if I'm 0.1 seconds away from someone else picking up the message, and the timeout is 30 seconds, I'd probably want to still drop it.

On the opt-in behaviour, the only thing I don't like is the way it's presented, when I look at the `ICancellableHandlerAsync<T>` type I see a message handler with cancellation, not as you describe. I'd like to see it be a little more unambiguous, but I haven't had much sleep, so I could just be not seeing things clearly 😄

slang25:
For some more reading in favour of extending timeouts, here's the guidance on the matter:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/working-with-messages.html#processing-messages-timely-manner

The thing with dropping messages is that the messages being dropped will be mostly random, so it's very easy for messages to be dropped multiple times, exceeding the max receive count and ending up in a dead letter queue. If you extend locks, you generally get a much more ordered processing of messages, which will leave messages in queues where they are nice and safe.

slang25:
I think it's worth comparing a prototype of that guidance, when a message is received at the earlier occasion we create a heartbeat that renews the lock, it would be configurable and defaults to half the timeout, and extends by the timeout.

So if it's 30 seconds, schedule a task in 15 secs to renew by 30 (45s accumulative), then again 15 secs later forever pushing out the lock until some overall timeout. Once the handler has completed, any heartbeat task is cancelled via a cts.

Waddaya say? I can have a bash and we can compare notes?

martincostello:
I'd be happy to drop the `ICancellableHandler<T>` part (though I think it's a nice addition) and just keep the _"this has been sat around internally for too long"_ part.

If you want to try a different tack and raise a PR I'll take a look.

slang25:
Thanks, I'll start work in a different PR (likely not today but soon).

Sorry if this sounds negative (you know I love your work ❤️), but now I've considered it some more I think this PR should be closed, as it's trading off one problem for another an not really solving the issue, I hope my points in the previous comments have persuaded you too.

martincostello:
I'll re-review when I get the chance and see what could be removed to make it simpler but still provide a mitigation for a runaway Task queue, as ultimately that issue would have brought down the worker due to a memory leak.

Also waiting for anyone else to chip in with their thoughts 😄

slang25:
In other news, sometimes a repo name tells a story...
image

slang25:
Oh god, I've just encountered the same issue, what gives? The fork upstream link is gone from GH repos for JustSaying 😞

JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs
```diff
@@ -177,6 +179,13 @@ internal async Task ListenLoopAsync(CancellationToken ct)
                     continue;
                 }
 
+                if (sqsMessageResponse == null || sqsMessageResponse.Messages.Count < 1)
```
martincostello: This block is duplicated due to a bad merge and needs removing.

PR #575, opened 2019-08-30
Issue-551 Add server-side encryption for sns topic

martalexandra:
Adds support for SNS server-side encryption by passing encryption option in config.
Follows a similar approach to #363

(#551)


JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -96,5 +97,68 @@ public async Task<bool> CreateAsync()
...
+                var request = new SetTopicAttributesRequest
+                {
+                    TopicArn = Arn,
+                    AttributeName = JustSayingConstants.AttributeEncryptionKeyId,
+                    AttributeValue = config?.KmsMasterKeyId ?? string.Empty
+                };
+
+                var response = await Client.SetTopicAttributesAsync(request).ConfigureAwait(false);
+
+                if (response.HttpStatusCode == HttpStatusCode.OK)
```
martincostello: We seem to be ignoring any HTTP errors here, they're not even being logged, which seems incorrect.

martalexandra: I used the same pattern that is there for the sqs queue, which also doesn't fail if the request to set the attributes fails, but agree it shouldn't simply ignore it, so added the logging

JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -96,5 +97,72 @@ public async Task<bool> CreateAsync()
...
+                {
+                    ServerSideEncryption = string.IsNullOrEmpty(config?.KmsMasterKeyId)
+                        ? null
+                        : config;
+                }
+                else
+                {
+                    _log.LogWarning($"Request to set topic attribute {request.AttributeName} to {request.AttributeValue} failed with status code {response.HttpStatusCode}.");
```
martincostello: This would be better if it used structured logging syntax:

```csharp
_log.LogWarning("Request to set topic attribute {TopicAttributeName} to {TopicAttributeValue} failed with status code {HttpStatusCode}.", request.AttributeName, request.AttributeValue, response.HttpStatusCode);
```

martalexandra: Done :)

martalexandra:
Hey, thanks for the approvals! Could you please retry the tests to re check? The failing is a previous one and I've only changed to use structured logging for the last run.

martalexandra:
Hey @martincostello could you please trigger the test again as per the above? Seems it still only has the first run and I don't think my change in the logging would cause the failure.
Also, do you have an idea when this could get merged, once the tests are passing?
Thanks

martincostello:
@martalexandra I've re-triggered the build. In terms of merging, I'd have thought we can merge it as soon as the statuses are all green.

PR #578, opened 2019-10-12
Build infrastructure updates

martincostello:
*  Build the whole solution, not just specific projects and TFMs on Linux/macOS.
  * Remove redundant package reference for AWSSDK.Core as its transient from the other two.
  * Use more recent minimum version for AWS SDK.
  * Update tests to latest AWS SDKs.
  * Use a build image for Visual Studio 2019 in AppVeyor.
  * Use newer format for NuGet Symbol packages.
  * Ignore code coverage artifacts.
  * Update various NuGet packages to their latest versions.
  * Fix or suppress new analyzer warnings.
  * Update to .NET Core 3.0 SDK.
  * Update tests and samples to .NET Core 3.0.
  * Delete or disable problematic tests (see #579).
  * Switch from OpenCover to coverlet.


Directory.Build.props
```diff
@@ -43,10 +47,21 @@
     <LangVersion>latest</LangVersion>
     <RepositoryUrl>$(PackageProjectUrl).git</RepositoryUrl>
     <SignAssembly>true</SignAssembly>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
slang25: I think I've raised this before, but this is redundant for this package as symbols are embedded.

In this file there should only be `SymbolPackageFormat` or

slang25:
Very nice 😃

Directory.Build.props
martincostello: I changed this to guard against an issue I had with publishing one of my own packages which stil used the old format where they get published from AppVeyor to a deprecated server which has since had expired TLS cert that made publishing fail: https://ci.appveyor.com/project/martincostello/sqllocaldb/builds/28012360#L181

Even if it's not in use, setting this means if we change in the future it's future-proofed.

Issue #579, opened 2019-10-12
Restore test coverage to previous levels


martincostello:
As part of refactoring the tests and removing JustBehave etc. the test coverage has taken a hit. Before shipping 7.0.0, we need to get the coverage back to a decent level with integration tests using the pattern we reach consensus on being "desired".

As part of this, some _specific_ tests have been deleted. The test cases removed are documented below.

  * `WithSqsTopicSubscriber.WhenSubscribingToTopicInAnotherAccount.ICanReceiveMessagePublishedToTopicInAnotherAccount`
  * `WithSqsTopicSubscriber.WhenOneBusIsSubscribedToTwoRegions.MessagesPublishedToBothRegionsWillBeReceived`
 * `WithSqsTopicSubscriber.WhenAFailoverRegionIsSetup.MessagesArePublishedToTheActiveRegion`
  * `WithSqsPointToPointSubscriber.WhenOneBusIsSubscribedToTwoRegions.MessagesPublishedToBothRegionsWillBeReceived`
  * `WithSqsPointToPointSubscriber.WhenAFailoverRegionIsSetup.MessagesArePublishedToTheActiveRegion`

~Some other tests are flaky and have been marked with `[AwsFact(Skip = "Test is flaky")]`~ Resolved by #580

gkinsman:
I think we can probably close this, as coverage is now up to ~81% or so in master?

PR #580, opened 2019-10-13
Improve tests

martincostello:
* Fix flaky tests for message locks by improving the message lock implementation and assertions (see #579).
  * Upload code coverage from Travis CI.
  * Add logging to `ExactlyOnceHandler`.
  * Make some types `internal`.


src/JustSaying/Messaging/MessageHandling/ExactlyOnceHandler.cs
```diff
@@ -1,22 +1,30 @@
 using System;
 using System.Threading.Tasks;
 using JustSaying.Models;
+using Microsoft.Extensions.Logging;
 
 namespace JustSaying.Messaging.MessageHandling
 {
-    public class ExactlyOnceHandler<T> : IHandlerAsync<T> where T : Message
+    internal sealed class ExactlyOnceHandler<T> : IHandlerAsync<T> where T : Message
```
adammorr: Apologies if I have missed this in a previous change but what's the reason for this no longer being public? Has the usage changed?

martincostello: I didn't see a specific use case for it being public given that it seems to be an implementation detail of elsewhere in JustSaying, so I figured I'd try and rationalise the public API surface a bit more. I can undo it if necessary, but 7's already a major change anyway so breaking changes like this are doable.

PR #617, opened 2019-12-12
Remove instance position from SqsReadConfiguration

adammorr:
It feels to me like instance position was an internal concern prior to releasing JustSaying as an open source project so this PR is to 🔥 it


adammorr:
Yeah I figured as much. I'll have a go at rationalising the SNS/SQS configuration in a follow up PR

adammorr:
I'll also fix that test...

tests/JustSaying.IntegrationTests/Fluent/Publishing/WhenPublishingWithoutAMonitor.cs
```diff
@@ -49,10 +49,10 @@ public async Task A_Message_Can_Still_Be_Published_To_A_Topic()
                         (config) => config.WithPublishFailureBackoff(TimeSpan.FromMilliseconds(1))
                                           .WithPublishFailureReattempts(1));
 
+                    ;
```
martincostello: ?

adammorr: 🤦‍♂ I'm blaming Rider

PR #620, opened 2019-12-13
Change how we expose the naming of topics and queues

adammorr:
...what could go wrong...

One of the things I feel is confusing about JustSaying is the naming strategy.

When configuring JustSaying subscribers and publishers, you are provided with a number of ways to configure resource names including properties on the configuration classes as well as any implementation of `INamingStrategy`, without it being clear how they interact.

What I'm proposing is we remove the `INamingStrategy` concept and only use the fluent API to configure resource names. I'll try to ensure that it's straightforward enough for someone to write an extension method in case a centralised means of naming is required...kind of like a naming strategy :)


src/JustSaying/JustSayingFluently.cs
```diff
@@ -469,16 +465,20 @@ public IInterrogationResponse WhatDoIHave()
...
+        private static string GetOrUseDefaultTopicName<T>(string topicName)
         {
-            _awsClientFactoryProxy.SetAwsClientFactory(awsClientFactory);
-            return this;
+            return string.IsNullOrWhiteSpace(topicName) ? typeof(T).ToTopicName() : topicName;
+        }
+
+        private static string GetOrUseDefaultQueueName<T>(string queueName)
+        {
+            return string.IsNullOrWhiteSpace(queueName) ? typeof(T).ToTopicName() : queueName;
```
maurofranchi: `typeoof(T).ToQueueName()`?

adammorr: That would make sense however I just wanted to mirror what was in the soon-to-be-possibly-removed [DefaultNamingStrategy](https://github.com/justeat/JustSaying/blob/master/src/JustSaying/DefaultNamingStrategy.cs#L20)

maurofranchi:
Is it something we really want to remove from JustSaying? The naming strategy is something consumers can plug-in if they want to change it and have a different name for their queues. Having a standard way to do it with the default Naming Strategy is not a bad thing IMO as it allows you not to configure extra parameters like the queue name itself (as it uses the message, environment and tenant values that you have to configure anyway).

src/JustSaying/JustSayingFluently.cs
maurofranchi: I figured but maybe as you are changing it, it might be worth having a `ToQueueName` extension name too. I am not that fussed about it though.

adammorr:
Maybe, but in its current form it's quite confusing especially when you're building your stack and there are methods like "WithQueue(...)" where you can provide a name - it doesn't make for a very intuitive experience. There's no reason why a consumer couldn't provide their own mechanism for creating queue names which could plug into the fluent API rather than it all be hidden away in a naming strategy by default.

I'm not pushing this change as something we should definitely do but if it at least creates a discussion around how we expose the naming of resources to the consumers of JustSaying in V7 then that's great!

src/JustSaying/JustSayingFluently.cs
adammorr: Sure, I'll make that change

adammorr: Added `ToQueueName()` and ensuring the resource name lengths are respected for queues and topics

jaimalchohan:
I agree, `INamingStrategy` is confusing.

Could you post a sample of what you propose the fluent configuration should look like (apologies, but I found this hard to grasp from the commits)

adammorr:
So my thinking was you either explicitly provide a topic / queue name through the API or the sanitised type name of the message will be used (similar to what was in the default naming strategy)...

* `ForTopic<Blah>()' would create a topic called "blah"
* 'ForTopic<Blah>("my-awesome-topic")` would create a topic called "my-awesome-topic"

If we decide this approach is sensible then I'll update the docs to make it clearer

martincostello:
I think removing the strategy and changing to _"whatever our defaults are"_ or _"explicitly what you say to use"_ is OK, provided that there's clear docs as to _what_ the defaults are (which could also contain an example of how to create a custom convention using an extension method?)

src/JustSaying/Extensions/TypeExtensions.cs
```diff
@@ -1,20 +1,74 @@
...
+        private const int MaxTopicNameLength = 256;
+        private const int MaxQueueNameLength = 80;
 
-        public static string ToTopicName(this Type type)
+        private static readonly Dictionary<Type, string> TypeToFriendlyName = new Dictionary<Type, string>
         {
-            var name = type.GetTypeInfo().IsGenericType
-                ? Regex.Replace(type.FullName, "\\W", "_").ToLowerInvariant()
-                : type.Name.ToLowerInvariant();
+            {typeof(string), "string"},
```
martincostello: Could we use the new `[typeof(string)] = "string",` syntax instead?

src/JustSaying/Extensions/TypeExtensions.cs
```diff
@@ -1,20 +1,74 @@
...
+            {typeof(double), "double"},
+            {typeof(short), "short"},
+            {typeof(int), "int"},
+            {typeof(long), "long"},
+            {typeof(sbyte), "sbyte"},
+            {typeof(float), "float"},
+            {typeof(ushort), "ushort"},
+            {typeof(uint), "uint"},
+            {typeof(ulong), "ulong"},
+            {typeof(void), "void"}
```
martincostello: Just spit-balling some defaults, `TimeSpan`? `DateTime`? `DateTimeOffset`?

src/JustSaying/Extensions/TypeExtensions.cs
```diff
@@ -1,20 +1,74 @@
...
+            if (TypeToFriendlyName.TryGetValue(type, out string friendlyName))
+            {
+                return friendlyName;
+            }
+
+            friendlyName = type.Name;
+
+            if (type.GetTypeInfo().IsGenericType)
+            {
+                var backtick = friendlyName.IndexOf('`');
```
martincostello: `index` or `indexOfBacktick`?

src/JustSaying/Extensions/TypeExtensions.cs
adammorr: Added

src/JustSaying/Extensions/TypeExtensions.cs
adammorr: Changed to `HashSet<Type>` rather than a dictionary as it was only storing the lower case name against the type anyway

src/JustSaying/Extensions/TypeExtensions.cs
adammorr: Done

adammorr:
Next bit is adding the custom convention that I'm definitely not calling a naming strategy

adammorr:
OK so I've added `IDefaultQueueNamingConvention` and `IDefaultTopicNamingConvention` with default implementations that continue to do what is currently done in JustSaying.

I'm not sure I'm happy with the names but it's a start. I also figured the sensible place for this to be exposed is through `builder.Messaging(...)` as below;

```
return new ServiceCollection()
    .AddJustSaying(
        (builder, serviceProvider) =>
        {
            builder.Messaging((options) =>
                {
                    options.WithRegion(RegionName)
                        .WithDefaultTopicNamingConvention<MyDefaultTopicNamingConvention>()
                        .WithDefaultQueueNamingConvention(new MyDefaultQueueNamingConvention());
                })
                .Client((options) =>
                {
                    options.WithSessionCredentials(AccessKeyId, SecretAccessKey, SessionToken)
                        .WithServiceUri(ServiceUri);
                });

            configure(builder, serviceProvider);
        });
```

slang25:
NamingConvertion is spot on 👍

src/JustSaying/Naming/IDefaultQueueNamingConvention.cs
```diff
@@ -0,0 +1,7 @@
+namespace JustSaying.Naming
+{
+    public interface IDefaultQueueNamingConvention
```
slang25: Should this just be `IQueueNamingConvention`? (same for topic API)

adammorr: The reason for including default is so that it aligns with the comments on https://github.com/justeat/JustSaying/blob/master/src/JustSaying/Fluent/TopicSubscriptionBuilder%601.cs#L34 and https://github.com/justeat/JustSaying/blob/master/src/JustSaying/Fluent/QueueSubscriptionBuilder%601.cs#L34

Maybe the comments could be updated on the above to make it clear that these will be used?

slang25: Ah, got it. The name makes sense (I was confused again 😄 )

slang25: So this only kicks in where no string is passed in `ForTopic/ForQueue`

adammorr: Yeah, that's right. If you provide a string then that exact value is used

adammorr:
This is nearly there as far as changes I want to make now...the last thing is changing how the extension methods in `TypeExtensions` are used as they can probably go now

adammorr:
Right, no more changes for me to push for now 👍

adammorr:
@slang25 any other thoughts on this?

slang25:
Nope, love it 😃

adammorr:
@martincostello @maurofranchi do either of you two have any further thoughts on this change?

src/JustSaying/Fluent/MessagingConfigurationBuilder.cs
```diff
@@ -352,6 +363,68 @@ public MessagingConfigurationBuilder WithRegions(IEnumerable<RegionEndpoint> reg
...
+        /// Specifies the <see cref="IDefaultTopicNamingConvention"/> to use.
+        /// </summary>
+        /// <param name="namingConvention">The <see cref="IDefaultTopicNamingConvention"/> to use.</param>
+        /// <returns>
+        /// The current <see cref="MessagingConfigurationBuilder"/>.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="namingConvention"/> is <see cref="null"/>.
+        /// </exception>
+        public MessagingConfigurationBuilder WithDefaultTopicNamingConvention(IDefaultTopicNamingConvention namingConvention)
```
maurofranchi: I honestly find a bit weird that we have to pass a `IDefaultTopicNamigConvention` to this method that should impliciting using the _default topic naming convention_ 😕 ... unless _default_ = _custom_ (meaning there is one that calculates the naming based on a specific logic) which is confusing as well 😄

src/JustSaying/Fluent/MessagingConfigurationBuilder.cs
```diff
@@ -352,6 +363,68 @@ public MessagingConfigurationBuilder WithRegions(IEnumerable<RegionEndpoint> reg
...
+        /// Specifies the <see cref="IDefaultTopicNamingConvention"/> to use.
+        /// </summary>
+        /// <typeparam name="T">The <see cref="IDefaultTopicNamingConvention"> to use which will be resolved from the <see cref="IServiceResolver"/></typeparam>
+        /// <returns>
+        /// The current <see cref="MessagingConfigurationBuilder"/>.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="T"/> is resolved to <see cref="null"/>.
+        /// </exception>
+        public MessagingConfigurationBuilder WithDefaultTopicNamingConvention<T>() where T : class, IDefaultTopicNamingConvention
```
maurofranchi: as above

src/JustSaying/Fluent/MessagingConfigurationBuilder.cs
```diff
@@ -352,6 +363,68 @@ public MessagingConfigurationBuilder WithRegions(IEnumerable<RegionEndpoint> reg
...
+        /// Specifies the <see cref="IDefaultQueueNamingConvention"/> to use.
+        /// </summary>
+        /// <param name="namingConvention">The <see cref="IDefaultQueueNamingConvention"/> to use.</param>
+        /// <returns>
+        /// The current <see cref="MessagingConfigurationBuilder"/>.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="namingConvention"/> is <see cref="null"/>.
+        /// </exception>
+        public MessagingConfigurationBuilder WithDefaultQueueNamingConvention(IDefaultQueueNamingConvention namingConvention)
```
maurofranchi: same

src/JustSaying/Fluent/MessagingConfigurationBuilder.cs
```diff
@@ -352,6 +363,68 @@ public MessagingConfigurationBuilder WithRegions(IEnumerable<RegionEndpoint> reg
...
+        /// Specifies the <see cref="IDefaultQueueNamingConvention"/> to use.
+        /// </summary>
+        /// <typeparam name="T">The <see cref="IDefaultQueueNamingConvention"> to use which will be resolved from the <see cref="IServiceResolver"/></typeparam>
+        /// <returns>
+        /// The current <see cref="MessagingConfigurationBuilder"/>.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="T"/> is resolved to <see cref="null"/>.
+        /// </exception>
+        public MessagingConfigurationBuilder WithDefaultQueueNamingConvention<T>() where T : class, IDefaultQueueNamingConvention
```
maurofranchi: and again

martincostello:
I haven't looked at it yet, but I will.

src/JustSaying/Fluent/MessagingConfigurationBuilder.cs
adammorr: Yeah I can see the confusion - it is indeed custom but the intention is for it to be used whenever `ForTopic<T>()` or `ForQueue<T>()` are invoked without a queue name provided which lets the consumer know it's using the default...not sure how to relay that more clearly 🤔

adammorr: Or we just call it `ITopicNamingConvention` and `IQueueNamingConvention` as @slang25 mentioned before and change the comments accordingly? This feels a bit clearer and we can convey that if you don't provide a queue / topic name then the default implementation of either the two interfaces will be used

maurofranchi: yeah I would prefer that

src/JustSaying/Naming/DefaultNamingConventions.cs
```diff
@@ -0,0 +1,77 @@
...
+            typeof(ushort),
+            typeof(uint),
+            typeof(ulong),
+            typeof(void),
+            typeof(TimeSpan),
+            typeof(DateTime),
+            typeof(DateTimeOffset)
+        };
+
+        public string TopicName<T>() => CreateResourceName(typeof(T), MaxTopicNameLength);
```
martincostello: Maybe we could make these two methods `virtual` so you could create a derived class to override just one if you so desired, or if you wanted to override both to do something simple like add on a prefix or suffix?

src/JustSaying/Naming/IQueueNamingConvention.cs
```diff
@@ -0,0 +1,7 @@
+namespace JustSaying.Naming
+{
+    public interface IQueueNamingConvention
```
martincostello: As it's new, `///` comments please.

src/JustSaying/Naming/ITopicNamingConvention.cs
```diff
@@ -0,0 +1,7 @@
+namespace JustSaying.Naming
+{
+    public interface ITopicNamingConvention
```
martincostello: As it's new, `///` comments please.

PR #655, opened 2020-03-12
New channel-based consumption pipeline

gkinsman:
**TODO**

- [ ] Load test new channels implementation
- [ ] Work out of there's an obvious path to multi-region support, and ensure we don't wall it off for the future.
- [x] Do we need an API to allow per-queue concurrency limits? Perhaps a multiplexer-per-queue-group?

This PR implements a new consumer pipeline to replace `INotificationSubscriber`, with the goal being to allow us to throttle consumers and protect downstream resources. The new implementation is based on `System.Threading.Channels`.

Previously, `INotificationSubscriber` played all of the roles of the new interfaces below, so we tried splitting it up.

The new core interfaces are:

**IMessageReceiveBuffer**
Receives messages from SQS and buffers them into a target channel.

```csharp
public interface IMessageReceiveBuffer
{
    Task Run(CancellationToken stoppingToken);
    ChannelReader<IQueueMessageContext> Reader { get; }
}
```

**IMultiplexer**
Reduces multiple input channel of messages into a single channel, interleaving them.
```csharp
internal interface IMultiplexer
{
    Task Run(CancellationToken stoppingToken);
    void ReadFrom(ChannelReader<IQueueMessageContext> reader);
    IAsyncEnumerable<IQueueMessageContext> Messages();
}
```

**IMultiplexerSubscriber**
Subscribes to an `IAsyncEnumerable<IQueueMessageContext>` and sends them to an `IDispatcher`
```csharp
internal interface IMultiplexerSubscriber
{
    Task Run(CancellationToken stoppingToken);
    void Subscribe(IAsyncEnumerable<IQueueMessageContext> messageSource);
}
```

**IQueueMessageContext**
A queue context that contains the `Message`, `MessageAttributes`, and also deletion and visibility update API's.
```csharp
public interface IQueueMessageContext
{
    Message Message { get; }

    MessageAttributes MessageAttributes { get; }

    Task ChangeMessageVisibilityAsync(TimeSpan visibilityTimeout);

    Task DeleteMessageFromQueueAsync();

    Uri QueueUri { get; }

    string QueueName { get; }
}
```

**ISubscriptionGroup**
Represents a single group of receive buffers, multiplexer, and consumers. Each subscription group can be configured independently of each other, and so can be used to control throttling for a set of queues.
```csharp
public interface ISubscriptionGroup : IInterrogable
{
    Task Run(CancellationToken stoppingToken);
}
```

**ISubscriptionGroupCollection**
Represents a group of `ISubscriptionGroup` that can be started together. This is the top level runtime interface that is used to control all of the subscribers. It is created from an `ISubscriptionGroupFactory`.
```csharp
public interface ISubscriptionGroupCollection : IInterrogable
{
    Task Run(CancellationToken stoppingToken);
}
```

**ISubscriptionGroupFactory**
Given a set of defaults, and per-subscription-group settings, creates each subscription group and brings them together into an `ISubscriptionGroupCollection` that can be started and stopped as a whole.
```csharp 
public interface ISubscriptionGroupFactory
{
    SubscriptionGroupCollection Create(
        SubscriptionConfigBuilder defaults,
        IDictionary<string, SubscriptionGroupConfigBuilder> consumerGroupSettings);
}
```

_This PR is the work of Maisie Sadler and George Kinsman_


martincostello:
I'm just about to start reviewing this, but based on the commit history, I'd suggest that when this is merged it's maybe done as a squash commit.

gkinsman:
> I'm just about to start reviewing this, but based on the commit history, I'd suggest that when this is merged it's maybe done as a squash commit.

Thanks! That sounds fine, and agree that a squash commit will be best. The history is long and we won't want it all in the main history.

src/JustSaying/AwsTools/QueueCreation/SqsBasicConfiguration.cs
```diff
@@ -43,6 +43,10 @@ public virtual void Validate()
                 throw new ConfigurationErrorsException(
                     $"Invalid configuration. {nameof(DeliveryDelay)} must be between {JustSayingConstants.MinimumDeliveryDelay} and {JustSayingConstants.MaximumDeliveryDelay}.");
             }
+
+            ValidateCustom();
         }
+
+        protected virtual void ValidateCustom() {}
```
martincostello:
```suggestion
        protected virtual void OnValidate()
        {
            // Any additional validation logic derived types require
        }
```

src/JustSaying/IMessagingBus.cs
```diff
@@ -11,6 +12,6 @@ public interface IMessagingBus
         /// Starts the message bus.
         /// </summary>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> which will stop the bus when signalled.</param>
```
martincostello: Add `<returns>` documentation.

src/JustSaying/IMessagingBus.cs
```diff
@@ -11,6 +12,6 @@ public interface IMessagingBus
         /// Starts the message bus.
```
martincostello:
```suggestion
        /// Starts the message bus as an asynchronous operation.
```

src/JustSaying/JustSaying.csproj
```diff
@@ -8,8 +8,10 @@
   <ItemGroup>
     <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.3.100.1" />
     <PackageReference Include="AWSSDK.SQS" Version="3.3.100.1" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
     <PackageReference Include="System.Text.Json" Version="4.6.0" />
+    <PackageReference Include="System.Threading.Channels" Version="4.7.0" />
```
martincostello: Would an older version such as `4.5.0` be sufficient to provide the same functionality so we have a wider compatibility range?

src/JustSaying/JustSayingBus.cs
```diff
@@ -76,24 +107,17 @@ public void AddNotificationSubscriber(string region, INotificationSubscriber sub
                 // Just re-use existing subscriber instead.
                 return;
             }
-            subscribersForRegion[subscriber.Queue] = subscriber;
+            subscribersForRegion[queue.QueueName] = queue;
 
-            AddSubscribersToInterrogationResponse(subscriber);
+            // todo: this could work if we make the sqsqueue generic?
```
martincostello: Unresolved TODO.

src/JustSaying/JustSayingBus.cs
```diff
@@ -115,27 +139,43 @@ private void AddSubscribersToInterrogationResponse(INotificationSubscriberInterr
...
             if (cancellationToken.IsCancellationRequested)
             {
-                return;
+                return Task.CompletedTask;
+            }
+
+            if (ConsumerBus != null)
+            {
+                _log.LogWarning("Attempting to start an already running Bus");
+                return ConsumerBus.Run(CancellationToken.None);
```
martincostello: Why isn't `cancellationToken` passed through?

src/JustSaying/JustSayingBus.cs
```diff
@@ -115,27 +139,43 @@ private void AddSubscribersToInterrogationResponse(INotificationSubscriberInterr
...
-                    {
-                        if (queueSubscriber.Value.IsListening)
-                        {
-                            continue;
-                        }
-
-                        queueSubscriber.Value.Listen(cancellationToken);
-                    }
+                    _log.LogWarning("Attempting to start an already running Bus");
+                    return ConsumerBus.Run(CancellationToken.None);
```
martincostello: Why isn't `cancellationToken` passed through?

martincostello: Duplication of code block.

src/JustSaying/Messaging/Channels/ChannelConsumer.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            _messageSource = messageSource;
+            return this;
+        }
+
+        public async Task Run(CancellationToken stoppingToken)
+        {
+            await foreach (var messageContext in _messageSource.WithCancellation(stoppingToken))
+            {
+                stoppingToken.ThrowIfCancellationRequested();
```
martincostello: Open question with no right answer, just for discussion:

Should we throw the operation cancelled exception here, or just break the loop and return?

src/JustSaying/Messaging/Channels/ConsumerBus.cs
```diff
@@ -0,0 +1,101 @@
...
+            IList<ISqsQueue> queues,
+            IConsumerConfig consumerConfig,
+            IMessageDispatcher messageDispatcher,
+            IMessageMonitor monitor,
+            ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<ConsumerBus>();
+            _consumerConfig = consumerConfig;
+
+            _multiplexer = new RoundRobinQueueMultiplexer(
```
martincostello: Should we DI-inject the multiplexer (or add a factory abstraction to create them) rather than hard-code it?

src/JustSaying/Messaging/Channels/ConsumerBus.cs
```diff
@@ -0,0 +1,101 @@
...
+            lock (_startLock)
+            {
+                if (_started) return _completion;
+
+                _completion = RunImpl(stoppingToken);
+
+                _started = true;
+                return _completion;
+            }
+        }
```
martincostello: We could have one return here, rather than three.

```suggestion
            if (!_started)
            {
                lock (_startLock)
                {
                    if (!_started)
                    {
                        _completion = RunImpl(stoppingToken);
                        _started = true;
                    }
                }
            }

            return _completion;
        }
```

src/JustSaying/Messaging/Channels/ConsumerBus.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        private Task RunImpl(CancellationToken stoppingToken)
+        {
+            var numberOfConsumers = _consumers.Count;
+            _logger.LogInformation(
+                "Starting up consumer bus with {ConsumerCount} consumers and {DownloadBufferCount} downloaders",
```
martincostello:
```suggestion
                "Starting up consumer bus with {ConsumerCount} consumers and {DownloadBufferCount} download buffers.",
```

src/JustSaying/Messaging/Channels/ConsumerBus.cs
```diff
@@ -0,0 +1,101 @@
...
+            // start
+            var completionTasks = new List<Task>();
+
+            completionTasks.Add(_multiplexer.Run(stoppingToken));
+            completionTasks.AddRange(_consumers.Select(x => x.Run(stoppingToken)));
+            completionTasks.AddRange(_buffers.Select(x => x.Run(stoppingToken)));
+
+            _logger.LogInformation("Consumer bus successfully started");
+
+            return Task.WhenAll(completionTasks);
```
martincostello: What's the behaviour here if the CancellationToken is fired? Will it wait for them all the finish and then return, or will it throw at the first one cancelled and abandon any further waiting?

src/JustSaying/Messaging/Channels/ConsumerBus.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        private IMessageReceiveBuffer CreateBuffer(
+            ISqsQueue queue,
+            IMessageMonitor monitor,
+            ILoggerFactory loggerFactory)
+        {
+            var buffer = new MessageReceiveBuffer(
```
martincostello: Again, should we DI or factory here to not create a specific implementation?

src/JustSaying/Messaging/Channels/IConsumerBus.cs
```diff
@@ -0,0 +1,10 @@
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.Channels
+{
+    public interface IConsumerBus
```
martincostello: `///` docs

src/JustSaying/Messaging/Channels/IConsumerConfig.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Collections.Generic;
+using JustSaying.Messaging.Middleware;
+
+namespace JustSaying.Messaging.Channels
+{
+    public interface IConsumerConfig
```
martincostello: `///` docs.

src/JustSaying/Messaging/Channels/IMessageReceiveBuffer.cs
```diff
@@ -0,0 +1,12 @@
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.Channels
+{
+    public interface IMessageReceiveBuffer
```
martincostello: `///` docs.

src/JustSaying/Messaging/Channels/IMultiplexer.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Threading.Tasks;
+using JustSaying.Messaging.MessageHandling;
+
+namespace JustSaying.Messaging.Channels
+{
+    internal interface IMultiplexer
+    {
+        Task Run(CancellationToken stoppingToken);
+        void ReadFrom(ChannelReader<IQueueMessageContext> reader);
+        IAsyncEnumerable<IQueueMessageContext> Messages();
```
martincostello: `GetMessagesAsync()`?

src/JustSaying/Messaging/Channels/IQueueMessageContext.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Threading.Tasks;
+using Amazon.SQS.Model;
+
+namespace JustSaying.Messaging.Channels
+{
+    public interface IQueueMessageContext
```
martincostello: `///` docs.

src/JustSaying/Messaging/Channels/IQueueMessageContext.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        Message Message { get; }
+
+        Task ChangeMessageVisibilityAsync(int visibilityTimeoutSeconds);
+
+        Task DeleteMessageFromQueue();
```
martincostello:
```suggestion
        Task DeleteMessageFromQueueAsync();
```

src/JustSaying/Messaging/Channels/IQueueMessageContext.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+using Amazon.SQS.Model;
+
+namespace JustSaying.Messaging.Channels
+{
+    public interface IQueueMessageContext
+    {
+        Message Message { get; }
+
+        Task ChangeMessageVisibilityAsync(int visibilityTimeoutSeconds);
```
martincostello:
```suggestion
        Task ChangeMessageVisibilityAsync(TimeSpan visibilityTimeout);
```

src/JustSaying/AwsTools/MessageHandling/Dispatch/HandlerMap.cs
```diff
@@ -0,0 +1,49 @@
+using System;
```
slang25: Are we still able to subscribe the message message into multiple queues? I know this is something that was once in use, for example you might want different queues and handlers in an app that sent email/sms/push for a certain event.

samples/src/JustSaying.Sample.Restaurant.KitchenConsole/BusService.cs
```diff
@@ -1,7 +1,11 @@
+using System;
 using System.Threading;
 using System.Threading.Tasks;
+using JustSaying.Messaging;
+using JustSaying.Messaging.Interrogation;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Serilog;
```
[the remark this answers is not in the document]
gkinsman: Happy to go with whichever pattern you're happy with here. The idiomatic usage for Serilog is to not inject, and instead use `_logger = Log.ForContext<T>`, as its not really necessary to inject. 

In apps I usually always take a dependency directly on Serilog and use its API, leaving the MEL as a lowest common denominator _specification_ for libraries. The reason being that given MEL is a LCD, it doesn't have many of the niceties that Serilog provides that you might want to take advantage of in an app. Depending on the goal of this sample app, perhaps its desirable to show this off or not.

samples/src/JustSaying.Sample.Restaurant.KitchenConsole/appsettings.json
```diff
@@ -1,11 +1,4 @@
 {
   "AWSRegion": "eu-west-1",
-  "AWSServiceUrl": "http://localhost:4100",
```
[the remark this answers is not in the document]
gkinsman: Only meant for this to be empty string, as then it attempts to connect to real SQS instead of localstack. Will fix 👍 .

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+using JustSaying.Messaging.Channels;
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.Messaging.Middleware
+{
+    public class DefaultSqsMiddleware : MiddlewareBase<GetMessagesContext, IList<Amazon.SQS.Model.Message>>
+    {
+        private readonly ILogger<DefaultSqsMiddleware> _logger;
+
+        public DefaultSqsMiddleware(ILogger<DefaultSqsMiddleware> logger) : base(null)
```
martincostello: `base(null)` seems like a design smell of the base class?

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+
+                _logger.LogTrace(
+                    "Polled for messages on queue '{QueueName}' in region '{Region}', and received {MessageCount} messages.",
+                    context.QueueName,
+                    context.RegionName,
+                    context.Count);
+
+                return results;
+            }
+            catch (InvalidOperationException ex)
```
martincostello: How does this exception type indicate that the count of messages to handle failed?

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                _logger.LogTrace(
+                    ex,
+                    "Could not determine number of messages to read from queue '{QueueName}' in '{Region}'.",
+                    context.QueueName,
+                    context.RegionName);
+            }
+            catch (OperationCanceledException ex)
+            {
+                _logger.LogTrace(
```
martincostello: Maybe this should be a warning?

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+                    ex,
+                    "Suspected no message on queue '{QueueName}' in region '{Region}'.",
```
martincostello: Maybe put something about a timeout in the message to give a clue to why this suspicion?

src/JustSaying/Messaging/Middleware/MiddlewareBuilder.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+
+namespace JustSaying.Messaging.Middleware
+{
+    public static class MiddlewareBuilder
+    {
+        public static MiddlewareBase<TContext, TOut> BuildAsync<TContext, TOut>(params Func<MiddlewareBase<TContext, TOut>, MiddlewareBase<TContext, TOut>>[] policies)
```
martincostello: `middlewares` rather than `policies`?

src/JustSaying/Messaging/Middleware/MiddlewareBuilder.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            MiddlewareBase<TContext, TOut> policy = new NoopMiddleware<TContext, TOut>();
+            return policy.WithAsync(policies);
+        }
+
+        public static MiddlewareBase<TIn, TOut> WithAsync<TIn, TOut>(this MiddlewareBase<TIn, TOut> inner, params Func<MiddlewareBase<TIn, TOut>, MiddlewareBase<TIn, TOut>>[] policies)
```
martincostello: As above.

src/JustSaying/Messaging/Middleware/Middleware`2.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.Middleware
+{
+    public abstract class MiddlewareBase<TContext, TOut>
```
martincostello: `MessageMiddleware` or `HandlerMiddleware`? `MiddlewareBase` just seems a bit too vague to me.

src/JustSaying/Messaging/Middleware/Middleware`2.cs
```diff
@@ -0,0 +1,32 @@
+using System;
```
martincostello: File name and class name don't match. If the class isn't renamed, should be `MiddlewareBase(backtick)2.cs`.

src/JustSaying/Messaging/Middleware/NoopMiddleware`2.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Threading.Tasks;
+
+namespace JustSaying.Messaging.Middleware
+{
+    internal class NoopMiddleware<TContext, TOut> : MiddlewareBase<TContext, TOut>
```
martincostello: Suggest based on the `NullLoggerFactory` and `NullLogger` convention of Microsoft's, we call this `NullMiddleware`?

src/JustSaying/Messaging/Monitoring/Operation.cs
```diff
@@ -0,0 +1,26 @@
...
+            _messageMonitor = messageMonitor ?? throw new ArgumentNullException(nameof(messageMonitor));
+            _onComplete = onComplete ?? throw new ArgumentNullException(nameof(onComplete));
+
+            _stopWatch = new Stopwatch();
+            _stopWatch.Start();
+        }
+
+        public void Dispose()
+        {
+            _onComplete(_stopWatch.Elapsed, _messageMonitor);
```
martincostello: Should the stopwatch be being stopped first? Leaving it running implies it would support `Dispose()` being called multiple times, but then that's a smell of the `IDisposable` pattern as it's not intended to be used that way.

tests/JustSaying.UnitTests/JustSaying.UnitTests.csproj
```diff
@@ -12,8 +12,10 @@
   <ItemGroup>
     <PackageReference Include="AWSSDK.SimpleNotificationService" Version="$(AwsSdkSnsVersion)" />
     <PackageReference Include="AWSSDK.SQS" Version="$(AwsSdkSqsVersion)" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
```
martincostello: I think you can remove this as it's a transient dependency from `JustSaying.csproj`.

tests/JustSaying.UnitTests/JustSayingBus/WhenRegisteringSubscribers.cs
```diff
@@ -1,60 +1,77 @@
...
-        [Fact]
-        public void CallingStartTwiceDoesNotStartListeningTwice()
-        {
-            _subscriber1.IsListening.Returns(true);
-            _subscriber2.IsListening.Returns(true);
-            SystemUnderTest.Start();
-            _subscriber1.Received(1).Listen(default);
-            _subscriber2.Received(1).Listen(default);
-        }
+        // todo: how can we check this?
```
martincostello: Better to uncomment and use `[Fact(Skip = "the reason")]`.

martincostello:
Just the new tests I haven't looked at now. I'll leave them until we're all more or less happy with the library code.

samples/src/JustSaying.Sample.Restaurant.KitchenConsole/BusService.cs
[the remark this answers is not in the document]
martincostello: I think MEL is nicer to use here for the sample, as then we're not potentially adding extra new things into the cognitive load of a "newbie" looking at them.

samples/src/JustSaying.Sample.Restaurant.OrderingApi/JustSaying.Sample.Restaurant.OrderingApi.csproj
```diff
@@ -9,6 +9,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Serilog.AspNetCore" Version="3.2.1-dev-00142" />
+    <PackageReference Include="Serilog.Sinks.Seq" Version="4.0.0" />
```
JosephWoodward: Did you mean to include this package, or was this for local debugging?

src/JustSaying/Messaging/Channels/MessageReceiveBuffer.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// <param name="stoppingToken"></param>
+        /// <returns>A task that completes when the cancellation token is fired.</returns>
+        public async Task Run(CancellationToken stoppingToken)
+        {
+            await Task.Yield();
+
+            ChannelWriter<IQueueMessageContext> writer = _channel.Writer;
+            try
+            {
+                while (true)
```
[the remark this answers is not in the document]
slang25: I raised this question in the last PR, but reversed. I think it's advised to use throw rather than gracefully exiting. I prefer it for consistency, otherwise it depends on the timing of when you cancel as to how it behaves.
Would appreciate your view.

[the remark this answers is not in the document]
gkinsman: The tests are rather nicely consistent now in that we _always_ throw when cancellation is fired, and we assert that is the case.

slang25:
Can `IQueueMessageContext` be made internal?
I wouldn't want `DeleteMessageFromQueue` or `ChangeMessageVisibilityAsync` exposed.

src/JustSaying/Messaging/Channels/MessageReceiveBuffer.cs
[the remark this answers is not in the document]
martincostello: I'm just wary of the `while (true)` construct in general.

samples/src/JustSaying.Sample.Restaurant.OrderingApi/JustSaying.Sample.Restaurant.OrderingApi.csproj
gkinsman: Yes, have been using seq to analyse the behaviour of the new consumer bits. Previously the internal library code used SerilogTimings, but I'm going to pull that out to an `IMessageMonitor` now that we support that.

Seq dashboards via structured logging are an excellent way to get metrics on consumption rates/handler durations/receive timings without depending on something like statsd.

gkinsman: I'd argue this should be left in, because the behaviour of a non-existent server at this address is just to backoff expontentially, it becomes a no-op, and the value is auto-sending logs in a local dev environment.

 I'm sure you're aware of all this 🙂

samples/src/JustSaying.Sample.Restaurant.KitchenConsole/Program.cs
```diff
@@ -1,83 +1,109 @@
...
+using Serilog;
 
 namespace JustSaying.Sample.Restaurant.KitchenConsole
 {
     internal class Program
     {
         public static async Task Main()
         {
+            Log.Logger = new LoggerConfiguration()
+                .WriteTo.Seq("http://localhost:5341")
```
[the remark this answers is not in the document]
gkinsman: Have some thoughts here: https://github.com/justeat/JustSaying/pull/655#discussion_r392254600. 

It would be great to have it for the next testing phase too, as it provides a really quick feedback cycle for running load test scenarios and seeing them graphed in Seq.

src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs
```diff
@@ -57,17 +54,17 @@ public async Task DispatchMessage(SQSMessage message, CancellationToken cancella
...
                     "Could not handle message with Id '{MessageId}' because a deserializer for the content is not configured. Message body: '{MessageBody}'.",
-                    message.MessageId,
-                    message.Body);
+                    messageContext.Message.MessageId,
+                    messageContext.Message.Body);
 
-                await DeleteMessageFromQueue(message.ReceiptHandle).ConfigureAwait(false);
-                _onError(ex, message);
+                await messageContext.DeleteMessageFromQueue().ConfigureAwait(false);
+                _onError(ex, messageContext.Message);
```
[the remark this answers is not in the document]
gkinsman: We discussed this today - it looks like the onError callback is only used for logging. I've been through 3 of the 9 or so pages of search so far (will finish searching) and they all just log the error. If this is all it's used for (what else?!), then we should move it to `IMessageMonitor` as part of v7.

samples/src/JustSaying.Sample.Restaurant.KitchenConsole/appsettings.json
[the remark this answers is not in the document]
gkinsman: Hmm, this diff is a bit confusing as it doesn't show that that line is now:

` "AWSServiceUrl": "",`

If you meant to question that, it's because this sample didn't work with my localstack, and I wanted to test it against real SQS. Can you think of a way we can leave this setting while devving against real SQS?

src/JustSaying/Messaging/Channels/ConsumerBus.cs
```diff
@@ -65,7 +65,7 @@ private Task RunImpl(CancellationToken stoppingToken)
         {
             var numberOfConsumers = _consumers.Count;
             _logger.LogInformation(
-                "Starting up consumer bus with {ConsumerCount} consumers and {DownloadBufferCount} downloaders",
+                "Starting up consumer bus with {ConsumerCount} consumers and {ReceiveBuffferCount} receivers",
```
martincostello:
```suggestion
                "Starting up consumer bus with {ConsumerCount} consumers and {ReceiveBuffferCount} receivers.",
```

src/JustSaying/Messaging/Channels/IMultiplexer.cs
```diff
@@ -7,6 +7,9 @@
 
 namespace JustSaying.Messaging.Channels
 {
+    /// <summary>
+    /// Reduces multiple input streams of messages into a single stream, interleaving them.
```
martincostello: Would be good on the members too, not just the interface itself.

src/JustSaying/AwsTools/MessageHandling/Dispatch/HandlerMap.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public void Add(Type messageType, HandlerFunc handlerFunc)
+        {
+            _handlers.Add(messageType, handlerFunc);
+        }
+
+        public HandlerFunc Get(Type messageType)
+        {
+            HandlerFunc handler;
+            return _handlers.TryGetValue(messageType, out handler) ? handler : null;
```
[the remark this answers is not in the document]
gkinsman: Hmm, looks like we currently succeed the handler if none is found:
```csharp
var handler = _handlerMap.Get(messageType);

if (handler == null)
{
    return true;
}
```

Is this really the expected behaviour? 🙁

src/JustSaying/JustSaying.csproj
gkinsman: I don't think an older version provides us with wider compatibility? This package on nuget supports any target framework we'd ever want to support.

gkinsman: In fact, Channels 4.5.0 doesn't support `.NETCoreApp 3.0`

src/JustSaying/JustSayingBus.cs
gkinsman: This is relying on the fact that calling `ConsumerBus.Run` multiple times will always return the same task, which perhaps isn't very clear. Only the first time will it set the cancellation token, which happens below.

I'm changing this to be explicit about the single-start safety by keeping a `Task _consumerCompletionTask` that we return on multiple-entries to this bus.

src/JustSaying/JustSayingBus.cs
gkinsman: Same reason as above

gkinsman: I think I've resolved this in a satisfactory way, take a look and see if you're happy with it.

src/JustSaying/Messaging/Channels/ConsumerBus.cs
gkinsman: Currently this task will return when any of the tokens are cancelled, because an exception triggers an immediate return. We perhaps want to wrap each of the completion tasks in a handler that handles an operation cancelled and completes the task without throwing.

If we did this, would we still want to throw a cancelled exception when all of these tasks have completed/cancelled?

src/JustSaying/AwsTools/MessageHandling/Dispatch/HandlerMap.cs
gkinsman: This is a good question, and might require @martincostello's input. There's a few concerns here, subscriptions to SQS/SNS, and actual `IHandlerAsync` registrations.

### Subscriptions ###
The new builder API subscribes message types to topic/queue names by convention, which is what JustSaying uses to register subscriptions/topics etc. From this perspective it's possible to subscribe to the same message to multiple queues using a custom naming convention.

### IHandlerAsync ###

From a handler perspective, there are two new API's:
`AddJustSayingHandler<TMessage,THandler>` and
`AddJustSayingHandlers(IEnumerable<IHandlerAsync<TMessage, THandler>>`

The first does as you'd expect, and registers a single handler with a transient lifetime.
The second wraps up the multiple handlers into a `ListHandler` which is basically a `Task.WhenAll` around the tasks returned by the handlers.

The important bit here is that the ListHandler is itself an `IHandlerAsync`, so it's still registered into DI as a single handler, it just contains multiple handlers. At resolution time, if there is more than one handler registered to a particular message type in the DI container it will throw.

This means that it's not possible to register a multiple handlers for the same message type, and have them execute in different queue scopes. I suspect this might be a missing feature. In the existing world, I think this is possible as `HandlerMap` is scoped per queue instead of globally?

### Possible Solution ###
MassTransit has a concept called [`receive endpoints`](https://masstransit-project.com/usage/consumers.html#consumer) which is a named group of message types that are sent to the same queue. As registration for message type handling and handlers happens on a per-endpoint basis, it's possible to specify specific handlers for specific endpoints.

Do we need something similar? Or do we just need to be able to somehow explicitly state which handler should handle a particular queue's message? Not sure how this would work by convention.

src/JustSaying/Messaging/Channels/RoundRobinQueueMultiplexer.cs
```diff
@@ -0,0 +1,152 @@
...
+                return Completion;
+            }
+        }
+
+        private async Task RunImpl()
+        {
+            await Task.Yield();
+
+            _logger.LogInformation(
+                "Starting up channel multiplexer with a queue capacity of {Capacity}",
```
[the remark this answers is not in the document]
gkinsman: I generally follow Serilog's template recommendations, which suggests that message templates are fragments and not sentences and shouldn't have trailing full stops when possible: https://github.com/serilog/serilog/wiki/Writing-Log-Events#message-template-recommendations

src/JustSaying/Messaging/Channels/ChannelConsumer.cs
gkinsman: I think I prefer this strategy for the consistency it gives us in cancellation (always throws, and we can test for it), but I'm not totally attached to how this works currently.

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
maisiesadler: All of this logic is copy-pasta'd from [here](https://github.com/justeat/JustSaying/blob/25b5d8dbe3697b227168242abd2ada82a57bd88a/src/JustSaying/AwsTools/MessageHandling/SqsNotificationListener.cs#L148)

maisiesadler: We might need to review what errors we are actually throwing now and whether these are still expected in these scenarios

src/JustSaying/Messaging/Middleware/NoopMiddleware`2.cs
maisiesadler: This was originally `InnerMiddleware` as it's just a wrapper around the actual operation.
We changed it because it sounded like something internal and originally it was public.

src/JustSaying/Messaging/Middleware/Middleware`2.cs
gkinsman: This is a base middleware type that can be implemented to be used in any of these situations, so I think the vagarity is warranted

src/JustSaying/Messaging/Monitoring/Operation.cs
gkinsman: I've made `MonitorExtensions` internal as Operation is only used there to provide a nice interface for wrapping timed operations. Making it internal hides the ability to dispose from users, so we don't need to worry.

src/JustSaying/AwsTools/MessageHandling/Dispatch/HandlerMap.cs
martincostello: I guess this is a wider question than just this PR? Maybe split it out into an issue and we can discuss whether or not we want to change the multiple-handler handling for v7 or not (I suspect not to limit scope).

src/JustSaying/Messaging/Middleware/Middleware`2.cs
martincostello: But it doesn't say _what_ it's a middleware of. In a .NET context, it makes me think of HTTP pipeline middlewares, which this isn't.

src/JustSaying/Messaging/Channels/MessageReceiveBuffer.cs
```diff
@@ -50,8 +50,8 @@ internal class MessageReceiveBuffer : IMessageReceiveBuffer
         /// <summary>
         /// Starts the receive buffer until it's cancelled by the stopping token.
         /// </summary>
-        /// <param name="stoppingToken"></param>
-        /// <returns>A task that completes when the cancellation token is fired.</returns>
+        /// <param name="stoppingToken">A CancellationToken token that will stop the buffer when fired</param>
```
martincostello:
```suggestion
        /// <param name="stoppingToken">A <see cref="CancellationToken"/> that will stop the buffer when signalled.</param>
```

src/JustSaying/Messaging/Channels/MessageReceiveBuffer.cs
```diff
@@ -50,8 +50,8 @@ internal class MessageReceiveBuffer : IMessageReceiveBuffer
...
+        /// <returns>A task that throws an `OperationCancelledException` when the cancellation token is fired.</returns>
```
martincostello:
```suggestion
        /// <returns>A <see cref="Task{TResult}"/> representing the asynchronous operation to receive the messages.</returns>
```

src/JustSaying/Messaging/Channels/MessageReceiveBuffer.cs
```diff
@@ -141,18 +141,17 @@ await _sqsQueue
         {
             while (!stoppingToken.IsCancellationRequested)
             {
+                using var timeoutToken = new CancellationTokenSource(TimeSpan.FromSeconds(2));
```
martincostello: Configurable?

src/JustSaying/Messaging/Channels/RoundRobinQueueMultiplexer.cs
```diff
@@ -128,9 +127,10 @@ private async Task RunImpl()
             }
         }
 
-        public async IAsyncEnumerable<IQueueMessageContext> Messages()
+        public async IAsyncEnumerable<IQueueMessageContext> GetMessagesAsync()
         {
-            if (!_started) throw new InvalidOperationException("");
+            if (!_started) throw new InvalidOperationException(
+                "Multiplexer must be started before listening to messages");
```
martincostello: I'm not a fan of eliding braces (but won't ask you to change them all) as I think it's less readable and creates more churn when code needs to change, but this one is confusing spreading over more than one line.

```suggestion
            if (!_started)
            {
                throw new InvalidOperationException("Multiplexer must be started before listening to messages");
            }
```

src/JustSaying/Messaging/Middleware/NoopMiddleware`2.cs
martincostello: I guess I'm just not a fan of `Noop`, as it's technically _"No-Op"_, which would be `NoOp`, and that looks equally weird to me.

src/JustSaying/AwsTools/MessageHandling/Dispatch/HandlerMap.cs
slang25: Seems reasonable. Just to be clear though, I think that this PR is removing support for handling multiple queues with the same message type. I know that at some point this feature was being used, which is why I raise it.

src/JustSaying/Messaging/Channels/RoundRobinQueueMultiplexer.cs
gkinsman: No problem here, will add a rule to editorconfig to fix on format. I think in library scenarios its ok to enforce always using braces.

src/JustSaying/AwsTools/MessageHandling/Dispatch/HandlerMap.cs
martincostello: Ah, I didn't get that nuance. I thought from George's reply the existing wrapper-ing meant it was fine.

gkinsman: I think we should be able to support this. We originally changed it because the HandlerMap had a lot of details that leaked into the SqsNotificationListener, so making it stateless seemed like an obvious move.

We have some further change coming along related to grouping queues to allows control over prefetch/multiplexer-size/queue size on a per-(queue/group) basis instead of all or nothing, so let's re-open this when that lands.

src/JustSaying/Messaging/Channels/Receive/IReceiveBufferFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.Messaging.Channels.Receive
+{
+    internal interface IReceiveBufferFactory
+    {
+        IMessageReceiveBuffer CreateBuffer(ISqsQueue queue, ConsumerGroupSettings settings);
+    }
+
+    internal class ReceiveBufferFactory : IReceiveBufferFactory
```
martincostello: Move to own file.

src/JustSaying/Messaging/Channels/ConsumerGroups/SingleConsumerGroupFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+            ILoggerFactory loggerFactory)
+        {
+            _multiplexerFactory = multiplexerFactory ??
+                                  throw new ArgumentNullException(nameof(multiplexerFactory));
+            _receiveBufferFactory = receiveBufferFactory ??
+                                    throw new ArgumentNullException(nameof(receiveBufferFactory));
+            _channelConsumerFactory = channelConsumerFactory ??
+                                        throw new ArgumentNullException(nameof(channelConsumerFactory));
+            _loggerFactory = loggerFactory ??
+                             throw new ArgumentNullException(nameof(loggerFactory));
```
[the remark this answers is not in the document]
gkinsman: agree, this is awful and I can't find a combination of R#/Rider formatting that makes it nice :/

src/JustSaying/JustSayingBus.cs
```diff
@@ -73,27 +81,33 @@ public JustSayingBus(IMessagingConfig config, IMessageSerializationRegister seri
             Monitor = new NullOpMessageMonitor();
             MessageContextAccessor = new MessageContextAccessor();
 
-            _subscribersByRegionAndQueue = new Dictionary<string, Dictionary<string, ISqsQueue>>();
             _publishersByRegionAndType = new Dictionary<string, Dictionary<Type, IMessagePublisher>>();
             SerializationRegister = serializationRegister;
             _publishers = new HashSet<IPublisher>();
 
-            _sqsQueues = new List<ISqsQueue>();
+            _consumerGroupSettings = new Dictionary<string, ConsumerGroupSettings>();
```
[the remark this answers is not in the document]
gkinsman: Yeah, the key here is `typeof(T).FullName` by default, but you could group them together into arbitrary groups. We want to be case sensitive here I think.

src/JustSaying/JustSayingBus.cs
```diff
@@ -148,13 +164,21 @@ private Task RunImpl(CancellationToken stoppingToken)
                 _messageBackoffStrategy,
                 MessageContextAccessor);
 
-            ConsumerBus = new ConsumerBus(
-                _sqsQueues,
-                Config.ConsumerConfig,
-                dispatcher,
-                Monitor,
+            var receiveBufferFactory = new ReceiveBufferFactory(_loggerFactory, Config.ConsumerGroupConfig, Monitor);
```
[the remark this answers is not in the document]
gkinsman: Yep, that's the next step. We've now pulled creation up to here so should be pretty straightforward to pull them from DI.

src/JustSaying/Messaging/Channels/ConsumerGroups/CombinedConsumerGroup.cs
```diff
@@ -0,0 +1,56 @@
...
+            if (_started) return _completion;
+            lock (_startLock)
+            {
+                if (_started) return _completion;
+
+                _completion = RunImpl(stoppingToken);
+
+                _started = true;
+                return _completion;
+            }
```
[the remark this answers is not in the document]
gkinsman: Have made this change 👍

src/JustSaying/Messaging/Channels/ConsumerGroups/SingleConsumerGroup.cs
```diff
@@ -0,0 +1,47 @@
...
+            _logger.LogInformation(
+                "Starting up consumer bus with {ConsumerCount} consumers and {ReceiveBuffferCount} receive buffers",
+                _consumers.Count,
+                _receiveBuffers.Count);
+
+            IEnumerable<Task> bufferTasks = _receiveBuffers.Select(buffer => buffer.Run(stoppingToken));
+            Task multiplexerTask = _multiplexer.Run(stoppingToken);
+            IEnumerable<Task> consumerTasks = _consumers.Select(consumer => consumer.Run(stoppingToken));
+
+            IEnumerable<Task> allTasks = bufferTasks.Concat(consumerTasks).Concat(new[] { multiplexerTask });
```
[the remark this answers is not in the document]
gkinsman: Nice, have fixed 👍

src/JustSaying/Messaging/Channels/Dispatch/ChannelConsumer.cs
```diff
@@ -15,20 +16,19 @@ public ChannelConsumer(IMessageDispatcher dispatcher)
             _dispatcher = dispatcher;
         }
 
-        public IChannelConsumer ConsumeFrom(IAsyncEnumerable<IQueueMessageContext> messageSource)
+        public void DispatchFrom(IAsyncEnumerable<IQueueMessageContext> messageSource)
```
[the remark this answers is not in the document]
gkinsman: There's no need for a fluent API here, and it's internal and only created by us. I think it was probably a mistake to make it fluent originally.

src/JustSaying/JustSayingBus.cs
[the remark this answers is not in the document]
martincostello: I'd suggest explicitly passing `StringComparer.OrderinalIgnoreCase` into the dictionary constructor to be explicit then.

src/JustSaying/JustSayingBus.cs
```diff
@@ -98,10 +99,14 @@ public void AddQueue(string region, string consumerGroup, ISqsQueue queue)
...
+            if (!_consumerGroupSettings.TryGetValue(consumerGroup, out ConsumerGroupSettingsBuilder consumerGroupSettings))
             {
-                consumerGroupSettings = Config.ConsumerGroupConfig.CreateConsumerGroupSettings();
+                ConsumerGroupConfig defaultSettings = Config.ConsumerGroupConfig;
+                consumerGroupSettings = new ConsumerGroupSettingsBuilder()
+                    .WithPrefetch(defaultSettings.DefaultPrefetch)
+                    .WithBufferSize(defaultSettings.DefaultBufferSize)
+                    .WithConsumerCount(defaultSettings.DefaultConsumerCount)
+                    .WithMultiplexerCapacity(defaultSettings.DefaultMultiplexerCapacity);
                 _consumerGroupSettings[consumerGroup] = consumerGroupSettings;
```
martincostello: What about this?

```suggestion
                _consumerGroupSettings[consumerGroup] = new ConsumerGroupSettingsBuilder()
                    .WithPrefetch(defaultSettings.DefaultPrefetch)
                    .WithBufferSize(defaultSettings.DefaultBufferSize)
                    .WithConsumerCount(defaultSettings.DefaultConsumerCount)
                    .WithMultiplexerCapacity(defaultSettings.DefaultMultiplexerCapacity);
```

src/JustSaying/Messaging/Channels/ConsumerGroups/ConsumerGroupSettingsBuilder.cs
```diff
@@ -0,0 +1,70 @@
...
+            return this;
+        }
+
+        public ConsumerGroupSettingsBuilder AddQueues(IEnumerable<ISqsQueue> sqsQueues)
+        {
+            _sqsQueues.AddRange(sqsQueues);
+            return this;
+        }
+
+        private int _consumerCount;
```
martincostello: I'd prefer it if we group all the fields at the top like the rest of the code.

src/JustSaying/Messaging/Channels/Configuration/ConsumerGroupConfig.cs
```diff
@@ -0,0 +1,57 @@
...
+            DefaultPrefetch = 10;
+            DefaultConsumerCount = Environment.ProcessorCount * MessageConstants.ParallelHandlerExecutionPerCore;
+        }
+
+        public int DefaultPrefetch { get; set; }
+        public int DefaultBufferSize { get; set; }
+        public int DefaultConsumerCount { get; set; }
+        public int DefaultMultiplexerCapacity { get; set; }
+
+        public ConsumerGroupSettings CreateConsumerGroupSettings(IList<ISqsQueue> sqsQueues = null)
```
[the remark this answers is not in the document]
gkinsman: This has moved elsewhere and become ICollection

src/JustSaying/Messaging/Channels/ConsumerGroups/CombinedConsumerGroup.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        private Task RunImpl(CancellationToken stoppingToken)
+        {
+            IEnumerable<Task> completionTasks = _buses.Select(bus => bus.Run(stoppingToken));
+
+            _logger.LogInformation("Consumer bus successfully started");
```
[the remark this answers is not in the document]
gkinsman: Fixed

src/JustSaying/JustSaying.csproj
martincostello: Users of `netcoreapp3.0` would get an upgrade to the later version from the framework itself, so they'd get a newer version with optimisations.

src/JustSaying/JustSaying.csproj
```diff
@@ -9,7 +9,7 @@
     <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.3.100.1" />
     <PackageReference Include="AWSSDK.SQS" Version="3.3.100.1" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.2" />
```
martincostello: Revert. This would block apps on 2.1 LTS from using JustSaying.

src/JustSaying/JustSayingBus.cs
```diff
@@ -138,23 +135,23 @@ public void AddQueue(string region, string consumerGroup, ISqsQueue queue)
...
+        private Task _subscriberCompletionTask;
+        private bool _subscriberStarted;
 
         public Task Start(CancellationToken stoppingToken)
         {
             if (stoppingToken.IsCancellationRequested) return Task.CompletedTask;
 
             // Double check lock to ensure single-start
-            if (_consumerStarted) return _consumerCompletionTask;
+            if (_subscriberStarted) return _subscriberCompletionTask;
```
martincostello: This block could do with the same refactor as the other code that reduced the number of duplicated `return` statements.

src/JustSaying/Messaging/Channels/SubscriptionGroups/ISubscriptionGroup.cs
```diff
@@ -1,10 +1,11 @@
...
-namespace JustSaying.Messaging.Channels.ConsumerGroups
+namespace JustSaying.Messaging.Channels.SubscriptionGroups
 {
-    public interface IConsumerGroup
+    public interface ISubscriptionGroup
     {
         Task Run(CancellationToken stoppingToken);
     }
 }
+
```
martincostello: Nit: delete

src/JustSaying/JustSaying.csproj
gkinsman: But this version targets netstandard2.0 which 2.1 LTS supports?

martincostello: It's part of the 3.x major release, and to be honest I don't know what the consequences would be.

It's a library of interfaces, so there's no need to update it to the latest version in a library. If we leave it as it is, then there's no potential for issues.

src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs
```diff
@@ -29,13 +30,15 @@ internal class MessageDispatcher : IMessageDispatcher
         public MessageDispatcher(
             IMessageSerializationRegister serializationRegister,
             IMessageMonitor messagingMonitor,
+            Action<Exception, SQSMessage> onError,
```
martincostello: Disadvantage of this it that it doesn't fit well with DI.

tests/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
```diff
@@ -96,9 +97,8 @@ private MessageDispatcher CreateSystemUnderTestAsync()
         {
             var dispatcher = new MessageDispatcher(
                 _serializationRegister,
-                _messageMonitor,
-                _handlerMap,
-                _loggerFactory,
+                _messageMonitor, _onError,
+                _handlerMap, _loggerFactory,
```
martincostello:
```suggestion
                _messageMonitor,
                _onError,
                _handlerMap,
                _loggerFactory,
```

src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs
gkinsman: You might be looking at the wrong PR here, I accidentally pushed it here but its separate for now to get thoughts: https://github.com/gkinsman/JustSaying/pull/32

src/JustSaying/AwsTools/MessageHandling/Dispatch/HandlerMap.cs
gkinsman: This is now supported, thanks to @maisiesadler

src/JustSaying/Messaging/Middleware/Middleware`2.cs
gkinsman: That's a fair point I think - would something like `MessagingMiddlewareBase` work better? 
here are the current implementations:
`DefaultSqsMiddleware`: Middleware for the receive pipeline, used by `IMessageReceiveBuffer`
`DelegateMiddleware`: Used to wrap new middleware
`ErrorHandlingMiddleware`: An example of using a middleware to handle errors in the receive pipeline
`PollyMiddleware`: An example of a policy that enables the use of a Polly `IAsyncPolicy`.

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
gkinsman: This exception is no longer thrown by the new message receive buffer, so we'll remove this exception handler.

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
gkinsman: `OperationCanceledException`'s thrown here will always be caused by a receive timeout from SQS, as its only handled in the context of the middleware.

src/JustSaying/Messaging/Middleware/DefaultSqsMiddleware.cs
gkinsman: Updated the message to better explain what's happening.

PR #727, opened 2020-07-02
Fix bug identified in diagnosis of production incident

shaynevanasperen:
* Log when we don't have a matching handler for the given message type.
* If two consumers are subscribed to a queue that has multiple message types in it, one of them may not be able to handle all the messages (each handled a subset). So we need to return the message to the queue for the other subscriber to try handling it.

### DIFF
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
@@ -90,12 +90,7 @@ namespace JustSaying.UnitTests.AwsTools.MessageHandling.MessageDispatcherTests
             [Fact]
             public void ShouldLogError()
             {
-                _logger.Received().Log(
-                    LogLevel.Error,
-                    0,
-                    Arg.Is<FormattedLogValues>(x => x.ToString() == "Handler for message of type 'JustSaying.TestingFramework.OrderAccepted' not found in handler map. Returning message to queue."),
-                    null,
-                    Arg.Any<Func<object, Exception, string>>());
+                _logger.ReceivedWithAnyArgs().LogError(0, null, "msg");
             }
 
             [Fact]
--- a/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
+++ b/JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
@@ -10,6 +10,7 @@ using JustSaying.Messaging.MessageSerialisation;
 using JustSaying.Messaging.Monitoring;
 using JustSaying.TestingFramework;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Internal;
 using Newtonsoft.Json;
 using NSubstitute;
 using NSubstitute.ExceptionExtensions;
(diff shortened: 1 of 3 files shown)


JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -108,7 +108,9 @@ private async Task<bool> CallMessageHandler(Message message)
 
             if (handler == null)
             {
-                return true;
+                _log.LogError($"Handler for message of type '{message.GetType().FullName}' not found in handler map. Returning message to queue.");
```
martincostello, line 111: I'd suggest using structured logging, rather than string interpolation.

```suggestion
                _log.LogError("Handler for message of type '{MessageTypeName}' not found in handler map. Returning message to queue.", message.GetType().FullName);
```

shaynevanasperen:
@martincostello 
> Needs a version bump for the v6 branch:

Thanks. Done 👍

martincostello:
Also need to make sure we don't forget to put this fix into the default branch as well.

JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
```diff
@@ -68,6 +69,42 @@ protected override MessageDispatcher CreateSystemUnderTest()
...
+            [Fact]
+            public void ShouldDeserializeMessage()
+            {
+                _serialisationRegister.Received(1).DeserializeMessage(Arg.Is<string>(x => x == _sqsMessage.Body));
+            }
+
+            [Fact]
+            public void ShouldLogError()
+            {
+                _logger.Received().Log(
```
gkinsman, line 93: I don't think we should be asserting logging at this level, because it's brittle and breaks tests with changes that shouldn't break tests.

It should be enough that the handler returns false and that we update the visibility, _maybe_ that _an_ error is logged, but not the whole log message.

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -108,7 +108,9 @@ private async Task<bool> CallMessageHandler(Message message)
 
             if (handler == null)
             {
-                return true;
+                _log.LogError("Handler for message of type '{MessageTypeName}' not found in handler map. Returning message to queue.", message.GetType().FullName);
```
gkinsman, line 111: Should we have a test to test that false is returned from the dispatcher when there is no handler for it?

JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
shaynevanasperen: Turns out this is not even possible in master: https://github.com/justeat/JustSaying/pull/728

What do others think?

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
shaynevanasperen: This is implied, by the test we already have, where it asserts `ShouldUpdateMessageVisibility()`, as it was already done for the `AndMessageProcessingFails` scenario.

gkinsman: I guess I don't quite understand why we're asserting that a side effect occurred, instead of directly asserting that the message dispatcher returned false. IMO we shouldn't need an SQS client to test the message dispatcher.

gkinsman: Ah ok I can see I've misunderstood a bit here - it is the only way to determine that as MessageDispatcher is a bit of a black box and doesn't return. I was confusing it with the handler 🙂. Fair go, thanks for clarifying 👍

slang25:
Apparently I closed this! That's news to me

JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
slang25: I've added a comment here:
https://github.com/justeat/JustSaying/pull/728#discussion_r449260954

JustSaying.UnitTests/AwsTools/MessageHandling/MessageDispatcherTests/WhenDispatchingMessage.cs
slang25: Agreed, the log message is a side effect, not really something that should be asserted on IMO

shaynevanasperen: Done.

PR #728, opened 2020-07-02
Fix bug identified in diagnosis of production incident

shaynevanasperen:
* Log when we don't have a matching handler for the given message type.
* If two consumers are subscribed to a queue that has multiple message types in it, one of them may not be able to handle all the messages (each handled a subset). So we need to return the message to the queue for the other subscriber to try handling it.

This is just copying the bugfix from #727

### DIFF
--- a/src/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
+++ b/src/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
@@ -145,7 +145,8 @@ namespace JustSaying.AwsTools.MessageHandling
 
             if (handler == null)
             {
-                return true;
+                _logger.LogError("Failed to dispatch. Handler for message of type '{MessageTypeName}' not found in handler map.", message.GetType().FullName);
+                return false;
             }
 
             var watch = System.Diagnostics.Stopwatch.StartNew();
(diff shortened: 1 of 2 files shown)


src/JustSaying/AwsTools/MessageHandling/MessageDispatcher.cs
```diff
@@ -145,7 +145,9 @@ private async Task<bool> CallMessageHandler(Message message)
 
             if (handler == null)
             {
-                return true;
+                _logger.LogError("Handler for message of type '{MessageTypeName}' not found in handler map. Returning message to queue.", message.GetType().FullName);
+                // Return false so that we give other queue subscribers a chance to handle the message
```
slang25, line 149: Very minor nit; I think the phrasing should indicate that dispatching has failed, rather than implying this is a supported pattern.
So maybe:
```csharp
_logger.LogError("Failed to dispatch. Handler for message of type '{MessageTypeName}' not found in handler map.", message.GetType().FullName);
```
and no comment.

shaynevanasperen: Done.

PR #745, opened 2020-08-05
GitHub actions

slang25:
This is a first cut, I haven't done the NuGet related stuff as I don't have permissions on this repo to setup secrets.

This PR adds:
- A GitHub Action to build, run tests of each platform, integration tests on linux only, and collects artifacts
- Enables deterministic builds under GitHub Actions, I had to swap out some acceptance test code as it expects a full path in the symbols, this may already be fixed in Shouldy master (Acceptance tests implementation have been removed in favour of a new library), but we are waiting on a pre-release build to try it.
- Disables MinVer when running locally in debug, just for speed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,60 @@
+name: build
+
+on:
+  push:
+    branches: [ master ]
+    tags: [ v* ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
+        include:
+          - os: macos-latest
+            os_name: macos
+          - os: ubuntu-latest
+            os_name: linux
+          - os: windows-latest
+            os_name: windows
+
+    steps:
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Setup .NET Core SDK
+      uses: actions/setup-dotnet@v1
+
+    - name: Build, Test and Package
+      shell: pwsh
+      run: ./Build.ps1
+      env:
+        DOTNET_CLI_TELEMETRY_OPTOUT: true
+        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+        NUGET_XMLDOC_MODE: skip
+
+    - uses: codecov/codecov-action@v1
+      name: Upload coverage to Codecov
+      with:
+        file: ./artifacts/coverage.netcoreapp3.1.cobertura.xml
+        flags: ${{ matrix.os_name }}
+
+    - name: Publish NuGet packages
+      uses: actions/upload-artifact@v1
+      with:
+        name: packages-${{ matrix.os_name }}
+        path: ./artifacts/packages
+
+    - name: Push NuGet packages to NuGet.org
+      run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.NUGET_TOKEN }} --skip-duplicate --source https://api.nuget.org/v3/index.json
+      if: ${{ github.repository_owner == 'justeat' && startsWith(github.ref, 'refs/tags/v') && runner.os == 'Windows' }}
--- a/Build.ps1
+++ b/Build.ps1
@@ -1,3 +1,5 @@
+#! /usr/bin/pwsh
+
 param(
     [Parameter(Mandatory = $false)][string] $Configuration = "Release",
     [Parameter(Mandatory = $false)][string] $VersionSuffix = "",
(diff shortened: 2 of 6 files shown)


gkinsman:
Nice one 👍 . I usually use https://github.com/droyad/Assent but since Shouldly was already in here I just used that. Should we wait for Shouldly to support DiffEngine and use that, or just use Assent now? I'm fine either way.

slang25:
Nice, I didn't know about Assent. Unfortunately they all suffer from the same issue, which is trying to resolve the file path doesn't work. Perhaps I should disable it for test runs?

gkinsman:
Ah ok, is that a side effect of using deterministic buiilds? I'm guessing it erases or standardises paths so that the build is the same wherever you run it? I suppose in that case we'd need a lib that can read embedded files as approval sources.

martincostello:
Deterministic builds causes issues with Code Coverage at the moment, so we should defer that until coverlet fixes it.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Setup .NET Core SDK
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.302
```
martincostello, line 38: You can remove this - the task will read it from `global.json` if it's not specified, so it's one less thing to keep updated.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+        dotnet pack --no-restore --verbosity normal -c Release  --output ./artifacts/packages ./src/JustSaying/JustSaying.csproj
+        dotnet pack --no-restore --verbosity normal -c Release  --output ./artifacts/packages ./src/JustSaying.Models/JustSaying.Models.csproj
+        dotnet pack --no-restore --verbosity normal -c Release  --output ./artifacts/packages ./src/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
+        dotnet pack --no-restore --verbosity normal -c Release  --output ./artifacts/packages ./src/JustSaying.Extensions.DependencyInjection.StructureMap/JustSaying.Extensions.DependencyInjection.StructureMap.csproj
+
+    - name: Publish NuGet packages
+      uses: actions/upload-artifact@v1
+      with:
+        name: packages-${{ matrix.os_name }}
+        path: ./artifacts/packages
```
martincostello, line 71: I'll set up the secret a bit later today, then you can add the publishing part. There's also a fix I need to add to httpclient-interception I need to add for that before you copy it over.

Directory.Build.props
```diff
@@ -54,6 +56,7 @@
     <EmbedAllSources Condition=" '$(IsTestProject)' != 'true' AND '$(NCrunch)' == '' ">true</EmbedAllSources>
     <!-- HACK Workaround flaky build in CI for Linux -->
     <BuildInParallel Condition=" '$(OS)' != 'Windows_NT' and '$(CI)' != '' ">false</BuildInParallel>
+    <ContinuousIntegrationBuild Condition="'$(GITHUB_ACTIONS)' == 'true'" >true</ContinuousIntegrationBuild>
```
martincostello, line 59: Can just use `$(CI)`.

tests/JustSaying.UnitTests/JustSaying.UnitTests.csproj
```diff
@@ -17,5 +17,9 @@
     <PackageReference Include="Polly" Version="7.2.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
+    <PackageReference Include="Quibble.Xunit" Version="0.2.51" />
```
martincostello, line 20: Doesn't build with this as it's not strong-named.

slang25: I wept a little when I saw the error

.github/workflows/build.yml
martincostello: Here's the fix: https://github.com/justeat/httpclient-interception/pull/234

martincostello: And I've added the `NUGET_TOKEN` secret.

slang25:
I'll rollback deterministic builds, it's causing too much pain right now.

On the build script, yeah good point. I like the way GHA presents the individual steps, but not at the cost of having duplicated definitions, I'll use the pwsh script 👍

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
...
+      
+      - name: Build, Test and Package
+      shell: pwsh
+      run: ./Build.ps1
+      env:
+        DOTNET_CLI_TELEMETRY_OPTOUT: true
+        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+        NUGET_XMLDOC_MODE: skip
+
+    - name: Pack
```
martincostello, line 46: I guess this isn't needed now?

.github/workflows/build.yml
```diff
@@ -0,0 +1,54 @@
+name: build
+
+on:
+  push:
+    branches: [ github-actions ]
```
martincostello, line 5: Pushes to `master` too.

martincostello:
Looks like the GHA workflow is missing the step to upload to codecov.

Build.ps1
```diff
@@ -1,3 +1,5 @@
+#! /usr/bin/pwsh
```
martincostello, line 1: Would be good to +x on this file.

martincostello (approved):

slang25:
That didn't go as well as I'd hoped: https://github.com/justeat/JustSaying/actions/runs/230252521

PR #747, opened 2020-08-06
Add support for System.Text.Json when parsing MessageAttributes

### LINKED ISSUE #723: Check SystemTextJsonSerializer GetMessageAttributes has been implemented
Currently not implemented

https://github.com/justeat/JustSaying/blob/5ead34edc7a6f63acac4c747f82f47ed92b55c79/src/JustSaying/Messaging/MessageSerialization/SystemTextJsonSerializer.cs#L59

gkinsman:
This PR resolves #723.

**Note on why some things have been deleted**

When the configuration lambda is passed to `AddJustSaying` or `ConfigureJustSaying`, the lambda is only executed _after_ the DI container is built, because the `MessageBusBuilder` needs an `IServiceProvider` to resolve components to build the bus:
```csharp
var builder = new MessagingBusBuilder()
    .WithServiceResolver(new ServiceProviderResolver(serviceProvider));
```

Unfortunately, this means that it's not possible to customise the `IServiceCollection` that becomes the `IServiceProvider` in the `AddJustSaying` configuration because the bus has already been built at this point.

**More detail** image

The current API to set the `IMessageSerializationFactory` uses methods on the `ServicesBuilder`: `WithSystemTextJson` and `WithNewtonsoftJson`. Those methods set the `ServicesBuilder.MessageSerializationFactory` to the corresponding newtonsoft/systemtextjson factory so that it can be passed into the `IMessageSerializationRegister` later. 

However, because  `IMessageSerializationRegister` is pulled from the DI container (unless it's been overriden), there is no way to give it the overridden `IMessageSerializationFactory`. We either pull the `IMessageSerializationRegister` from DI (in which case it wouldn't use the overridden `IMessageSerializationFactory`), or it has been overridden and we can't pass the custom `IMessageSerializationFactory` to it anyway.

Instead, I've just overridden it in the container directly:
```csharp
var services = GivenJustSaying()
    .ConfigureJustSaying((builder) => builder
        .WithLoopbackTopic<SimpleMessage>(UniqueName))
    .AddSingleton<IMessageSerializationFactory, SystemTextJsonSerializationFactory>()
    .AddSingleton<IHandlerAsync<SimpleMessage>>(handler);
```

It's perhaps not an ideal solution, but I can't see a way to cleanly add it to the `AddJustSaying` API. Is this potentially a weakness in that API that we want to resolve? Can anyone think  of any other extensions we'd want to add that would require updating the DI container that would necessitate changing it?

**tl;dr: I don't believe the `AddJustSaying` API can be used to modify the DI container in its current form, so I removed this API that needed to do that.**


src/JustSaying/Fluent/ServicesBuilder.cs
```diff
@@ -44,11 +44,6 @@ internal ServicesBuilder(MessagingBusBuilder busBuilder)
         /// </summary>
         internal Func<IMessageLockAsync> MessageLock { get; private set; }
 
-        /// <summary>
-        /// Gets or sets a delegate to a method to create the <see cref="IMessageSerializationFactory"/> to use.
-        /// </summary>
-        internal Func<IMessageSerializationFactory> MessageSerializationFactory { get; private set; }
```
martincostello, line 50: I think we'd still want this hook for usage of JustSaying _without_ a DI container.

src/JustSaying/Fluent/ServicesBuilderExtensions.cs
```diff
@@ -1,145 +0,0 @@
...
-using System.Text.Json;
-using Newtonsoft.Json;
-
-namespace JustSaying.Fluent
-{
-    /// <summary>
-    /// A class containing extension methods for the <see cref="ServicesBuilder"/> class.  This class cannot be inherited.
-    /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public static class ServicesBuilderExtensions
```
martincostello, line 12: Again, this is fluency for the JustSaying set up, not the MS DI one.

src/JustSaying/Fluent/ServicesBuilder.cs
gkinsman: Given that we expose a setter on `SerializationRegister` which takes a dependency on `MessageSerializationFactory ` - I don't think we can even pass this factory in to anywhere that could use it. 

Perhaps we need to change `CreateBus` from:
```csharp
        private JustSayingBus CreateBus(IMessagingConfig config, ILoggerFactory loggerFactory)
        {
            IMessageSerializationRegister register =
                ServicesBuilder?.SerializationRegister?.Invoke() ?? ServiceResolver.ResolveService<IMessageSerializationRegister>();

            return new JustSayingBus(config, register, loggerFactory);
        }
```

to something like this?
```csharp
        private JustSayingBus CreateBus(IMessagingConfig config, ILoggerFactory loggerFactory)
        {
            IMessageSerializationRegister register =
                ServicesBuilder?.SerializationRegister?.Invoke() ??
                ServiceResolver?.ResolveService<IMessageSerializationRegister>() ??
                new MessageSerializationRegister(<subjectProviderFromSomewhere>, 
                    ServicesBuilder.MessageSerializationFactory ?? new NewtonsoftSerializationFactory());

            return new JustSayingBus(config, register, loggerFactory);
        }
```

Do we have tests around using JS without DI? I'm not sure how I'd even go about using it without DI 😁

martincostello: I can't remember if there's an explicit tests for it without one to be honest, but the DI packages build on top of it to plug the dependencies into the service resolver, so it's sort of implicitly tested that way.

gkinsman: Ah ok fair. Yeah as for this, it appears that we're able to support a custom `IMessageSerializationFactory` already via the `IMessageSerializationRegister` override - the user could new up a `MessageSerializationRegister` and choose the serialization factory then. I'm not sure how this would work otherwise.

gkinsman:
@martincostello happy with this one? 🙂

PR #757, opened 2020-08-14
Add SNS request Id to log message when message is published

maurofranchi:
We are currently experiencing some SNS issues in production and AWS always ask us to provide them with some SNS Request Ids related to the problematic publish operations. Unfortunately JustSaying is not returning it.

This PR is to simply add the returned request Id (if present) to the message logged when a message is published.
I know V6 is currently in maintenance mode but this could help us investigate a bit better on any issue we might have on AWS.

### DIFF
--- a/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
+++ b/JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
@@ -44,7 +44,7 @@ namespace JustSaying.AwsTools.MessageHandling
         }
 
         public abstract Task<bool> ExistsAsync();
-        
+
         public Task PublishAsync(Message message) => PublishAsync(message, CancellationToken.None);
 
         public async Task PublishAsync(Message message, CancellationToken cancellationToken)
@@ -54,7 +54,11 @@ namespace JustSaying.AwsTools.MessageHandling
             try
             {
                 var response = await Client.PublishAsync(request, cancellationToken).ConfigureAwait(false);
-                _eventLog.LogInformation($"Published message: '{request.Subject}' with content {request.Message}");
+                _eventLog.LogInformation(
+                    "Published message: '{Subject}' with content {Message} and request Id '{SnsRequestId}'",
+                    request.Subject,
+                    request.Message,
+                    response?.ResponseMetadata?.RequestId);
 
                 MessageResponseLogger?.Invoke(new MessageResponse
                 {
@@ -92,7 +96,7 @@ namespace JustSaying.AwsTools.MessageHandling
                         DataType = source.Value.DataType
                     };
                 });
-            
+
             return new PublishRequest
             {
                 TopicArn = Arn,


JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
```diff
@@ -54,7 +54,11 @@ public async Task PublishAsync(Message message, CancellationToken cancellationTo
             try
             {
                 var response = await Client.PublishAsync(request, cancellationToken).ConfigureAwait(false);
-                _eventLog.LogInformation($"Published message: '{request.Subject}' with content {request.Message}");
+                _eventLog.LogInformation(
```
martincostello, line 57: It's probably best to leave this as non-structured for the v6 branch for consistency.

I don't know what would be better, but also as the content isn't escaped this might could out looking a bit messy.

Meh, I don't have strong opinions either way.

slang25:
If the publish is fails with an exception then today you can use `HandleException` to then log the request id. @JosephWoodward added this functionality for this reason (AWS wanted request ids to be able to provide support)
https://github.com/justeat/JustSaying/pull/350

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
maurofranchi: I have used the structured version because I wanted to make sure the Request Id was always logged and not truncated because the logged message reaches the logging client limit. I will refactor it and put it in front to make sure it is always logged.

martincostello:
Ah - for some reason I thought that capability only existed in v7. In that case, we can just use that and not bother changing the logs.

martincostello:
Just checked the tags of the merge commit (https://github.com/justeat/JustSaying/commit/ef8334a099e130ae06fa4aff61de21898321a98c) and it's in all the v6 releases.

maurofranchi:
> If the publish is fails with an exception then today you can use `HandleException` to then log the request id. @JosephWoodward added this functionality for this reason (AWS wanted request ids to be able to provide support)
> #350

I saw the messageId is returned, but lately AWS started asking for RequestId which is not currently added to the `MessageResponse` object from that PR you linked. Do you want me to add it to that object?

maurofranchi:
> I saw the messageId is returned, but lately AWS started asking for RequestId which is not currently added to the `MessageResponse` object from that PR you linked. Do you want me to add it to that object?

Also I forgot to say that in our case the request doesn't fail, so we would not have the `handleException`. It just takes a bit longer to publish the message.

slang25:
Ah yeah, that doesn't help

slang25:
I don't think adding it to the existing object makes sense, cause it relies on an exception occuring. I think this change makes most sense 👍

JustSaying/AwsTools/MessageHandling/SnsTopicBase.cs
martincostello: Hmm, in that case let's leave it structured, otherwise it might break if anyone is grepping the logs.

PR #769, opened 2020-08-30
Update AWS SDK to 3.5.0

slang25:
Fixes #768

PR #783, opened 2020-09-09
Update AWS SDK to 3.5.0

### LINKED ISSUE #768: Update to AWS SDK 3.5
This might be a good time to take the update.
https://aws.amazon.com/blogs/developer/aws-sdk-for-net-v3-5-now-generally-available/

Old:
image
New:
image

#### FuGet diffs:
https://www.fuget.org/packages/AWSSDK.SQS/3.5.0.5/lib/netstandard2.0/diff/3.3.103.26/
https://www.fuget.org/packages/AWSSDK.SimpleNotificationService/3.5.0/lib/netstandard2.0/diff/3.3.102.17/
https://www.fuget.org/packages/AWSSDK.Core/3.5.0/lib/netstandard2.0/diff/3.3.107.40/

### Pros
- Fewer assemblies for legacy runtimes (smaller nuget downloads)
- Extra async enumerable APIs available to take advantage of potentially in future

### Cons
- Required version uplift for consumers
- Low risk breaking changes exist in `AWSSDK.Core` (If consumers are using "Pubternal" APIs 🤷‍♀️)

slang25:
#769 - take 2

Fixes #768

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,8 +6,8 @@
     <NewtonsoftJsonVersion>12.0.3</NewtonsoftJsonVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <AwsSdkSnsVersion>3.5.0</AwsSdkSnsVersion>
-    <AwsSdkSqsVersion>3.5.0</AwsSdkSqsVersion>
+    <AwsSdkSnsVersion>3.5.0.9</AwsSdkSnsVersion>
+    <AwsSdkSqsVersion>3.5.0.9</AwsSdkSqsVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="coverlet.msbuild" Version="2.9.0" PrivateAssets="All" />
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,8 +6,8 @@
     <NewtonsoftJsonVersion>12.0.3</NewtonsoftJsonVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <AwsSdkSnsVersion>3.5.0.9</AwsSdkSnsVersion>
-    <AwsSdkSqsVersion>3.5.0.9</AwsSdkSqsVersion>
+    <AwsSdkSnsVersion>3.5.0</AwsSdkSnsVersion>
+    <AwsSdkSqsVersion>3.5.0</AwsSdkSqsVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="coverlet.msbuild" Version="2.9.0" PrivateAssets="All" />
--- a/src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
+++ b/src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
@@ -1,3 +1,4 @@
+#pragma warning disable 618
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
--- a/src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
+++ b/src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
@@ -1,55 +1,56 @@
-#pragma warning disable 618
-using System.Collections.Generic;
-using System.Linq;
+using System;
+using System.Text.Json;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
-using Amazon.Auth.AccessControlPolicy;
-using Amazon.Auth.AccessControlPolicy.ActionIdentifiers;
 using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
 
 namespace JustSaying.AwsTools.MessageHandling
 {
-    public class SnsPolicy
+    internal static class SnsPolicy
     {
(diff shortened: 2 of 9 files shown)


src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
```diff
@@ -1,56 +1,58 @@
...
-
-            var snsPolicy = new Policy()
-                .WithStatements(GetDefaultStatement(sourceArn))
-                .WithStatements(new Statement(Statement.StatementEffect.Allow)
-                    .WithPrincipals(_accountIds.Select(a => new Principal(a)).ToArray())
-                    .WithResources(new Resource(sourceArn))
-                    .WithActionIdentifiers(actions));
-            var attributeValue = snsPolicy.ToJson();
-            var setQueueAttributesRequest = new SetTopicAttributesRequest(sourceArn, "Policy", attributeValue);
+            var sourceAccountId = ExtractSourceAccountId(policyDetails.SourceArn);
```
gkinsman, line 14: Is switching from the fluent api to json here because the fluent api is out of date from the SQS API?

gkinsman: Looks like its from the SDK - did they just remove it?

slang25: It's been marked as obsolete, and the message says to use JSON going forwards

slang25: It's not broken, but they say they won't be keeping it up to date

Directory.Build.props
```diff
@@ -6,8 +6,8 @@
     <NewtonsoftJsonVersion>12.0.3</NewtonsoftJsonVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <AwsSdkSnsVersion>3.5.0.9</AwsSdkSnsVersion>
-    <AwsSdkSqsVersion>3.5.0.9</AwsSdkSqsVersion>
+    <AwsSdkSnsVersion>3.5.0</AwsSdkSnsVersion>
+    <AwsSdkSqsVersion>3.5.0</AwsSdkSqsVersion>
```
martincostello, line 10: Tests can use latest.

```suggestion
    <AwsSdkSnsVersion>3.5.0.9</AwsSdkSnsVersion>
    <AwsSdkSqsVersion>3.5.0.9</AwsSdkSqsVersion>
```

src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
gkinsman: Ah that's interesting, fair enough. I wonder if we should keep this json in an external file as a template that we load on startup. Likely something we can put off until we do the work to disable creation on startup, as we'd want to remove the permissions for those actions too.

src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
```diff
@@ -1,56 +1,58 @@
...
+            var policyJson = $@"{{
+    ""Version"" : ""2012-10-17"",
+    ""Statement"" : [
+        {{
+            ""Sid"" : ""{Guid.NewGuid().ToString().Replace("-", "")}"",
```
martincostello, line 19: Do we want the SIDs to be non-deterministic? Or maybe could add a way to override the SID to something custom?

src/JustSaying/AwsTools/MessageHandling/SqsPolicy.cs
```diff
@@ -1,46 +1,48 @@
...
-        {
-            var topicArnWildcard = CreateTopicArnWildcard(sourceArn);
-            ActionIdentifier[] actions = { SQSActionIdentifiers.SendMessage };
+            var topicArnWildcard = CreateTopicArnWildcard(policyDetails.SourceArn);
+
+            var policyJson = $@"{{
+    ""Version"" : ""2012-10-17"",
+    ""Statement"" : [
+        {{
+            ""Sid"" : ""{Guid.NewGuid().ToString().Replace("-", "")}"",
```
martincostello, line 18: Same comments as above.

src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
slang25: We could remove them entirely, I was tempted to while making this change. I opted here for matching the existing behaviour.

src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
martincostello: Trying to ship a template as an external file via NuGet would be a bit bleurgh - we could embed it as a JSON resource, but it still wouldn't be user editable. Would be good to provide a way to override the policy and let the user provide an entire custom one instead.

gkinsman: I was thinking as an embedded file, yeah. If we wanted it to be user editable I guess we could provide well-known settings that we substitute in where it makes sense.

martincostello (approved):

PR #784, opened 2020-09-09
Publish pre-release packages to github packages

gkinsman:
Shall we try Github packages for pre-releases to make consuming them from tooling a little nicer than the zip artifacts?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,7 +67,7 @@ jobs:
 
       - name: Publish pre-release NuGet packages to Github packages
         run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.GITHUB_TOKEN }} --skip-duplicate  --source https://nuget.pkg.github.com/justeat/index.json
-        if: ${{ runner.os == 'Windows' }}
+        if: ${{ github.repository_owner == 'justeat' && runner.os == 'Windows' }}
 
       - name: Push NuGet packages to NuGet.org
         run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.NUGET_TOKEN }} --skip-duplicate --source https://api.nuget.org/v3/index.json
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,7 @@ jobs:
       - name: Publish pre-release NuGet packages to GitHub packages
         run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.GITHUB_TOKEN }} --skip-duplicate  --source https://nuget.pkg.github.com/justeat/index.json
         if: ${{ github.repository_owner == 'justeat' && runner.os == 'Windows' }}
+        continue-on-error: true
 
       - name: Push NuGet packages to NuGet.org
         run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.NUGET_TOKEN }} --skip-duplicate --source https://api.nuget.org/v3/index.json
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
           name: packages-${{ matrix.os_name }}
           path: ./artifacts/packages
 
-      - name: Publish pre-release NuGet packages to Github packages
+      - name: Publish pre-release NuGet packages to GitHub packages
         run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.GITHUB_TOKEN }} --skip-duplicate  --source https://nuget.pkg.github.com/justeat/index.json
         if: ${{ github.repository_owner == 'justeat' && runner.os == 'Windows' }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -65,6 +65,10 @@ jobs:
           name: packages-${{ matrix.os_name }}
           path: ./artifacts/packages
 
+      - name: Publish pre-release NuGet packages to Github packages
```
martincostello, line 68: Nit:

```suggestion
      - name: Publish pre-release NuGet packages to GitHub packages
```

.github/workflows/build.yml
```diff
@@ -65,6 +65,10 @@ jobs:
...
+        run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.GITHUB_TOKEN }} --skip-duplicate  --source https://nuget.pkg.github.com/justeat/index.json
+        if: ${{ runner.os == 'Windows' }}
```
martincostello, line 70: Otherwise forks will start publishing packages as well, which the arbitrary forker probably doesn't want.

```suggestion
        if: ${{ github.repository_owner == 'justeat' && runner.os == 'Windows' }}
```

martincostello (dismissed):
Great if it works, but I tried to get this working on some of my personal repos a few months ago and I couldn't ever get the upload to work.

.github/workflows/build.yml
martincostello: Also you'll need to not do this on PRs, fork PRs don't have a token.

gkinsman: Ahh that explains why the build isn't working - I was curious how that worked for PR's. Thanks!

gkinsman:
> Great if it works, but I tried to get this working on some of my personal repos a few months ago and I couldn't ever get the upload to work.

Ah fair - it looks like there's some reasonable docs around it now. Shall we merge it and see if we can get master pushing to packages?

martincostello:
Happy to merge to try it once it's green, but if it fails you might need to do a follow up to just comment it out.

gkinsman:
> Happy to merge to try it once it's green, but if it fails you might need to do a follow up to just comment it out.

Yep going to timebox this and revert it if I can't get it working 😁

martincostello (approved):

gkinsman:
I'm going to just set `continue-on-error: true` so its best effort

slang25:
Great idea. Who do we beg to make GitHub NuGet feeds public?

gkinsman:
> Great idea. Who do we beg to make GitHub NuGet feeds public?

They are public [here](https://github.com/orgs/justeat/packages), we just haven't published any yet :)

slang25:
No way, ok that's awesome! I thought GH had them private only, which would be a pain.

martincostello:
> GitHub Packages is not available for **private** repositories on legacy per-repository billing plans.

We're good for public - for private, someone in the finance/procurement sphere will have to change the billing plan.

gkinsman:
I don't understand why the coverage is failing now - it seems to think there are changes where there aren't any 🙃

image

martincostello:
The builds aren't 100% deterministic, so the coverage varies. Note that the coverage statuses aren't required.

.github/workflows/build.yml
```diff
@@ -65,6 +65,11 @@ jobs:
           name: packages-${{ matrix.os_name }}
           path: ./artifacts/packages
 
+      - name: Publish pre-release NuGet packages to GitHub packages
+        run: dotnet nuget push "artifacts\packages\*.nupkg" --api-key ${{ secrets.GITHUB_TOKEN }} --skip-duplicate  --source https://nuget.pkg.github.com/justeat/index.json
+        if: ${{ github.repository_owner == 'justeat' && runner.os == 'Windows' }}
+        continue-on-error: true
```
martincostello, line 71: I think this would be neater than just ignoring the error:

```suggestion
        if: ${{ github.repository_owner == 'justeat' && github.ref == 'refs/heads/master' && runner.os == 'Windows' }}
```

martincostello (dismissed):

gkinsman:
> The builds aren't 100% deterministic, so the coverage varies. Note that the coverage statuses aren't required.

Oh righto, awesome. In that case this PR just needs your green tick please, and we can try it out :)

martincostello (approved):

martincostello:
Looks like the same sort of problems I was getting, where the responses aren't helpful at all:

https://github.com/justeat/JustSaying/runs/1091741953?check_suite_focus=true#step:8:8

gkinsman:
Hmmm, if we want to keep going down this route, it looks like we might need to authenticate with GPR before attempting the push, as described [here](https://stackoverflow.com/a/58025355). There's an open issue on NuGet with sending the API key directly: https://github.com/NuGet/Home/issues/8580.

gkinsman:
@martincostello we explicitly specify nuget when pushing there so I think we could set this up and regular nuget push should still work

martincostello:
You can try that too, but I did a lot of banging my head against a wall trying to get all this working including explicitly configuring NuGet.exe and the .NET Core SDK 😅

gkinsman:
> You can try that too, but I did a lot of banging my head against a wall trying to get all this working including explicitly configuring NuGet.exe and the .NET Core SDK 😅

Yeah I'm ready to give up if [this PR](https://github.com/justeat/JustSaying/pull/785) doesn't work

PR #785, opened 2020-09-09
Explicitly authenticate with GPR before pushing packages

gkinsman:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,8 +65,15 @@ jobs:
           name: packages-${{ matrix.os_name }}
           path: ./artifacts/packages
 
+      - name: Authenticate with github packages
+        uses: actions/setup-dotnet@v1
+        with:
+          source-url: https://nuget.pkg.github.com/justeat/index.json
+        env:
+          NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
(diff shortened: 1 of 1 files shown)


gkinsman:
@martincostello shall we give this a go?

.github/workflows/build.yml
```diff
@@ -65,8 +65,15 @@ jobs:
           name: packages-${{ matrix.os_name }}
           path: ./artifacts/packages
 
+      - name: Authenticate with github packages
```
martincostello: Nit:

```suggestion
      - name: Authenticate with GitHub packages
```

.github/workflows/build.yml
```diff
@@ -65,8 +65,15 @@ jobs:
...
+        uses: actions/setup-dotnet@v1
+        with:
+          source-url: https://nuget.pkg.github.com/justeat/index.json
+        env:
+          NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
martincostello, line 73: Probably want this step to have the same restrictions as the push itself, otherwise it'll try and run and authenticate with our feed in PRs when there's no token.

gkinsman: done

martincostello (approved):

PR #789, opened 2020-09-11
[WIP] Remove message lock

slang25:
This removes `IMessageLockAsync`, and the `ExactlyOnce` attribute stuff.


martincostello (dismissed):

maurofranchi (approved):

src/JustSaying/AwsTools/MessageHandling/SqsQueueByUrl.cs
```diff
@@ -1,33 +0,0 @@
...
-using System.Linq;
-using System.Threading.Tasks;
-using Amazon;
-using Amazon.SQS;
-using Amazon.SQS.Model;
-using Microsoft.Extensions.Logging;
-
-namespace JustSaying.AwsTools.MessageHandling
-{
-    public class SqsQueueByUrl : SqsQueueBase
```
gkinsman, line 11: It feels good to be rid of this - was it ever used anywhere?

tests/JustSaying.IntegrationTests/Fluent/Subscribing/WhenTwoDifferentHandlersHandleAMessage.cs
```diff
@@ -19,8 +18,11 @@ public WhenTwoDifferentHandlersHandleAnExactlyOnceMessage(ITestOutputHelper outp
         public async Task Then_Both_Handlers_Receive_The_Message()
         {
             // Arrange
-            var handler1 = new ExactlyOnceHandlerNoTimeout();
-            var handler2 = new ExactlyOnceHandlerNoTimeout();
+            var nullStoreLogger = NullLogger<TestMessageStore<SimpleMessage>>.Instance;
+            var nullHandlerLogger = NullLogger<MessageStoringHandler<SimpleMessage>>.Instance;
+
+            var handler1 = new MessageStoringHandler<SimpleMessage>(new TestMessageStore<SimpleMessage>(nullStoreLogger), nullHandlerLogger);
```
gkinsman, line 24: If you wanted to simplify this a little, there's also an `InspectableHandler<T>` that just collects messages on the handler, instead of using a separate store. MessageStoringHandler is useful when using DI to resolve a handler instead of creating it like here :)

gkinsman:
Curious, is anyone at JE still using this? Or would we just say to them: "if you want to upgrade, don't try to do exactly-once"?

src/JustSaying/AwsTools/MessageHandling/SqsQueueByUrl.cs
slang25: Yeah, unused, maybe should be a separate PR 🤷

tests/JustSaying.IntegrationTests/Fluent/Subscribing/WhenTwoDifferentHandlersHandleAMessage.cs
slang25: Thanks, yeah I'll take a look, this felt a bit clunky

slang25:
> Curious, is anyone at JE still using this? Or would we just say to them: "if you want to upgrade, don't try to do exactly-once"?

I've seen a couple of implementations of `IMessageLock` (i.e. before v6 when it was changed to `IMessageLockAsync`), this adds another barrier to migrating those things.

I'm a little conflicted by removing the support for this. I feel happy we are removing something that is mostly unused, keeping JustSaying simpler and focused on its core sweet spot. On the other hand, this code isn't particularly burdensome and will continue to work if we leave it in.

If we plan to make the dispatcher middleware a first-class extension point, we could include a message locking example in the docs.

martincostello (approved):

shaynevanasperen (approved):

gkinsman:
I think this is a good move.

After this, we could also remove the `StopwatchHandler` which is the other interceptor, and remove all of the wrapping code that wraps the original handler, which would mean we have the _actual_ handler type in the `HandlerMap`. This means we can remove the `IMeasureHandlerExecutionTime` interface and put that method directly on `IMessageMonitor`, removing yet another place that we do monitoring slightly differently.

There's plenty of complexity caused by these message wrappers, that we should move to middlewares for.

slang25:
Superseded by this better approach: https://github.com/justeat/JustSaying/pull/809

PR #791, opened 2020-09-14
Update JE logo

slang25:
Using new logo.

Opted for lower resolution layout but rebuilt from downscaling a higher-res version. Ignoring the 128x128 NuGet recommendation for the more common 512x512.

Reduced pallet to 16 colour and removed all metadata, optimized to 8kb, not bad.


martincostello (approved):
Do you want to do the other repos too? 😉

brainmurphy (approved):

PR #807, opened 2020-10-15
Misc fixes for issues encountered in testing

gkinsman:
This PR wraps up a number of fixes for situations and issues I've encountered while testing JS 7 in a real app.

- Interrogation:  
  - Remove serialization of full RegionEndpoint type and just use the SystemName

- Queue Creation:
  - Remove bool return on topic creation and just use exceptions (the return value wasn't used outside tests.
    - This means that there's no observable difference between calling CreateAsync on a queue that does/does not exist. Exceptions are thrown in error cases.
  - No longer cache queues by region, because we don't support multiple regions anymore
  - Little more logging during queue/topic existence checks
  - More null checks to fail fast on startup
- Monitoring
  - Add a test for IMessageMonitor to ensure the key monitor methods are called
- MessageAttributes
  - The correct data type for string attributes is "String" and not "StringValue". LocalStack will just return what its given, but real AWS will not filter queues or return the attributes.
- Formatting
  - Formatted touched files to match the editorconfig


src/JustSaying/AwsTools/MessageHandling/SqsQueueByName.cs
```diff
@@ -38,6 +39,10 @@ public override async Task<bool> CreateAsync(SqsBasicConfiguration queueConfig,
                         }).ConfigureAwait(false);
                     }
                 }
+                else
+                {
+                    Logger.LogInformation("Error Queue {QueueName} already exists, skipping", ErrorQueue.QueueName);
```
martincostello, line 44:
```suggestion
                    Logger.LogInformation("Error queue {QueueName} already exists, skipping", ErrorQueue.QueueName);
```

src/JustSaying/AwsTools/QueueCreation/AmazonQueueCreator.cs
```diff
@@ -111,6 +104,11 @@ public QueueWithAsyncStartup<SqsQueueByName> EnsureQueueExists(string region, Sq
             IAmazonSimpleNotificationService amazonSimpleNotificationService,
             string topicArn, IAmazonSQS amazonSQS, Uri queueUrl, string filterPolicy)
         {
+            if(amazonSimpleNotificationService == null) throw new ArgumentNullException(nameof(amazonSimpleNotificationService));
```
martincostello, line 107:
```suggestion
            if (amazonSimpleNotificationService == null) throw new ArgumentNullException(nameof(amazonSimpleNotificationService));
```

src/JustSaying/AwsTools/QueueCreation/AmazonQueueCreator.cs
```diff
@@ -111,6 +104,11 @@ public QueueWithAsyncStartup<SqsQueueByName> EnsureQueueExists(string region, Sq
...
+            if (amazonSQS == null) throw new ArgumentNullException(nameof(amazonSQS));
+            if (queueUrl == null) throw new ArgumentNullException(nameof(queueUrl));
+            if (string.IsNullOrEmpty(topicArn)) throw new ArgumentException("topicArn cannot be null or empty", nameof(topicArn));
```
martincostello, line 110:
```suggestion
            if (string.IsNullOrEmpty(topicArn)) throw new ArgumentException("topicArn cannot be null or empty.", nameof(topicArn));
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -194,7 +194,7 @@ private async Task RunImplAsync(CancellationToken stoppingToken)
             catch (OperationCanceledException)
             {
                 _log.LogDebug("Suppressed an exception of type {ExceptionType} which likely" +
```
martincostello, line 196: I think it would be clearer to add the space to the end of the previous sentence (or just put it on one line, the splitting it into two doesn't really seem necessary other than arbitrary length length).

tests/JustSaying.IntegrationTests/Fluent/Configuration/WhenUsingACustomSubjectProvider.cs
```diff
@@ -53,17 +53,21 @@ public async Task ThenItIsUsed()
                     await listener.StartAsync(cancellationToken);
                     await publisher.StartAsync(cancellationToken);
 
-                    // Let's send an OrderPlaced, but the subject will be SimpleMessage
+                    // Let's send an OrderPlaced, but the subject will be a guid
```
martincostello, line 56:
```suggestion
                    // Let's send an OrderPlaced, but the subject will be a GUID
```

tests/JustSaying.IntegrationTests/Fluent/IntegrationTestBase.cs
```diff
@@ -123,5 +123,6 @@ protected async Task RunActionWithTimeout(Func<CancellationToken, Task> action)
                 await actionTask;
             }
         }
+
```
martincostello, line 126:
```suggestion
```

tests/JustSaying.TestingFramework/TrackingLoggingMonitor.cs
```diff
@@ -11,7 +11,7 @@ public class TrackingLoggingMonitor : IMessageMonitor
     {
         private readonly ILogger _logger;
 
-        public TrackingLoggingMonitor(ILogger logger)
+        public TrackingLoggingMonitor(ILogger<IMessageMonitor> logger)
```
martincostello, line 14:
```suggestion
        public TrackingLoggingMonitor(ILogger<TrackingLoggingMonitor> logger)
```

tests/JustSaying.UnitTests/JustSayingBus/WhenPublishingMessageWithoutMonitor.cs
```diff
@@ -40,7 +40,7 @@ public void SettingANullMonitorSetsTheMonitorToNullOpMonitor()
         [Fact]
         public void SettingANewMonitorIsAccepted()
         {
-            SystemUnderTest.Monitor = new TrackingLoggingMonitor(NullLogger.Instance);
+            SystemUnderTest.Monitor = new TrackingLoggingMonitor(NullLogger<IMessageMonitor>.Instance);
```
martincostello, line 43:
```suggestion
            SystemUnderTest.Monitor = new TrackingLoggingMonitor(NullLogger<TrackingLoggingMonitor>.Instance);
```

tests/JustSaying.UnitTests/Messaging/Channels/MessageReceiveBufferTests/WhenSqsIsSlow.cs
```diff
@@ -35,7 +36,7 @@ public WhenSqsIsSlow(ITestOutputHelper testOutputHelper)
             queue.Uri.Returns(new Uri("http://test.com"));
             queue.Client.Returns(sqsClient);
             var monitor = new TestingFramework.TrackingLoggingMonitor(
-                loggerFactory.CreateLogger("Monitor"));
+                loggerFactory.CreateLogger<IMessageMonitor>());
```
martincostello, line 39:
```suggestion
                loggerFactory.CreateLogger<TrackingLoggingMonitor>());
```

tests/JustSaying.UnitTests/Messaging/Channels/MessageReceiveBufferTests/WhenSubscriberIsSlow.cs
```diff
@@ -34,7 +35,7 @@ public WhenSubscriberIsSlow(ITestOutputHelper testOutputHelper)
             queue.Uri.Returns(new Uri("http://test.com"));
             queue.Client.Returns(sqsClient);
             var monitor = new TestingFramework.TrackingLoggingMonitor(
-                loggerFactory.CreateLogger("Monitor"));
+                loggerFactory.CreateLogger<IMessageMonitor>());
```
martincostello, line 38:
```suggestion
                loggerFactory.CreateLogger<TrackingLoggingMonitor>());
```

tests/JustSaying.UnitTests/Messaging/Channels/MessageReceiveBufferTests/WhenThereAreNoSubscribers.cs
```diff
@@ -38,7 +39,7 @@ public WhenThereAreNoSubscribers(ITestOutputHelper testOutputHelper)
             queue.Uri.Returns(new Uri("http://test.com"));
             queue.Client.Returns(sqsClient);
             var monitor = new TestingFramework.TrackingLoggingMonitor(
-                loggerFactory.CreateLogger("Monitor"));
+                loggerFactory.CreateLogger<IMessageMonitor>());
```
martincostello, line 42:
```suggestion
                loggerFactory.CreateLogger<TrackingLoggingMonitor>());
```

tests/JustSaying.UnitTests/Messaging/Channels/SubscriptionGroupTests/BaseSubscriptionGroupTests.cs
```diff
@@ -62,7 +63,7 @@ private void GivenInternal()
         {
             Queues = new List<ISqsQueue>();
             Handler = new InspectableHandler<SimpleMessage>();
-            Monitor = new TrackingLoggingMonitor(Logger);
+            Monitor = new TrackingLoggingMonitor(LoggerFactory.CreateLogger<IMessageMonitor>());
```
martincostello, line 66:
```suggestion
            Monitor = new TrackingLoggingMonitor(LoggerFactory.CreateLogger<TrackingLoggingMonitor>());
```

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -85,49 +85,49 @@ public override async Task<bool> ExistsAsync()
...
                 var response = await Client.CreateTopicAsync(new CreateTopicRequest(TopicName))
                     .ConfigureAwait(false);
 
-                if (!string.IsNullOrEmpty(response.TopicArn))
+                if (string.IsNullOrEmpty(response.TopicArn))
                 {
-                    Arn = response.TopicArn;
-                    _log.LogInformation("Created topic '{TopicName}' on ARN '{Arn}'.", TopicName, Arn);
-                    return true;
+                    _log.LogError("Failed to create topic or obtain ARN for topic {TopicName}", TopicName);
```
maurofranchi, line 97: I would not repeat the word `topic` as you are looking for the ARN of the created or existing topic. 
```suggestion
                    _log.LogError("Failed to create or obtain ARN for topic {TopicName}", TopicName);
```

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -85,49 +85,49 @@ public override async Task<bool> ExistsAsync()
...
+                    throw new Exception($"Failed to create topic or receive ARN '{TopicName}'.");
```
maurofranchi, line 98: Why do you want to have a different error message?
```suggestion
                    throw new Exception($"Failed to create or obtain ARN for topic '{TopicName}'.");
```

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -85,49 +85,49 @@ public override async Task<bool> ExistsAsync()
...
-
-            return created;
         }
 
         private async Task<ServerSideEncryption> ExtractServerSideEncryptionFromTopicAttributes()
         {
             var attributesResponse = await Client.GetTopicAttributesAsync(Arn).ConfigureAwait(false);
 
-            if (!attributesResponse.Attributes.TryGetValue(JustSayingConstants.AttributeEncryptionKeyId, out var encryptionKeyId))
+            if (!attributesResponse.Attributes.TryGetValue(JustSayingConstants.AttributeEncryptionKeyId,
```
maurofranchi, line 129:
```suggestion
            if (!attributesResponse.Attributes.TryGetValue(
                JustSayingConstants.AttributeEncryptionKeyId,
```

gkinsman: Not sure this one makes sense - I'm formatting with the editorconfig settings and this line doesn't pass the char limit.

src/JustSaying/JustSayingBus.cs
gkinsman: Done 👍

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
gkinsman: Done 👍

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
gkinsman: Fixed it up so they're the same 👍

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
martincostello: Maybe just remove the "length" rule. It's causing so much code churn and making it less readable IMHO.

gkinsman: I think the churn is because we have many files that haven't been formatted before, so as we touch them and format them as they change. 

We could possibly do a global format, though it would likely invalidate Stu's other PR.

I think we're going to get churn with auto formatting rules regardless until everything is formatted with the ruleset. Personally I feel that auto formatting is a better solution than ad-hoc fixes in every PR.

martincostello: If it were, I wouldn't auto-format the files and only commit the changes I'm intentionally making.

gkinsman: I'm curious what your workflow is, as I'm keen to find a solution here! Do you use auto-format? If so is it just for the code you change? What if it's surrounded by unformatted code? 

If we were in a utopia where everything was already formatted, then it wouldn't really be a problem as the only changes that would exist after an auto-format would be the intentional changes.

Can we get there somehow? Do we want to?

martincostello: Visual Studio doesn't appear to be super-aggressive with formatting the whole document you're working on (VS Code is, for example) unless you specifically ask it to, and that's what I use. I don't have ReSharper either, I've never liked it.

Then I use GitHub Desktop as my Git client, and I manually deselect (or undo in the GHD UI) lines that aren't anything to do with what I'm trying to change.

I might bring a handful of fix-ups with me, but I won't generally do wholesale changes when others have to review it, as it just adds extra friction to the process of getting the PR over the line for merging. Otherwise you can end-up with a snowball effect where a few simple changes end up as a giant PR, with each bit of feedback being addressed changing more and more formatting.

If we want to make every single file be 100% compliant with the EditorConfig file, then that can be done separately as a dedicated piece of work if that's something we feel is worth doing.

JustSaying is 6+ years old and I'm not sure churning the whole codebase to hit an arbitrary formatting Nirvana is worth the time and investment given its used less these days with many workers being replaced with Lambdas with SQS triggers.

For my 2 cents, the EditorConfig is just there as a preference on entirely new code and edits to existing code, rather than a set of rules that must be obeyed. If we wanted formatting _rules_ I'd opt for StyleCop (which I use in my own projects).

maurofranchi: > Not sure this one makes sense - I'm formatting with the editorconfig settings and this line doesn't pass the char limit.

I simply suggested the change because on line 162 you put all the parameters on a separate line, so, for consistency, I thought you should do the same on this too. I didn't want to create a long thread about formatting conventions 😅

gkinsman: Thanks @martincostello, I appreciate your perspective. I'll avoid changing formatting going forward, and maybe consider doing a global format but as you say, there might not be a huge benefit. I tend to compulsively format on save in rider, so it can be a challenge to contribute to code bases without auto format rules setup 😅.

Thanks for your time and patience!

gkinsman: As for this specific comment, I've moved the parameter to the next line, so will resolve.

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -126,7 +126,8 @@ private async Task<ServerSideEncryption> ExtractServerSideEncryptionFromTopicAtt
         {
             var attributesResponse = await Client.GetTopicAttributesAsync(Arn).ConfigureAwait(false);
 
-            if (!attributesResponse.Attributes.TryGetValue(JustSayingConstants.AttributeEncryptionKeyId,
+            if (!attributesResponse.Attributes.TryGetValue(
+                JustSayingConstants.AttributeEncryptionKeyId,
                 out var encryptionKeyId))
```
martincostello:
```suggestion
                    JustSayingConstants.AttributeEncryptionKeyId,
                    out var encryptionKeyId))
```

gkinsman:
Is this ready to merge @martincostello? 🙂

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -85,49 +85,50 @@ public override async Task<bool> ExistsAsync()
...
                     .ConfigureAwait(false);
 
-                if (!string.IsNullOrEmpty(response.TopicArn))
+                if (string.IsNullOrEmpty(response.TopicArn))
                 {
-                    Arn = response.TopicArn;
-                    _log.LogInformation("Created topic '{TopicName}' on ARN '{Arn}'.", TopicName, Arn);
-                    return true;
+                    _log.LogError("Failed to create or obtain ARN for topic {TopicName}.", TopicName);
+                    throw new Exception($"Failed to create or obtain ARN for topic '{TopicName}'.");
```
martincostello, line 98: Let's use `InvalidOperationException` instead of `Exception`.

martincostello: Is there more context that can be added to the message here? Like something about the ARN not being retrieved?

src/JustSaying/AwsTools/QueueCreation/AmazonQueueCreator.cs
```diff
@@ -109,13 +108,27 @@ public QueueWithAsyncStartup<SqsQueueByName> EnsureQueueExists(string region, Sq
...
         private static async Task SubscribeQueueAndApplyFilterPolicyAsync(
             IAmazonSimpleNotificationService amazonSimpleNotificationService,
-            string topicArn, IAmazonSQS amazonSQS, Uri queueUrl, string filterPolicy)
+            string topicArn,
+            IAmazonSQS amazonSQS,
+            Uri queueUrl,
+            string filterPolicy)
         {
-            var subscriptionArn = await amazonSimpleNotificationService.SubscribeQueueAsync(topicArn, amazonSQS, queueUrl.AbsoluteUri)
+            if (amazonSimpleNotificationService == null)
```
martincostello, line 116: Don't necessarily have to change this, but I'm not a fan of the "line width" rule creating different indentation style for all these guard clauses.

Would be better either to have them all with braces, all indented below, or all on the same line. Having two different styles for 4 lines of code isn't great.

gkinsman: Yeah I think you're right, the line limit has started annoying me a bit too. Will turf it 👍

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
gkinsman: Should we be throwing an InvalidOperationException for something the developer can't fix?

We can add the request id that comes back, but there's no error message or anything aside from a map of repsonse metadata. 

The message could be changed to be something closer to retrieval failure?

src/JustSaying/AwsTools/QueueCreation/AmazonQueueCreator.cs
gkinsman: Have removed the line wrap rule, and fixed up these checks 👍

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
gkinsman: Have changed this a bit, let me know what you think.

src/JustSaying/AwsTools/QueueCreation/AmazonQueueCreator.cs
```diff
@@ -113,12 +113,10 @@ private static bool TopicExistsInAnotherAccount(SqsReadConfiguration queueConfig
...
             string filterPolicy)
         {
-            if (amazonSimpleNotificationService == null)
-                throw new ArgumentNullException(nameof(amazonSimpleNotificationService));
+            if (amazonSimpleNotificationService == null) throw new ArgumentNullException(nameof(amazonSimpleNotificationService));
             if (amazonSQS == null) throw new ArgumentNullException(nameof(amazonSQS));
             if (queueUrl == null) throw new ArgumentNullException(nameof(queueUrl));
-            if (string.IsNullOrEmpty(topicArn))
-                throw new ArgumentException("topicArn cannot be null or empty.", nameof(topicArn));
+            if (string.IsNullOrEmpty(topicArn)) throw new ArgumentException("topicArn cannot be null or empty.", nameof(topicArn));
```
martincostello, line 119:
```suggestion
            if (string.IsNullOrEmpty(topicArn)) throw new ArgumentException($"{nameof(topicArn)} cannot be null or empty.", nameof(topicArn));
```

martincostello (commented):
Final comment 😄

maurofranchi (approved):
👍

martincostello (approved):

PR #808, opened 2020-10-16
Add documentation for IMessageMonitor

yetanotherchris:
This adds XML documentation to `IMessageMonitor`.

### DIFF
--- a/src/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
+++ b/src/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
@@ -3,16 +3,69 @@ using Amazon.SQS.Model;
 
 namespace JustSaying.Messaging.Monitoring
 {
+    /// <summary>
+    /// Provides a mechanism for monitoring when various message actions are
+    /// performed in JustSaying. This is typically used for metrics gathering.
+    /// </summary>
     public interface IMessageMonitor
     {
+        /// <summary>
+        /// Monitors when a specific message type threw an exception while being handled
+        /// by a subscriber.
+        /// </summary>
+        /// <param name="messageType">The type of the message that was being handled.</param>
         void HandleException(Type messageType);
+
+        /// <summary>
+        /// Monitors when a particular message that threw an exception while being handled
+        /// by a subscriber.
+        /// </summary>
+        /// <param name="ex">The exception that was thrown.</param>
+        /// <param name="message">The message being handled when the exception was thrown.</param>
         void HandleError(Exception ex, Message message);
+
+        /// <summary>
+        /// Monitors the length of time a message took to be handled by a subscriber.
+        /// </summary>
+        /// <param name="duration">The length of time taken.</param>
         void HandleTime(TimeSpan duration);
+
+        /// <summary>
+        /// Monitors when a message was not successfully published after waiting for
+        /// the <see cref="IPublishConfiguration.PublishFailureReAttempts">configured amount of retries</see>.
+        /// </summary>
         void IssuePublishingMessage();
+
+        /// <summary>
+        /// Monitors when each message is handled by a subscriber.
+        /// </summary>
+        /// <param name="message">The message that was handled.</param>
         void Handled(JustSaying.Models.Message message);
+
+        /// <summary>
+        /// Monitors each time a message took over 1ms to be written to a ChannelWriter.
+        /// </summary>
         void IncrementThrottlingStatistic();
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


src/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
```diff
@@ -3,16 +3,69 @@
...
+    /// performed in JustSaying. This is typically used for metrics gathering.
+    /// </summary>
     public interface IMessageMonitor
     {
+        /// <summary>
+        /// Monitors when a specific message type threw an exception while being handled
+        /// by a subscriber.
+        /// </summary>
+        /// <param name="messageType">The type of the message that was being handled.</param>
         void HandleException(Type messageType);
```
gkinsman, line 17: I think in most other xml comments we stick to present tense, so maybe this could be re-phrased as:

`Called when a specific message type throws an exception while being handled by a <see cref="IHandlerAsync<T>" />.` ?

src/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
```diff
@@ -3,16 +3,69 @@
...
+
+        /// <summary>
+        /// Monitors when a particular message that threw an exception while being handled
+        /// by a subscriber.
+        /// </summary>
+        /// <param name="ex">The exception that was thrown.</param>
+        /// <param name="message">The message being handled when the exception was thrown.</param>
         void HandleError(Exception ex, Message message);
```
gkinsman, line 25: Similar again language-wise, though this method is called if something goes wrong in the dispatcher when deserialising the `TMessage` , so we don't have a TMessage to call `HandleException` with.

Perhaps:

`Called when an error occurs while attempting to deserialise a message from the `Amazon.SQS.Model.Message.` ?

The `message` parameter in this case is actually the raw SQS message and not a JustSaying model.

src/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
```diff
@@ -3,16 +3,69 @@
...
+
+        /// <summary>
+        /// Monitors the length of time a message took to be handled by a subscriber.
```
gkinsman, line 28: `Called when a handler has completed, with the time taken by the handler.` ?

src/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
```diff
@@ -3,16 +3,69 @@
...
         void IssuePublishingMessage();
+
+        /// <summary>
+        /// Monitors when each message is handled by a subscriber.
+        /// </summary>
+        /// <param name="message">The message that was handled.</param>
         void Handled(JustSaying.Models.Message message);
+
+        /// <summary>
+        /// Monitors each time a message took over 1ms to be written to a ChannelWriter.
```
gkinsman, line 46: We perhaps don't want to be too specific about the implementation details of how things are throttled. Would it be enough to just say something like `Called when a message pipeline is full, and a receiver must wait until there is room to continue downloading messages`?

src/JustSaying/Messaging/Monitoring/IMessageMonitor.cs
```diff
@@ -3,16 +3,69 @@
...
+        /// </summary>
         void IncrementThrottlingStatistic();
+
+        /// <summary>
+        /// Monitors the length of time a message took to be written to a ChannelWriter,
+        /// if that duration was longer than 1ms.
+        /// </summary>
+        /// <param name="duration">The length of time the write took.</param>
         void HandleThrottlingTime(TimeSpan duration);
```
gkinsman, line 55: This is the same as above, just that we actually pass a duration in here.

I kinda wonder if we should also be passing in the queue/message that is being throttled, else I'm not sure this stat has much value...

martincostello:
Could you use the review functionality rather than lots of individual _Add single comment_ uses? It means I only get one email for a code review instead of N 😄

gkinsman:
> Could you use the review functionality rather than lots of individual _Add single comment_ uses? It means I only get one email for a code review instead of N 😄

Ah sure, will do from now 🙂

yetanotherchris:
@gkinsman I've updated, I've kept to the use of "subscriber" now rather than receiver to keep it obvious which are subscriber and which are publisher related.

maurofranchi (approved):

gkinsman:
Sorry for the delay on this one @yetanotherchris, always appreciate documentation updates! Thanks 🙂

gkinsman (approved):

PR #809, opened 2020-10-20
Implement Middleware for Handlers

gkinsman:
#### This PR implements a middleware layer around handlers. Middlewares have a number of advantages over existing solutions:

- User provided middlewares can be provided via extension methods packages in nuget packages
- They are composable with other middlewares
- They don't require any custom code in or around `IHandlerAsync<T>`
- They are cancellable with `CancellationToken`, whereas `IHandlerAsync<T>` cannot change to add one (or at least we haven't decided to do that)
- They create a nice place for users to add in cross cutting code, like custom stats, retry strategies, locking, etc etc without needing to change JustSaying.

This work is largely based on the existing middleware code added by @maisiesadler that is used in the `MessageReceiveBuffer`'s receive pipeline.

### Some example code

[Middlewares can be pulled from DI, or passed in explicitly](https://github.com/gkinsman/JustSaying/blob/31502b59ff0544c22d04ae6e1dead7f3a8c8e717/tests/JustSaying.IntegrationTests/Fluent/Subscribing/WhenUsingMultipleMiddlewares.cs#L35-L41):
```csharp
var callRecord = new List<string>();

void Before(string id) => callRecord.Add($"Before_{id}");
void After(string id) => callRecord.Add($"After_{id}");

var outerMiddleware = new TrackingMiddleware("outer", Before, After);
var middleMiddleware = new TrackingMiddleware("middle", Before, After);
var innerMiddleware = new TrackingMiddleware("inner", Before, After);

var services = GivenJustSaying()
    .AddSingleton(outerMiddleware)
    .AddSingleton<IHandlerAsync<SimpleMessage>>(handler)
    .ConfigureJustSaying(builder => builder.WithLoopbackTopic<SimpleMessage>(UniqueName,
        topic => topic.WithMiddlewareConfiguration(app =>
        {
            app.Use<TrackingMiddleware>(); // from DI
            app.Use(() => middleMiddleware); // provide a Func<MiddlewareBase<HandleMessageContext, bool>
            app.Use(innerMiddleware); // Existing instance
        })));
```

Produces:
```
Before_outer
Before_middle
Before_inner
After_inner
After_middle
After_outer
```

[Implementation of an extension method that makes a `UseExactlyOnce` method available to the pipeline, and resolves some services from DI](https://github.com/gkinsman/JustSaying/blob/770a129bbb590be5a795c514cf63235d29f237d7/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs#L23-L41)

```csharp
public static HandlerMiddlewareBuilder UseExactlyOnce<TMessage>(
    this HandlerMiddlewareBuilder builder,
    string lockKey,
    TimeSpan? lockDuration = null)
{
    HandleMessageMiddleware CreateMiddleware()
    {
        var messageLock = builder.ServicesBuilder?.MessageLock?.Invoke() ?? builder.ServiceResolver.ResolveService<IMessageLockAsync>();
        var logger = builder.ServiceResolver.ResolveService<ILogger<ExactlyOnceMiddleware<TMessage>>>();

        return new ExactlyOnceMiddleware<TMessage>(messageLock,
            lockDuration ?? TimeSpan.MaxValue,
            lockKey,
            logger);
    }

    builder.Use(CreateMiddleware);

    return builder;
}
```

[Usage of `UseExactlyOnce`](https://github.com/gkinsman/JustSaying/blob/3f71f09917074881e678e573bf1cedeeb73130e8/tests/JustSaying.IntegrationTests/Fluent/Subscribing/WhenHandlerIsDeclaredAsExactlyOnce.cs#L29-L33)

```csharp
.ConfigureJustSaying((builder) =>
    builder.WithLoopbackTopic<SimpleMessage>(UniqueName,
        c =>
            c.WithMiddlewareConfiguration(m =>
                m.UseExactlyOnce<SimpleMessage>("simple-message-lock"))))
```

[They can be chained](https://github.com/gkinsman/JustSaying/blob/770a129bbb590be5a795c514cf63235d29f237d7/tests/JustSaying.UnitTests/Messaging/Middleware/HandlerMiddlewareBuilderTests.cs#L34-L56):
```csharp
.Configure(pipe =>
                {
                    pipe.Use(m1);
                    pipe.Use(m2);
                    pipe.Use(m3);
                })
```


src/JustSaying/Messaging/Middleware/Handle/HandlerInvocationMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+            if (context == null) throw new ArgumentNullException(nameof(context));
+
+            stoppingToken.ThrowIfCancellationRequested();
+
+            var resolutionContext = new HandlerResolutionContext(context.QueueName);
+
+            IHandlerAsync<T> handler = _handlerResolver(resolutionContext);
+
+            //TODO: cache these
+            Type handlerType = handler.GetType();
```
slang25, line 32: Can we instead use `var task = handler.Handle(context.MessageAs<T>());` and flow the generic constraint?

gkinsman: Ahh, yes we can! I did it that way initially but then wanted to remove the generic parameter from `HandlerInvocationMiddleware<T>` for some reason, but that turned out to not be necessary.

Great stuff!

gkinsman: Have fixed this up 👍

gkinsman:
@martincostello can we target this for post-beta? I'm really keen to get the beta process rolling so we can get some features using the new backend.

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using System.Runtime.CompilerServices;
+using JustSaying.Messaging.MessageHandling;
+using JustSaying.Messaging.Middleware.ExactlyOnce;
+using JustSaying.Models;
+using Microsoft.Extensions.Logging;
+
+using HandleMessageMiddleware = JustSaying.Messaging.Middleware.MiddlewareBase<JustSaying.Messaging.Middleware.Handle.HandleMessageContext, bool>;
+
+
```
martincostello, line 10:
```suggestion
using HandleMessageMiddleware = JustSaying.Messaging.Middleware.MiddlewareBase<JustSaying.Messaging.Middleware.Handle.HandleMessageContext, bool>;

```

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace JustSaying.Messaging.Middleware.Handle
+{
+    public static class HandlerMiddlewareBuilderExtensions
+    {
+        public static HandlerMiddlewareBuilder UseHandler<TMessage>(this HandlerMiddlewareBuilder builder,
+            Func<HandlerResolutionContext, IHandlerAsync<TMessage>> handler) where TMessage : Message
+        {
+            if(builder == null) throw new ArgumentNullException(nameof(builder));
+            if(handler == null) throw new ArgumentNullException(nameof(handler));
```
martincostello, line 19:
```suggestion
            if (builder == null) throw new ArgumentNullException(nameof(builder));
            if (handler == null) throw new ArgumentNullException(nameof(handler));
```

martincostello:
> @martincostello can we target this for post-beta? I'm really keen to get the beta process rolling so we can get some features using the new backend.

We can have as many alpha/beta/gamma/whatevers as we want before release internally, and we have the ProGet feed we can host packages in, so I don't think it really matters what label we give it or whatever the release cadence is (or the need to be super-strict on SemVer).

I'd only worry about these distinctions for putting something in NuGet.org, and I think we'd want to be fairly confidently "done" and happy it's working before publishing a beta there.

gkinsman:
> > @martincostello can we target this for post-beta? I'm really keen to get the beta process rolling so we can get some features using the new backend.
> 
> We can have as many alpha/beta/gamma/whatevers as we want before release internally, and we have the ProGet feed we can host packages in, so I don't think it really matters what label we give it or whatever the release cadence is (or the need to be super-strict on SemVer).
> 
> I'd only worry about these distinctions for putting something in NuGet.org, and I think we'd want to be fairly confidently "done" and happy it's working before publishing a beta there.

That makes perfect sense, thanks 👍. The main reason for a beta was so that we can switch to a package feed instead of local packages, so as you say we can just use proget 🙂.

src/JustSaying/Fluent/ISubscriptionBuilder`1.cs
```diff
@@ -26,7 +26,9 @@ internal interface ISubscriptionBuilder<out T>
         void Configure(
             JustSayingBus bus,
             IHandlerResolver resolver,
+            IServiceResolver serviceResolver,
```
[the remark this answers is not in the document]
gkinsman: Yeah I agree that having both resolvers plus the ServicesBuilder here isn't ideal. We could perhaps introduce an `IMiddlewareResolver`, though that adds a bit more machinery for questionable gain.

The ServicesBuilder is needed because we special case the registration/resolution of `IMessageLockAsync` because of JustSaying's explicit support for ExactlyOnce. We could remove this dependency by removing the `WithMessageLock` API and just requiring the user to register an `IMessageLockAsync` into the DI container, which would be resolvable in the middleware.

I suspect we might need to keep `IHandlerResolver` and `IServiceResolver` though, as a custom `IHandlerResolver` could be provided even if the user uses a DI container - we can't just resolve the `IHandlerResolver` from the `IServiceResolver`

src/JustSaying/Messaging/Middleware/Handle/HandleMessageContext.cs
```diff
@@ -0,0 +1,23 @@
...
+            QueueName = queueName;
+        }
+
+        public string QueueName { get; }
+
+        public Type MessageType { get; }
+
+        public Message Message { get; }
+
+        public TMessage MessageAs<TMessage>() where TMessage : Message => Message as TMessage;
```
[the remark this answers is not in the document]
gkinsman: Have moved this out to an extension method. I think it's a useful hint to middleware authors that it's possible to pull out a Message of type `T` (which is available on the middleware).

src/JustSaying/Fluent/ISubscriptionBuilder`1.cs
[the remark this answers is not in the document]
gkinsman: I had a punt at doing this over here https://github.com/gkinsman/JustSaying/pull/1

src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs
```diff
@@ -157,13 +159,13 @@ public class MessageDispatcher : IMessageDispatcher
...
             var messageType = message.GetType();
 
-            var handler = _handlerMap.Get(queueName, messageType);
+            var middleware = _middlewareMap.Get(queueName, messageType);
 
-            if (handler == null)
+            if (middleware == null)
             {
                 _logger.LogError(
                     "Failed to dispatch. Handler for message of type '{MessageTypeName}' not found in handler map.",
```
adammorr, line 171: Do you think this log message needs updating now that we're using middleware instead?

adammorr: Just re-read your description, I'll hold off on these things until it's ready for review

src/JustSaying/Fluent/ISubscriptionBuilder`1.cs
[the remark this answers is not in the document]
gkinsman: Considering this resolved 👍

gkinsman:
This is now ready for review! I've documented the public interfaces, added tests, and removed the old exactly-once and stopwatch handlers and reimplemented as middlewares.

tests/JustSaying.TestingFramework/JustSaying.TestingFramework.csproj
```diff
@@ -11,7 +11,7 @@
     <PackageReference Include="MartinCostello.Logging.XUnit" Version="0.1.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.8" />
     <PackageReference Include="NSubstitute" Version="4.2.2" />
-    <PackageReference Include="Shouldly" Version="4.0.0" />
+    <PackageReference Include="Shouldly" Version="3.0.2" />
```
gkinsman, line 14: I downgraded this because v4 uses a new approval test engine called DiffEngine that triggers a type load exception for some reason, in a dependent lib called EmptyFiles (https://www.nuget.org/packages/EmptyFiles/). I couldn't get to the bottom of it so downgraded for now.

gkinsman: Ah, this looks like an issue with `net5.0` and Shouldly 4.0 - it works in `netcoreapp3.1`. Will keep this downgraded till there's a fix.

slang25: I'm trying to recreate this so I can file a bug report, but haven't been able to yet, is there a commit in this PR I should use?

gkinsman: Hey 🙂, I managed to reproduce it before over here: https://github.com/gkinsman/ShouldlyRepro. It fails locally with `dotnet test` and in Rider, but succeeds when run in github actions - weird!

I was going to log a bug but hadn't quite narrowed down if it was my environment yet. Does that project build locally for you?

slang25: Thanks @gkinsman, that was really helpful: https://github.com/shouldly/shouldly/issues/717

src/JustSaying/JustSayingBus.cs
```diff
@@ -46,15 +49,9 @@ public IMessageMonitor Monitor
...
-        public IMessageLockAsync MessageLock
-        {
-            get => HandlerMap.MessageLock;
-            set => HandlerMap.MessageLock = value;
-        }
-
         public IMessageContextAccessor MessageContextAccessor { get; set; }
 
-        public HandlerMap HandlerMap { get; }
+        public MiddlewareMap MiddlewareMap { get; }
```
slang25, line 54: If we hid this from the public API we could make `MiddlewareMap` internal (we'd need to also make `MessageDispatcher` internal, which is probably a good idea).
Perhaps we evaluate what we can hide as a separate task outside of this PR.

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
```diff
@@ -0,0 +1,139 @@
...
+        ///
+        /// This is added automatically as the innermost handler to all pipelines, and doesn't need
+        /// to be called manually.
+        /// </summary>
+        /// <typeparam name="TMessage"></typeparam>
+        /// <returns>The current HandlerMiddlewareBuilder.</returns>
+        /// <exception cref="InvalidOperationException">
+        /// If a HandlerInvocationMiddleware already exists in this pipeline, it cannot be added again.
+        /// </exception>
+        public HandlerMiddlewareBuilder UseHandler<TMessage>() where TMessage : Message
```
slang25, line 87: `internal` perhaps?

src/JustSaying/Messaging/Middleware/Metrics/StopwatchMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using JustSaying.Messaging.Middleware.Handle;
+using JustSaying.Messaging.Monitoring;
+
+namespace JustSaying.Messaging.Middleware.Metrics
+{
+    /// <summary>
+    /// This middleware measures handlers' execution duration and reports the results to <see cref="IMessageMonitor"/>.
```
slang25, line 11:
```suggestion
    /// This middleware measures the handler's execution duration and reports the results to <see cref="IMessageMonitor"/>.
```
(this is measuring 1 handler, not multiple right?)

tests/JustSaying.TestingFramework/JustSaying.TestingFramework.csproj
slang25: We can bump this back to `4.0.1` now

slang25 (commented):
Some minor comments, overall a really good addition 😃

src/JustSaying/Messaging/Middleware/Metrics/StopwatchMiddleware.cs
gkinsman: Yep, think yours is clearer 👍

src/JustSaying/JustSayingBus.cs
gkinsman: We can make this internal while leaving the MessageDispatcher as is - MiddlewareMap doesn't have the same strange dependencies that HandlerMap had because it doesn't explicitly know about message locks anymore.

I think its still a good idea to review visibility separately - we might be able to hide some more things now that we couldn't before 👍

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
gkinsman: This has been changed to internal. If in the future we want to open up the whole pipeline for customisation (so e.g. you could use a delegate as a handler instead of an `IHandlerAsync`) then we can re-think how we supply this pipeline through the API, but this should get us started.

slang25 (dismissed):

src/JustSaying/AwsTools/MessageHandling/Dispatch/MiddlewareMap.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            if (queueName is null) throw new ArgumentNullException(nameof(queueName));
+            if (messageType is null) throw new ArgumentNullException(nameof(messageType));
+
+            return _middlewares.TryGetValue((queueName, messageType), out var middleware) ? middleware : null;
+        }
+
+        public InterrogationResult Interrogate()
+        {
+            var middlewares = _middlewares.Select(item =>
```
martincostello, line 89: You should `ToList()` this, otherwise it's not materialized when it's returned and could cause different results if called multiple times and/or `InvalidOperationException` to be thrown if an item is added to the dictionary mid-enumeration.

src/JustSaying/Fluent/SqsReadConfigurationBuilder.cs
```diff
@@ -20,12 +21,53 @@ public sealed class SqsReadConfigurationBuilder : SqsConfigurationBuilder<SqsRea
...
+        /// A sample configuration:
+        /// <code>
+        /// WithMiddlewareConfiguration(pipe =>
+        /// {
+        ///     pipe.Use&lt;SomeCustomMiddleware&gt;();
+        ///     pipe.Use&lt;SomeOtherCustomMiddleware&gt;();
+        /// });
+        /// </code>
+        /// would yield this order of execution:
+        /// Before_SomeCustomMiddleware
```
martincostello, line 54: Could format this as an HTML `<ul>`?

src/JustSaying/Fluent/SqsReadConfigurationBuilder.cs
```diff
@@ -20,12 +21,53 @@ public sealed class SqsReadConfigurationBuilder : SqsConfigurationBuilder<SqsRea
...
+        /// Before_SomeOtherCustomMiddleware
+        /// Before_StopwatchMiddleware
+        /// Before_HandlerInvocationMiddleware
+        /// After_HandlerInvocationMiddleware
+        /// After_StopwatchMiddleware
+        /// After_SomeOtherCustomMiddleware
+        /// After_SomeCustomMiddleware
+        /// </example>
+        /// <returns></returns>
```
martincostello, line 63: Missing returns value docs.

src/JustSaying/JustSaying.csproj
```diff
@@ -14,4 +14,7 @@
     <PackageReference Include="System.Text.Json" Version="4.6.0" />
     <PackageReference Include="System.Threading.Channels" Version="4.5.0" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Messaging\Middleware\Interrogation" />
+  </ItemGroup>
```
martincostello, line 19:
```suggestion
```

src/JustSaying/Messaging/Middleware/ExactlyOnce/ExactlyOnceMiddleware.cs
```diff
@@ -1,38 +1,38 @@
 using System;
+using System.Threading;
 using System.Threading.Tasks;
-using JustSaying.Models;
+using JustSaying.Messaging.MessageHandling;
+using JustSaying.Messaging.Middleware.Handle;
 using Microsoft.Extensions.Logging;
 
-namespace JustSaying.Messaging.MessageHandling
+namespace JustSaying.Messaging.Middleware.ExactlyOnce
```
martincostello, line 8: Does this need to live in its own namespace? Seems a bit OTT for one class and one extension method to me.

src/JustSaying/Messaging/Middleware/ExactlyOnce/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <param name="builder">The current <see cref="HandlerMiddlewareBuilder"/>.</param>
+        /// <param name="lockKey">A unique key to identify this lock, e.g. the name of the handler.</param>
+        /// <param name="lockDuration">The length of time to lock messages while handling them.</param>
+        /// <typeparam name="TMessage">The type of the message that should be locked.</typeparam>
+        /// <returns>The current <see cref="HandlerMiddlewareBuilder"/>.</returns>
+        public static HandlerMiddlewareBuilder UseExactlyOnce<TMessage>(
+            this HandlerMiddlewareBuilder builder,
+            string lockKey,
+            TimeSpan? lockDuration = null)
+        {
```
martincostello, line 23: Null check the builder (and the key?).

src/JustSaying/Messaging/Middleware/ExactlyOnce/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using JustSaying.Messaging.MessageHandling;
+using JustSaying.Messaging.Middleware.Handle;
+using Microsoft.Extensions.Logging;
+using HandleMessageMiddleware = JustSaying.Messaging.Middleware.MiddlewareBase<JustSaying.Messaging.Middleware.Handle.HandleMessageContext, bool>;
+
+namespace JustSaying.Messaging.Middleware.ExactlyOnce
```
martincostello, line 7: This API would be more discoverable if it was in the same namespace as the builder it's extending.

src/JustSaying/Messaging/Middleware/Handle/HandleMessageContext.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using JustSaying.Models;
+
+namespace JustSaying.Messaging.Middleware.Handle
+{
+    public sealed class HandleMessageContext
```
martincostello, line 6: `///` docs

src/JustSaying/Messaging/Middleware/Handle/HandleMessageContextExtensions.cs
```diff
@@ -0,0 +1,16 @@
+
+using System;
+using JustSaying.Models;
+
+namespace JustSaying.Messaging.Middleware.Handle
+{
+    public static class HandleMessageContextExtensions
```
martincostello, line 7: `///` docs

src/JustSaying/Messaging/Middleware/Handle/HandleMessageContextExtensions.cs
```diff
@@ -0,0 +1,16 @@
+
```
martincostello, line 1:
```suggestion
```

src/JustSaying/Messaging/Middleware/Handle/HandleMessageContextExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using JustSaying.Models;
+
+namespace JustSaying.Messaging.Middleware.Handle
+{
+    public static class HandleMessageContextExtensions
+    {
+        public static TMessage MessageAs<TMessage>(this HandleMessageContext context) where TMessage : Message
+        {
+            if(context == null) throw new ArgumentNullException(nameof(context));
```
martincostello, line 11:
```suggestion
            if (context == null) throw new ArgumentNullException(nameof(context));
```

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// <typeparam name="TMessage"></typeparam>
+        /// <returns>The current HandlerMiddlewareBuilder.</returns>
+        /// <exception cref="InvalidOperationException">
+        /// If a HandlerInvocationMiddleware already exists in this pipeline, it cannot be added again.
+        /// </exception>
+        internal HandlerMiddlewareBuilder UseHandler<TMessage>() where TMessage : Message
+        {
+            if (_handlerMiddleware != null)
+            {
+                throw new InvalidOperationException($"Handler middleware has already been specified " +
```
martincostello, line 91: Do we need to split this message over two lines?

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using JustSaying.Messaging.Middleware.ExactlyOnce;
+using JustSaying.Messaging.Middleware.Metrics;
+using JustSaying.Messaging.Monitoring;
+using JustSaying.Models;
+using Microsoft.Extensions.Logging;
+using HandleMessageMiddleware = JustSaying.Messaging.Middleware.MiddlewareBase<JustSaying.Messaging.Middleware.Handle.HandleMessageContext, bool>;
+
+namespace JustSaying.Messaging.Middleware.Handle
+{
+
```
martincostello, line 12:
```suggestion
```

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Adds a <see cref="HandlerInvocationMiddleware{T}"/> to the current pipeline.
+        /// </summary>
+        /// <param name="builder">The current <see cref="HandlerMiddlewareBuilder"/>.</param>
+        /// <param name="handler">A factory that creates <see cref="IHandlerAsync{T}"/> instances from
+        /// a <see cref="HandlerResolutionContext"/>.</param>
+        /// <typeparam name="TMessage">The type of the message that should be handled</typeparam>
+        /// <returns>The current <see cref="HandlerMiddlewareBuilder"/>.</returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="builder"/> is <see langword="null"/>.
```
martincostello, line 24:
```suggestion
        /// <paramref name="builder"/> or <paramref name="handler"/> is <see langword="null"/>.
```

src/JustSaying/Messaging/Middleware/Metrics/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using JustSaying.Messaging.Middleware.Handle;
+using JustSaying.Messaging.Monitoring;
+
+namespace JustSaying.Messaging.Middleware.Metrics
```
martincostello, line 5: Similar comments about discoverability putting everything in discrete namespaces.

src/JustSaying/Messaging/Middleware/Metrics/StopwatchMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+            _handlerType = handlerType;
+        }
+
+        protected override async Task<bool> RunInnerAsync(HandleMessageContext context, Func<CancellationToken, Task<bool>> func, CancellationToken stoppingToken)
+        {
+            var watch = Stopwatch.StartNew();
+
+            bool result = await func(stoppingToken).ConfigureAwait(false);
+
+            watch.Stop();
```
martincostello, line 36: The timing won't be recorded if the func throws, is that intentional?

tests/JustSaying.UnitTests/JustSaying.UnitTests.csproj
```diff
@@ -19,4 +19,9 @@
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Messaging\Channels\SubscriptionGroupTests\Support" />
+    <Folder Include="Messaging\Middleware\Approvals" />
+    <Folder Include="Messaging\Monitoring" />
+  </ItemGroup>
```
martincostello, line 26: Are these needed?

src/JustSaying/AwsTools/MessageHandling/Dispatch/MiddlewareMap.cs
gkinsman: Ah good catch, thanks 👍

src/JustSaying/Fluent/SqsReadConfigurationBuilder.cs
gkinsman: We could do - this is the appearance in rider(half with, half without):

image

I checked the formatting in VS and Rider, and had to change a few names to make it format correcly, but maybe a ul is a better way to go.

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
gkinsman: Moved it to a single line 👍

src/JustSaying/Messaging/Middleware/ExactlyOnce/ExactlyOnceMiddleware.cs
gkinsman: You might be right - I was thinking that it'd be good to keep these separate as we add more (e.g. a polly retry/rate limiting), but having separate namespaces is maybe a bit unnecessary.

I've moved everything into a base `JustSaying.Messaging.Middleware` namespace while retaining the folder structure to keep things tidy, which makes things a bit more discoverable as you noted below.

What do you think?

src/JustSaying/Messaging/Middleware/ExactlyOnce/HandlerMiddlewareBuilderExtensions.cs
gkinsman: Done 👍

src/JustSaying/Messaging/Middleware/Metrics/HandlerMiddlewareBuilderExtensions.cs
gkinsman: These are all now in the same namespace, `JustSaying.Messaging.Middleware`, and I've marked the folders as not contributing to namespaces to avoid the squiggly in jetbrains products.

src/JustSaying/Messaging/Middleware/Metrics/StopwatchMiddleware.cs
gkinsman: So this is the same logic as the previous handler, [found here](https://github.com/justeat/JustSaying/blob/master/src/JustSaying/Messaging/Monitoring/StopwatchHandler.cs).

I guess it makes sense if you assume that `IMessageMonitor` only calls `HandlerExecutionTime` for successful executions. I'm not sure if that's what I'd expect, but its the way it works now.

Should we change this?

tests/JustSaying.UnitTests/JustSaying.UnitTests.csproj
gkinsman: Nope, will remove 👍

src/JustSaying/Fluent/SqsReadConfigurationBuilder.cs
gkinsman: Have moved this to ul for consistent formatting across ide's

src/JustSaying/JustSaying.csproj.DotSettings
```diff
@@ -0,0 +1,4 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
martincostello, line 1: Maybe this should be in the `.gitignore`?

src/JustSaying/Messaging/Middleware/ExactlyOnce/ExactlyOnceHandlerMiddlewareBuilderExtensions.cs
```diff
@@ -21,6 +20,9 @@ public static class HandlerMiddlewareBuilderExtensions
             string lockKey,
             TimeSpan? lockDuration = null)
         {
+            if(builder == null) throw new ArgumentNullException(nameof(builder));
+            if(string.IsNullOrEmpty(lockKey)) throw new ArgumentException("Parameter cannot be null or empty", nameof(lockKey));
```
martincostello, line 24:
```suggestion
            if (builder == null) throw new ArgumentNullException(nameof(builder));
            if (string.IsNullOrEmpty(lockKey)) throw new ArgumentException("Parameter cannot be null or empty.", nameof(lockKey));
```

src/JustSaying/Messaging/Middleware/Metrics/StopwatchMiddleware.cs
martincostello: If it's consistent with existing behaviour, then that's fine.

src/JustSaying/JustSaying.csproj.DotSettings
gkinsman: It looks like this file is intended to be for user settings - I thought it was a project settings file. I've removed it and moved the rule exclusions into code using `// ReSharper disable once CheckNamespace` to avoid the squiggly.

martincostello (approved):

Issue #830, opened 2020-12-08
Add support for SNS tagging


adammorr:
**Is your feature request related to a problem? Please describe.**
n/a

**Describe the solution you'd like**
Add configuration support for tagging topics

**Describe alternatives you've considered**
Manually tag topics

**Additional context**
This feels like an AWS feature which could be nicely exposed through the topic configuration API and potentially used in a myriad of ways by consumers.

* [TagResource action](https://docs.aws.amazon.com/sns/latest/api/API_TagResource.html)
* [.NET SDK](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/index.html?page=SNS/MSNSTagResourceTagResourceRequest.html)

adammorr:
I'm going to pick this one up and see if I can get something suitable up for review

gkinsman:
Nice! Keen to see what you come up with 👍

adammorr:
PR raised https://github.com/justeat/JustSaying/pull/837

adammorr:
#837 merged

PR #837, opened 2020-12-21
ISSUE-830 Add support for tagging topics

adammorr:
This PR implements the suggestion in #830 which is to be able to tag SNS topics. 

I've only exposed this through the publication builder API for now as that feels the most appropriate place for this to happen.


slang25:
Do we want to look at tagging queues as well?

adammorr:
Yeah sure, I don't see it growing this PR too much

gkinsman:
Looks good! How do you imagine these being used? Is there some kind of convention that we could automatically apply that would be useful? E.g. we already know the component name, should we always add a `je-published-by` tag with that value to all published messages?

adammorr:
Yeah I was thinking whoever publishes to the topic as well as indicating if an event is sensitive in nature, part of a particular flow etc. I know of at least one internal use case where this would be helpful

adammorr:
@slang25 I'll raise queue tagging in a separate PR rather than bundling with this change now, in order to keep this one smallish

slang25 (approved):
Very nice @adammorr

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -14,8 +14,10 @@ namespace JustSaying.AwsTools.MessageHandling
 {
     public class SnsTopicByName : SnsTopicBase
     {
+        private readonly ILogger _logger;
+
         public string TopicName { get; }
-        private readonly ILogger _log;
+        public Dictionary<string, string> Tags { get; set; }
```
martincostello, line 20: Better to use `IDictionary<string, string>` than use the specific concrete type?

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -29,6 +30,11 @@ internal TopicPublicationBuilder()
         /// </summary>
         private Action<SnsWriteConfiguration> ConfigureWrites { get; set; }
 
+        /// <summary>
+        /// Gets the tags to add to the topic.
+        /// </summary>
+        private Dictionary<string, string> Tags { get; } = new(StringComparer.OrdinalIgnoreCase);
```
martincostello, line 36: Same here - the public API would be more lenient to allow an `IDictionary<string, string>` rather than enforce a specific concrete type.

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -71,6 +77,43 @@ public TopicPublicationBuilder<T> WithWriteConfiguration(Action<SnsWriteConfigur
...
         }
 
+        /// <summary>
+        /// Creates a tag with no value that will be assigned to the SNS topic.
+        /// </summary>
+        /// <param name="key">The key for the tag.</param>
+        /// <returns>
+        /// The current <see cref="TopicPublicationBuilder{T}"/>.
+        /// </returns>
+        /// <remarks>Tag keys are case-sensitive. A new tag with a key identical to that of an existing one will overwrite it.</remarks>
```
martincostello, line 87: The docs say the key is case sensitive, but you're using a case-insensitive comparer for the default `Tags` dictionary, meaning you could have only have the key "foo" though the docs imply you could have "foo", "Foo" and "fOo".

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -71,6 +77,43 @@ public TopicPublicationBuilder<T> WithWriteConfiguration(Action<SnsWriteConfigur
...
+
+        /// <summary>
+        /// Creates a tag with a value that will be assigned to the SNS topic.
+        /// </summary>
+        /// <param name="key">The key for the tag.</param>
+        /// <param name="value">The value associated with this tag.</param>
+        /// <returns>
+        /// The current <see cref="TopicPublicationBuilder{T}"/>.
+        /// </returns>
+        /// <remarks>Tag keys are case-sensitive. A new tag with a key identical to that of an existing one will overwrite it.</remarks>
```
martincostello, line 101: As above.

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -71,6 +77,43 @@ public TopicPublicationBuilder<T> WithWriteConfiguration(Action<SnsWriteConfigur
...
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="key"/> is <see langword="null"/> or whitespace.
+        /// </exception>
+        public TopicPublicationBuilder<T> WithTag(string key, string value)
+        {
+            if (string.IsNullOrWhiteSpace(key))
+            {
+                throw new ArgumentNullException(nameof(key));
```
martincostello, line 109: Shouldn't really throw `ArgumentNullException` if `key` isn't exactly `null`. Should either use `ArgumentException` for both cases and an add an appropriate message, or use `ArgumentNullException` for `null` and `ArgumentException` for white space.

tests/JustSaying.IntegrationTests/Fluent/Publishing/WhenRegisteringAPublisherWithTags.cs
```diff
@@ -0,0 +1,79 @@
...
+            : base(outputHelper)
+        { }
+
+        [NotSimulatorFact]
+        public async Task Then_A_Topic_Is_Created_With_The_Correct_Tags()
+        {
+            // Arrange
+            var tags = new Dictionary<string, string>
+            {
+                {
```
martincostello, line 29: Could make even terser as:

```csharp
[Guid.NewGuid().ToString()] = null,
[Guid.NewGuid().ToString()] = "Value",
```

tests/JustSaying.UnitTests/AwsTools/MessageHandling/Sns/TopicByName/WhenApplyingTags.cs
```diff
@@ -0,0 +1,76 @@
...
+            };
+
+            return Task.FromResult(topicByName);
+        }
+
+        protected override void Given()
+        {
+            _tags = new Dictionary<string, string>
+            {
+                {
```
martincostello, line 34: Similarly:

```csharp
["TagOne"] = "Tag-One",
["TagTwo"] = "Tag-Two",
```

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
```diff
@@ -66,14 +68,34 @@ public async Task EnsurePolicyIsUpdatedAsync(IReadOnlyCollection<string> additio
...
+
+            var tagRequest = new TagResourceRequest
+            {
+                ResourceArn = Arn,
+                Tags = Tags.Select(CreateTag).ToList()
+            };
+
+            await Client.TagResourceAsync(tagRequest).ConfigureAwait(false);
+
+            _logger.LogInformation("Added {TagCount} tags to topic {TopicName}", tagRequest.Tags.Count, TopicName);
```
martincostello, line 88: Maybe we should word this differently? If it had 3 tags before and you set a different 3, you'd still have 3 tags as the old 3 would have been removed rather than _adding_ 3 extra?

martincostello (commented):
Some comments you may or may not want to address in the PR for the SQS tagging.

adammorr:
@martincostello all sound sensible, I’ll address them in a separate PR before the SQS one, thanks for adding them 👍

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
adammorr: This is only added to as a result of calls to WithTag so would changing it be of any benefit here?

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
adammorr: 🤦 good point. I changed the docs after realising they’re case sensitive but not the actual type declaration

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
adammorr: I don’t believe old tags are removed. You’d just replace any that had the same key

adammorr:
https://github.com/justeat/JustSaying/pull/840 raised to address comments

src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
martincostello: Ah ok, I thought it was more of a "tags are these and these only" operation than additive.

PR #849, opened 2021-01-22
Add TimeStampWithOffset

gareththackeray:
In a global system, it is far better to be explicit about timestamps and include the offset to remove any ambiguity.

Ideally we could switch `TimeStamp` to use `DateTimeOffset` but this would be a breaking change which could have an impact on any message consumers that were expecting an offsetless datetime format.

### DIFF
--- a/src/JustSaying.Models/Message.cs
+++ b/src/JustSaying.Models/Message.cs
@@ -6,14 +6,16 @@ namespace JustSaying.Models
     {
         protected Message()
         {
-            TimeStampWithOffset = DateTimeOffset.Now;
-            TimeStamp = TimeStampWithOffset.UtcDateTime;            
+            TimeStamp = DateTime.UtcNow;            
             Id = Guid.NewGuid();            
         }
 
         public Guid Id { get; set; }
         public DateTime TimeStamp { get; set; }
-        public DateTimeOffset TimeStampWithOffset { get; set; }
+        //NOTE: 
+        //1. this is nullable because it is not present in older versions of JustSaying and we need versions to interop seamlessly
+        //2. we do not set it in the constructor, since it would mean when consuming messages that don't have this property it gets set spuriously
+        public DateTimeOffset? TimeStampWithOffset { get; set; }
         public string RaisingComponent { get; set; }
         public string Version { get; private set; }
         public string SourceIp { get; private set; }
--- a/src/JustSaying.Models/Message.cs
+++ b/src/JustSaying.Models/Message.cs
@@ -7,11 +7,13 @@ namespace JustSaying.Models
         protected Message()
         {
             TimeStamp = DateTime.UtcNow;
-            Id = Guid.NewGuid();
+            TimeStampWithOffset = DateTimeOffset.Now;
+            Id = Guid.NewGuid();            
         }
 
         public Guid Id { get; set; }
         public DateTime TimeStamp { get; set; }
+        public DateTimeOffset TimeStampWithOffset { get; set; }
         public string RaisingComponent { get; set; }
         public string Version { get; private set; }
         public string SourceIp { get; private set; }
--- a/src/JustSaying.Models/Message.cs
+++ b/src/JustSaying.Models/Message.cs
@@ -6,8 +6,8 @@ namespace JustSaying.Models
     {
         protected Message()
         {
-            TimeStamp = DateTime.UtcNow;
             TimeStampWithOffset = DateTimeOffset.Now;
+            TimeStamp = TimeStampWithOffset.UtcDateTime;            
             Id = Guid.NewGuid();            
         }
 
--- a/src/JustSaying/JustSayingBus.cs
+++ b/src/JustSaying/JustSayingBus.cs
@@ -247,6 +247,12 @@ namespace JustSaying
             CancellationToken cancellationToken)
         {
             attemptCount++;
+
+            if (message.TimeStampWithOffset == null)
+            {
+                message.TimeStampWithOffset = DateTimeOffset.Now;
+            }
+
             try
             {
                 using (Monitor.MeasurePublish())
--- a/tests/JustSaying.UnitTests/Messaging/Serialization/Newtonsoft/WhenSerializingAndDeserializing.cs
+++ b/tests/JustSaying.UnitTests/Messaging/Serialization/Newtonsoft/WhenSerializingAndDeserializing.cs
@@ -33,6 +33,7 @@ namespace JustSaying.UnitTests.Messaging.Serialization.Newtonsoft
             _messageOut.EnumVal.ShouldBe(_messageIn.EnumVal);
             _messageOut.RaisingComponent.ShouldBe(_messageIn.RaisingComponent);
             _messageOut.TimeStamp.ShouldBe(_messageIn.TimeStamp);
+            _messageOut.TimeStampWithOffset.ShouldBe(_messageIn.TimeStampWithOffset);
         }
 
         [Fact]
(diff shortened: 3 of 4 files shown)


src/JustSaying.Models/Message.cs
```diff
@@ -7,11 +7,13 @@ public abstract class Message
         protected Message()
         {
             TimeStamp = DateTime.UtcNow;
-            Id = Guid.NewGuid();
+            TimeStampWithOffset = DateTimeOffset.Now;
+            Id = Guid.NewGuid();
```
martincostello, line 11: I think it would be better to put this into a local, and then reference from there. Otherwise you have the possibility of having, albeit slight in magnitude, probably just a few ticks, differences in the value of the time in the two properties.

Something like:

```csharp
var now  = DateTimeOffset.Now;
TimeStampWithOffset = now;
TimeStamp = now.UtcDateTime;
```

gkinsman: Should we perhaps call this something like UtcTimeStamp? Or TimeStampUtc?

That it has an offset is a .net implementation detail - the key bit is as you say, we are explicitly declaring it a UTC timestamp.

gkinsman: Also, do we want to use DateTimeOffset.UtcNow? Or is this expected to be the servers Timex even if that's not in UTC?

gareththackeray: > Should we perhaps call this something like UtcTimeStamp? Or TimeStampUtc?
> 
> That it has an offset is a .net implementation detail - the key bit is as you say, we are explicitly declaring it a UTC timestamp.

Hmm I don't know - I think having an offset could be useful sometimes.  It gives you some extra informaion (what was the offset where this message was sent) without any loss.
We also can't guarantee it won't be overridden with a different offset.

Note that as currently written, it won't (necessarily) be "UTC" (i.e. offset 0)

martincostello: Renaming it from `TimeStamp` would be a breaking change.

I would agree that I'd prefer to just use UTC for both, rather than introduce two different ones that make assumptions about the time zone of the executing machine (which in prod for us will be UTC anyway).

I didn't bring that up initially though as I assumed there was offline conversation between Luke, Stu and Gareth that resulted in this PR using local times in the first place.

gareththackeray: Oh I see @gkinsman was referring to the existing timestamp.

The decision to use local was mine since, as I say it contains strictly more information.  Didn't discuss with Luke or Stu.

gkinsman: Apologies if I wasn't clear - I was referring to the new field. If it's going to be a `DateTimeOffset.UtcNow`, then we may as well be clear in the naming that it's a UTC DateTimeOffset?

martincostello: If I had to choose I'd just have it as the UTC offset (so `+00:00`). It's still more informative and it's more-or-less explicit, it just lacks a non-zero offset value, which for for non-UI isn't really a problem IMHO.

As "+01:00" in the UK in the summer is just an inference of UK local time to us, it could equally be the time in the Canary Islands. Without specifying the time zone's _name_ you can't be 100% sure of what the offset is specifically referring to anyway.

slang25:
First thoughts from me; we need to be really careful about adding non-nullable property to the `Message` base class as it will likely fair deserialization from older producers. It must be able to handle the case where the property is missing in JSON.

src/JustSaying.Models/Message.cs
gareththackeray: > If I had to choose I'd just have it as the UTC offset (so `+00:00`). It's still more informative and it's more-or-less explicit, it just lacks a non-zero offset value, which for for non-UI isn't really a problem IMHO.
> 
> As "+01:00" in the UK in the summer is just an inference of UK local time to us, it could equally be the time in the Canary Islands. Without specifying the time zone's _name_ you can't be 100% sure of what the offset is specifically referring to anyway.

Not sure I agree Martin.  The point is it could tells you what time it was for a relevant party at the time the performed the action, which is extra info.  I don't think the name of the timezone is important.

This is moot for us since as you say it will always be UTC and I'm happy to go either way, but local is my preference.  @stuart-lang any thoughts?

gareththackeray:
> First thoughts from me; we need to be really careful about adding non-nullable property to the `Message` base class as it might fail deserialization from older producers. It must be able to handle the case where the property is missing in JSON.

That's a good point.  And aside from the ability to deserialise, you have the scenario the sender doesn't have it on their schema and it deserialises ok but goes to its default value, with bugs on the consumer side ensuing.

I'll change it to nullable.

src/JustSaying.Models/Message.cs
gkinsman: I think we need to be a bit careful about adding fields to this base class, and only do so when we have a really good reason, and scenarios that make it really worthwhile.

To add some more data, let's consider Australia's timezones. Our servers are located in `ap-southeast-2`, which is in Sydney. Australia has three timezones, some of which apply DST, and others that do not, at different times of the year. Given that Menulog is available all over Australia, the local timezone that an order is placed in could be quite different to the timezone that the server that handled that request is in. In this case, using this new field for anything would be quite risky, as the local time could be either ahead or behind the server time - we would need to account for it in any usages of the field.

In the above scenario, if we did actually want to capture the local time where the order was placed, I think it would make more sense to be explicit and to use a `LocalOrderPlacedAt` DateTimeOffset field on whichever messages that fulfill that flow, which would be populated from an incoming request. 

I'm not flat out suggesting we don't do this, just that I think we need some solid scenarios and use cases to support its addition.

gareththackeray: The key reason for this field isn't to capture something about timezone, it's to ensure we have an unambiguous timestamp available on the message.  For JE it is really moot whether we use `Now` or `UtcNow` because the servers are set to UTC anyway.

src/JustSaying.Models/Message.cs
```diff
@@ -6,12 +6,14 @@ public abstract class Message
     {
         protected Message()
         {
-            TimeStamp = DateTime.UtcNow;
-            Id = Guid.NewGuid();
+            TimeStampWithOffset = DateTimeOffset.Now;
```
gareththackeray, line 9: There's a problem here.  If this is used as a base class on the receiving end and the publisher was using a version that didn't have the property, the ctor is going to set the value to Now and leave it like that.

I guess the publisher should set it if its null.  And for those features that don't use the JS publisher, their local publishers will just need to do likewise.

src/JustSaying.Models/Message.cs
martincostello: The offset is only useful for caring about that specific message at that point in time. You can't use the offset to infer future things from it, as the offset may change during the life of the order, which again makes me wonder how useful it is to have a non-UTC value.

The only truly unambiguous way to represent the time zone is to use its name (e.g. `Europe/London` in IANA format).

Sure you can say for a specific event that it happened at a wall-clock time of 09:52 to the user, but you can't rely on the offset to compute other timestamps relative to it (e.g. what is the wall-clock time an hour later), which is what makes me think you might as well just encode it as `+00:00` to make it clear it's UTC.

gareththackeray: It potentially makes things easier like "find all the events that happened between 9 and 5 local time".  I have a very weak opinion on this though.  If the consensus is for UTC I'll do that.  And it's 2 on 1 so far. 

Be good to get your thougts on my [comment below](https://github.com/justeat/JustSaying/pull/849#discussion_r562702600) as well.

src/JustSaying.Models/Message.cs
gkinsman: Perhaps we could leave it null in the constructor, and populate it in the library before we send it? That way we keep the init out of the Models lib, and provides a possible extension point for what you were thinking of with other users possibly using a different 'base' TZ.

src/JustSaying/JustSayingBus.cs
```diff
@@ -247,6 +247,12 @@ private IMessagePublisher GetPublisherForMessage(Message message)
             CancellationToken cancellationToken)
         {
             attemptCount++;
+
+            if (message.TimeStampWithOffset == null)
```
gkinsman, line 251: I wonder if we should move the set of the existing TimeStamp property here as well, so that both are populated at the same time. Given the models package is being versioned as well, we can make that change safely.

gareththackeray:
It turns out that, by default, `DateTime`s created using `DateTime.UtcNow` get serialised with a `Z` at the end to signify UTC.  So although a `DateTimeOffset` would be preferable in a couple of ways, there is no actual issue with the current state of affairs.  Closing PR.

PR #851, opened 2021-01-27
Reduce the log level of some frequent messages

gkinsman:
In testing JS7 with a feature that subscribes to 200 queues (due to tenanting), it turns out that we log quite a lot of information multiple times, because of the way startup works. I think these messages are still useful, but perhaps mainly for debugging purposes.

I propose the following messages are reduced from Information to Debug:

`2021-01-27 13:19:30.5446|Info|JustSaying|Checking if queue '<queuename>' exists completed in 00:00:00.8224467|`
`2021-01-27 13:19:30.5446|Info|JustSaying|Created topic '<topicname>' with ARN '<arn>'.|`
`2021-01-27 13:19:29.6835|Info|JustSaying.Messaging.Channels.Multiplexer.MergingMultiplexer|Starting up channel multiplexer with a queue capacity of 100|`

The first two can't be filtered by namespace, so there's no other way to filter them out. 

Note that we still log an interrogation message with queue names (though no arns), and lots of other information on concurrency/buffer/prefetch etc.

It's likely there are other low hanging fruit here too, but I think this is a start.

### DIFF
--- a/src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
+++ b/src/JustSaying/AwsTools/MessageHandling/SnsTopicByName.cs
@@ -123,7 +123,7 @@ namespace JustSaying.AwsTools.MessageHandling
                 }
 
                 Arn = response.TopicArn;
-                _logger.LogInformation("Created topic '{TopicName}' with ARN '{Arn}'.", TopicName, Arn);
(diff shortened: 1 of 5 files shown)


martincostello (approved):

gkinsman:
Thanks @martincostello!

PR #853, opened 2021-02-02
Move 'Succeeded handling...' log message to JustSaying.Dispatch namespace

gkinsman:
The log message with template `"{Status} handling message with Id '{MessageId}' of type {MessageType} in {TimeToHandle}ms."` is new in JS7, and we should allow users to opt out of it by putting it in a separate namespace from everything else that happens in JS. In this PR I've moved it to `JustSaying.Dispatch`, so that it can be ignored in configuration.

This is the only per-message information log event that I can see that's new in JS7, so it's worth making sure we don't hammer our log cluster by accident 😁. 

Perhaps this should even be off by default?


martincostello:
> Perhaps this should even be off by default?

Maybe it could be Debug?

PR #859, opened 2021-02-11
Make Queue and Topic Infrastructure Code Internal

slang25:
~~This change shouldn't alter any behavior~~, but should ease changes to this area, which we suspect we might want to do for a 7.1 release.


src/JustSaying/Fluent/QueueSubscriptionBuilder`1.cs
```diff
@@ -171,8 +172,7 @@ public QueueSubscriptionBuilder<T> WithTag(string key, string value)
             subscriptionConfig.Validate();
 
             var queue = creator.EnsureQueueExists(config.Region, subscriptionConfig);
-            bus.AddStartupTask(
-                queue.Queue.EnsureQueueAndErrorQueueExistAndAllAttributesAreUpdatedAsync(subscriptionConfig));
+            bus.AddStartupTask(queue.StartupTask);
```
slang25, line 175: So it turns out we were calling this method (`EnsureQueueAndErrorQueueExistAndAllAttributesAreUpdatedAsync`, what a mouth full!) and setting the `StartupTask` to resulting task, that might have resulted in weird behavior as we are running it twice in 2 different places and awaiting one of them.

This means that `IVerifyAmazonQueues` no longer needs to be generic, and the everything can be internalized. @gkinsman can you double-check my thinking here.

maurofranchi (approved):

src/JustSaying/AwsTools/MessageHandling/QueueWithAsyncStartup.cs
```diff
@@ -6,27 +6,26 @@ namespace JustSaying.AwsTools.MessageHandling
     /// Represents an <see cref="ISqsQueue"/> with an associated startup task that must be run before
     /// the queue is ready to be used.
     /// </summary>
-    /// <typeparam name="TQueue">The type of the SQS queue.</typeparam>
-    public class QueueWithAsyncStartup<TQueue> where TQueue : ISqsQueue
+    public class QueueWithAsyncStartup
     {
         /// <summary>
-        /// Creates an instance of <see cref="QueueWithAsyncStartup{TQueue}"/> for which the <see cref="TQueue"/>
+        /// Creates an instance of <see cref="QueueWithAsyncStartup"/> for which the <see cref="Queue"/>
```
martincostello, line 12: Should it be `ISqsQueue`?

```suggestion
        /// Creates an instance of <see cref="QueueWithAsyncStartup"/> for which the <see cref="ISqsQueue"/>
```

src/JustSaying/AwsTools/MessageHandling/QueueWithAsyncStartup.cs
```diff
@@ -6,27 +6,26 @@ namespace JustSaying.AwsTools.MessageHandling
...
-        public QueueWithAsyncStartup(TQueue queue)
+        public QueueWithAsyncStartup(ISqsQueue queue)
         {
             StartupTask = Task.CompletedTask;
             Queue = queue;
         }
 
         /// <summary>
-        /// Creates an instance of <see cref="QueueWithAsyncStartup{TQueue}"/> that requires the <see cref="TQueue"/>
+        /// Creates an instance of <see cref="QueueWithAsyncStartup"/> that requires the <see cref="Queue"/>
```
martincostello, line 23: +1

src/JustSaying/Properties/InternalsVisibleTo.cs
```diff
@@ -2,3 +2,4 @@
 
 [assembly: InternalsVisibleTo("JustSaying.IntegrationTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001009d30c64bc42ba5037aec3cf60334df3a4c140ffc1dab19fdd31c9765e4e29afa441396ff9437a764b618c6473a3b350c56d706a0fbe83915f1f250ee23eb548b30306187ac685f65caa48303dc86f08c1c24d99dc84966273c207eaa8570b440004c7f49cef4ec77bc69118610ae2b53db7d8abeb465cbcd4bd190feaf517aad")]
 [assembly: InternalsVisibleTo("JustSaying.UnitTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001009d30c64bc42ba5037aec3cf60334df3a4c140ffc1dab19fdd31c9765e4e29afa441396ff9437a764b618c6473a3b350c56d706a0fbe83915f1f250ee23eb548b30306187ac685f65caa48303dc86f08c1c24d99dc84966273c207eaa8570b440004c7f49cef4ec77bc69118610ae2b53db7d8abeb465cbcd4bd190feaf517aad")]
+[assembly: InternalsVisibleTo("JustSaying.Tools, PublicKey=00240000048000009400000006020000002400005253413100040000010001009d30c64bc42ba5037aec3cf60334df3a4c140ffc1dab19fdd31c9765e4e29afa441396ff9437a764b618c6473a3b350c56d706a0fbe83915f1f250ee23eb548b30306187ac685f65caa48303dc86f08c1c24d99dc84966273c207eaa8570b440004c7f49cef4ec77bc69118610ae2b53db7d8abeb465cbcd4bd190feaf517aad")]
```
martincostello, line 5:
```suggestion
[assembly: InternalsVisibleTo("JustSaying.Tools, PublicKey=00240000048000009400000006020000002400005253413100040000010001009d30c64bc42ba5037aec3cf60334df3a4c140ffc1dab19fdd31c9765e4e29afa441396ff9437a764b618c6473a3b350c56d706a0fbe83915f1f250ee23eb548b30306187ac685f65caa48303dc86f08c1c24d99dc84966273c207eaa8570b440004c7f49cef4ec77bc69118610ae2b53db7d8abeb465cbcd4bd190feaf517aad")]
[assembly: InternalsVisibleTo("JustSaying.UnitTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001009d30c64bc42ba5037aec3cf60334df3a4c140ffc1dab19fdd31c9765e4e29afa441396ff9437a764b618c6473a3b350c56d706a0fbe83915f1f250ee23eb548b30306187ac685f65caa48303dc86f08c1c24d99dc84966273c207eaa8570b440004c7f49cef4ec77bc69118610ae2b53db7d8abeb465cbcd4bd190feaf517aad")]
```

src/JustSaying/AwsTools/MessageHandling/QueueWithAsyncStartup.cs
slang25: I thought that referring to the  instance was more specific, however I could go either way

src/JustSaying/Properties/InternalsVisibleTo.cs
slang25: I was actually trying to put them in alphabetical order, but forgot how to alphabet 😆

src/JustSaying/Fluent/QueueSubscriptionBuilder`1.cs
gkinsman: Ahh, good catch! This is unintentional. Thinking about it again now, I'm not sure we should be starting the task in AmazonQueueCreator as it will start running the task in the background, when we really want to start running it when the bus is started. 

The intention here was for it to be a cold task, but I think I've been writing too much JS recently as C# doesn't do that 😅. We might need to change the startup task to be a `Func<Task>`.

src/JustSaying/AwsTools/MessageHandling/QueueWithAsyncStartup.cs
slang25: No, you're right, this is the constructor so the call site is concerned with what to pass in.

src/JustSaying/AwsTools/MessageHandling/QueueWithAsyncStartup.cs
```diff
@@ -9,7 +9,7 @@ namespace JustSaying.AwsTools.MessageHandling
     public class QueueWithAsyncStartup
     {
         /// <summary>
-        /// Creates an instance of <see cref="QueueWithAsyncStartup"/> for which the <see cref="Queue"/>
+        /// Creates an instance of <see cref="QueueWithAsyncStartup"/> for which the <see cref="queue"/>
```
martincostello: So should it be a `paramref` then?

gkinsman (dismissed):

martincostello (approved):

gkinsman (approved):

PR #877, opened 2021-03-25
Add a serializer in queue publication builder

gkinsman:
This fixes a bug in which we weren't adding a serializer when registering a queue for publications, so attempting to publish would throw an exception.


slang25 (approved):
Nice catch!

martincostello (approved):

gkinsman:
What confuses me here is why the queue publication tests work - they shouldn't but I can't yet see why!

gkinsman:
This was also fixed in #876

Issue #867, opened 2021-03-03
Improve per-message logging to make it more user friendly


gkinsman:
As part of making JustSaying as friendly as possible to use, we want to ensure our logs are useful, structured, and optional.

Currently, we have two log messages that we emit per message: one at publish time, and one at receive time.

### Publish time

If the message is being published to SNS, we use this template:
`"Published message: '{SnsSubject}' with content {SnsMessage} and request Id '{SnsRequestId}'"` with context properties:
`SnsMessage`: Full body of the SNS message
`SnsRequestId`: Request id of the AWS publish call
`SnsSubject`: Subject that we derive from the message type and set on the request

If the message is being published to SQS, we use this template:
`"Published message to queue '{QueueUrl}' with content '{MessageBody}'."` with context properties:
`QueueUrl`: URL of the queue we're directly publishing to
`MessageBody`: Full body of the SQS message

I don't think we should be logging the bodies of messages in these log messages - it will put a huge amount of strain on whatever logging infrastructure is in use, and leads to questions around GDPR, amongst others. If users want to do this, I think they should create their own middlewares that log out message bodies.

In addition, these log messages are both logged at `Information` level with the context/category being `JustSaying`, which makes it impossible to filter out. If we instead published these messages, still at information, but with the context `JustSaying.Publish`, then the logs can be filtered out using whatever API is available in the logging framework in use.

*Question*: Should we log SQS/SNS publishes under different categories? Publish/Send perhaps?
*Question*: Should we be using the same template for both SQS/SNS messages? I think so, but how do we accomodate the extra info?

### Receive time

When we handle a message in the dispatcher, we log a message with the template 
`"{Status} handling message with Id '{MessageId}' of type {MessageType} in {TimeToHandle}ms."`, with the context properties:
`Status`: Either `Succeeded` or `Failed` depending on what the dispatcher pipeline returns.
`MessageId`: The JustSaying.Models.Message.Id property
`MessageType`: Should be the same as the `Subject` property on publish for most use cases.
`TimeToHandle`: Number of milliseconds that were spent in the dispatcher pipeline.

I think this message is pretty good as it is, although it also has a context/category of `JustSaying`. I think we should make it's context `JustSaying.Handle` or `JustSaying.Receive` so that we can filter it out, similarly to publishes.

### Misc

[...]

PR #897, opened 2021-04-14
Unify subscribe logs

gkinsman:
This PR completes #867 by adding log context to match the publish logs.

The list of properties for this log are:
Status: Succeeded or Failed
`MessageId`: MessageId guid
`MessageType`: Type of the message
`TimeToHandle`: ms taken to handle the message
`SourceType`: Always "Queue", here to be consistent with publish logs (we don't currently support subscribe from topic)
`MessageSource`: The name of the queue the message came from

To compare with publish logs:

`MessageId`
`MessageType`
`DestinationType`: Queue or Topic
`PublishDestination`: the topic or queue arn
`AwsRequestId`

### DIFF
--- a/src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs
+++ b/src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs
@@ -175,10 +175,10 @@ namespace JustSaying.AwsTools.MessageHandling.Dispatch
 
             using (_messagingMonitor.MeasureDispatch())
             {
+                var context = new HandleMessageContext(message, messageType, queueName);
                 bool dispatchSuccessful = false;
                 try
                 {
-                    var context = new HandleMessageContext(message, messageType, queueName);
                     dispatchSuccessful = await middleware.RunAsync(context, null, cancellationToken)
                         .ConfigureAwait(false);
                 }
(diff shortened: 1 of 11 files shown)


src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs
```diff
@@ -175,34 +175,41 @@ private async Task<bool> CallMessageHandler(string queueName, Message message, C
...
                     {
-                        _logger.LogInformation(logMessage,
-                            "Succeeded",
-                            message.Id,
-                            messageType,
-                            watch.ElapsedMilliseconds);
-                    }
-                    else
+                        new KeyValuePair<string, object>("MessageSource", context.QueueName),
+                        new KeyValuePair<string, object>("SourceType", "Queue")
```
martincostello, line 192: Would `KeyValuePair.Create(..., ...)` work here and the co(contra?)-variance would make it compile with it wanting the second type parameter needing to be an object and not the actual value's type, then it's a little less verbose. No worries if not.

slang25: It wouldn't be able to pick a best inferred type from experiments without giving the array a type. We don't have `KeyValuePair.Create` unfortunately anyway as it's a netstandard2.1 thing

gkinsman: The [Serilog.AspNetCore docs](https://github.com/serilog/serilog-aspnetcore#pushing-properties-to-the-iloggert) show this pattern instead, which is a bit nicer:

```csharp
// Microsoft.Extensions.Logging ILogger<T>
// Yes, it's required to use a dictionary. See https://nblumhardt.com/2016/11/ilogger-beginscope/
using (logger.BeginScope(new Dictionary<string, object>
{
    ["UserId"] = "svrooij",
    ["OperationType"] = "update",
}))
{
   // UserId and OperationType are set for all logging events in these brackets
}
```

slang25 (approved):
Nice one @gkinsman

PR #907, opened 2021-04-26
remove some redundant state machines

SimonCropp:


martincostello (commented):
The code changes are fine for what's been done, but the question is whether we want to accept this change or not as it goes against the recommendation here: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

Which do we prefer, the performance or the easier debuggability?

Personally I prefer the latter (which is why the code is the way it is), but I'm happy to go with the majority opinion.

SimonCropp:
i have read that advice. the strange thing i find is that aspnetcore does the opposite of the advice. it returns task when possible.

SimonCropp:
just did a check asp net core contains ~2000 instances of Task being directly returned.

martincostello:
> just did a check asp net core contains ~2000 instances of Task being directly returned.

I believe ASP.NET Core does the opposite because of their performance targets:

> 💡NOTE: There are performance considerations when using an async state machine over directly returning the Task. It's always faster to directly return the Task since it does less work but you end up changing the behavior and potentially losing some of the benefits of the async state machine.

Where as the general advice is to err on the improved developer experience.

iancooper:
I'm probably an outlier but for my part: returning a Task indicates parallelism, returning async indicates concurrency using evented I/O. I'd use a bare Task to indicate that I was using a thread, which the Task is a wrapper over, and an async Task to indicate a pending callback. Sure, both represent promises, but they promise quite different things IMO.

SimonCropp:
seems this is controversial

PR #910, opened 2021-04-26
top level methods for the win

SimonCropp:


src/JustSaying.Tools/Program.cs
```diff
@@ -1,23 +1,15 @@
...
+}
 
-        private static async Task<bool> ProcessLine(string line)
-        {
-            return await CommandParser.ParseAndExecuteAsync(line).ConfigureAwait(false);
-        }
-    }
+static async Task<bool> ProcessLine(string line)
+{
+    return await CommandParser.ParseAndExecuteAsync(line).ConfigureAwait(false);
```
martincostello, line 14: Could just inline this in the `if` block?

```csharp
await CommandParser.ParseAndExecuteAsync(line).ConfigureAwait(false);
```

PR #930, opened 2021-08-03
Switch to MELT for testing logging

gkinsman:
Microsoft.Extensions.Logging.Testing aka MELT on nuget is a useful lib that captures logs and provides access to them in tests. It's also a lot more reliable to test than approvals as approvals rely on the order of output.

Other misc things:
- Fix some issues with property values being types and not strings
- Use FullName everywhere


martincostello (approved):

slang25 (dismissed):
Very nice

martincostello (approved):

martincostello:
If this works then just need to rebase #929 once this is merged.

martincostello (approved):

Issue #875, opened 2021-03-24
Add back some of the extensibility to JustSaying 7 APIs


slang25:
As part of the fluent configuration rework, some of the lesser known APIs were not fully implemented:
https://github.com/justeat/JustSaying/blob/main/src/JustSaying/Fluent/SqsConfigurationBuilder.cs#L149-L159

We have users that use:
- `MessageBackoffStrategy` (multiple usages)
- `FilterPolicy` (1 usage)
- `ErrorQueueRetentionPeriod` (very rare, but there is at least 1 usage)

We might want to resurface some of these for v7, and assess the others.

gkinsman:
MessageBackoffStrategy will at least be needed to ensure a smooth migration path from JS 6

PR #932, opened 2021-08-07
Fix messagebackoffstrategy

gkinsman:
This PR partially resolves https://github.com/justeat/JustSaying/issues/875 by moving much of the post-processing code from the `MessageDispatcher` to separate middlewares.

The default middleware chain for all handlers is now:
```csharp
"MiddlewareChain": [
  "MessageContextAccessorMiddleware",
  "ErrorHandlerMiddleware",
  "LoggingMiddleware",
  "StopwatchMiddleware",
  "SqsPostProcessorMiddleware",
  "HandlerInvocationMiddleware`1"
]
```

It is also now possible to modify the middleware configuration from the subscription builders directly, where before you had to go via the SqsReadConfiguration. 

Before:
```csharp
(TopicSubscriptionBuilder<SimpleMessage> c) => 
  c.WithReadConfiguration(rc =>
      rc.WithMiddlewareConfiguration(m =>
        m.UseExactlyOnce<SimpleMessage>("simple-message-lock")))))
```
After:
```csharp
(TopicSubscriptionBuilder<SimpleMessage> c) => 
    c.WithMiddlewareConfiguration(m =>
        m.UseExactlyOnce<SimpleMessage>("lock-simple-message")
         .UseDefaults<SimpleMessage>(handler.GetType()))))
```

The resulting middleware chain:
```csharp
        "MiddlewareChain": [
          "ExactlyOnceMiddleware`1",
          "MessageContextAccessorMiddleware",
          "ErrorHandlerMiddleware",
          "LoggingMiddleware",
          "StopwatchMiddleware",
          "SqsPostProcessorMiddleware",
          "HandlerInvocationMiddleware`1"
        ]
```

The `UseDefaults` method will add all of the above middlewares as a shortcut for having to add them all manually. Previously it wasn't possible to have a no-op pipeline as the `HandlerInvocationMiddleware` was always added, but now you can by not calling this.

#### A bit of debt
The need to include the handler type is to maintain the `IMessageMonitor.HandlerExecutionTime(_handlerType, context.MessageType, watch.Elapsed)` contract. Possibly this call could be moved to the handler invocation middleware where we have the handler type (so that we don't need to resolve the handler in each middleware). 

### DI Wireup
There has also been a change to how we resolve services from configuration (using `ServicesBuilder`) vs the provided `IServiceResolver`. Previously both the `IServiceResolver` and the `ServicesBuilder` were checked to see if the user has overridden various services such as `IMessageMonitor` via config instead of adding it to the container. `IMessageMonitor` is used all over the place, so this check had to be done in each location (it wasn't, which was causing bugs).

There is now an `IServiceResolver` over the `ServicesBuilder` (`ServiceBuilderServiceResolver`) that enables resolution from anywhere without needing to care where the component was registered. A `CompoundServiceResolver` provides a single service resolver for both the service builder resolver, and the user provided resolver (or the `DefaultServiceResolver` if none supplied).

This is particularly useful for middlewares, as previously we had to know about the ServicesBuilder in order to find the correct `IMessageMonitor`, for example. We can now wholly rely on a single `IServicesResolver`.

I'm keen for feedback on this design - it's a little ugly, but IMO superior to infecting the rest of the codebase with knowledge of the services builder.

TODO:
- [ ] Add XML docs to new public members
- [ ] Cleanup any dead code/things hanging around
- [ ] Add some more tests for middleware building


src/JustSaying/Messaging/Middleware/ErrorHandling/ErrorHandlerMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace JustSaying.Messaging.Middleware.ErrorHandling
+{
+    public static class ErrorHandlerMiddlewareBuilderExtensions
+    {
+        public static HandlerMiddlewareBuilder UseErrorHandler(this HandlerMiddlewareBuilder builder)
+        {
+            if (builder == null) throw new ArgumentNullException(nameof(builder));
+
+            IMessageMonitor monitor = builder.ServiceResolver.ResolveService<IMessageMonitor>();
```
gkinsman, line 12: We can't just resolve the `ErrorHandlerMiddleware` from the container, because if the `IMessageMonitor` was added via the ServicesBuilder, then resolving the middleware from the container will bypass the SB and resolve from the container, which is a no-op by default.

The downside is that middlewares that do this are instance-per-queue instead of shared.

src/JustSaying.Tools/Commands/MoveCommand.cs
```diff
@@ -1,93 +0,0 @@
...
-            var sourceQueue = new SqsQueueByName(config.RegionEndpoint, SourceQueueName, client, JustSayingConstants.DefaultHandlerRetryCount, loggerFactory);
-            var destinationQueue = new SqsQueueByName(config.RegionEndpoint, DestinationQueueName, client, JustSayingConstants.DefaultHandlerRetryCount, loggerFactory);
-
-            await EnsureQueueExistsAsync(sourceQueue).ConfigureAwait(false);
-            await EnsureQueueExistsAsync(destinationQueue).ConfigureAwait(false);
-
-            var messages = PopMessagesFromSourceQueue(sourceQueue);
-            var receiptHandles = messages.ToDictionary(m => m.MessageId, m => m.ReceiptHandle);
-
-            var sendResponse = destinationQueue.Client.SendMessageBatch(new SendMessageBatchRequest
```
gkinsman, line 46: This Client property is private now as we no longer expose the IAmazonSqs, so I've removed this command as I don't think its used very much. We have other ways of handling retries.

src/JustSaying/Messaging/Channels/Receive/MessageReceiveBuffer.cs
```diff
@@ -101,12 +102,16 @@ public async Task RunAsync(CancellationToken stoppingToken)
...
+                        _logger.LogTrace("Downloaded {MessageCount} messages from queue {QueueName}.", messages.Count, _sqsQueueReader.QueueName);
+                    }
+
                     foreach (Message message in messages)
                     {
                         IQueueMessageContext messageContext = _sqsQueueReader.ToMessageContext(message);
 
-                        // Complete all messages in the batch, rather than observing the CancellationToken to stop
-                        await writer.WriteAsync(messageContext, CancellationToken.None).ConfigureAwait(false);
+                        await writer.WriteAsync(messageContext, stoppingToken).ConfigureAwait(false);
```
gkinsman, line 114: Passing CancellationToken.None here was causing a deadlock when there was no consumer attached, so this would wait forever after shutdown as the queue wasnt draining.

slang25: Damn! Nice find

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
         /// </summary>
-        IAmazonSQS Client { get; }
+        /// <param name="queueUrl">The URL of the queue to delete a message from</param>
+        /// <param name="receiptHandle">The receipt handle of the message to delete</param>
+        /// <param name="cancellationToken">A cancellation token to cancel this operation.</param>
+        /// <returns>A <see cref="Task"/> will complete when the message has been deleted, or the task has faulted.</returns>
+        public Task DeleteMessageAsync(string queueUrl, string receiptHandle, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Tags a queue with one or more key=value pairs.
```
martincostello, line 46:
```suggestion
        /// Tags a queue with one or more key-value pairs.
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
+        /// </summary>
+        /// <param name="queueUrl">The URL of the queue to delete a message from</param>
```
martincostello, line 48:
```suggestion
        /// <param name="queueUrl">The URL of the queue to tag a message from.</param>
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
+        /// <param name="tags">A <see cref="Dictionary{string, string}"/> of tags to tag this queue with.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel this operation.</param>
+        /// <returns>A <see cref="Task"/> will complete when the queue has been tagged, or the task has faulted.</returns>
+        public Task TagQueueAsync(string queueUrl, Dictionary<string, string> tags, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Downloads messages from a queue.
+        /// </summary>
+        /// <param name="queueUrl">The URL of the queue to delete a message from</param>
```
martincostello, line 57:
```suggestion
        /// <param name="queueUrl">The URL of the queue to download a message from.</param>
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
         string Arn { get; }
 
         /// <summary>
-        /// Gets the SQS queue client.
+        /// Deletes a message from a queue
```
martincostello, line 37:
```suggestion
        /// Deletes a message from a queue.
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
         /// </summary>
-        IAmazonSQS Client { get; }
+        /// <param name="queueUrl">The URL of the queue to delete a message from</param>
```
martincostello, line 39:
```suggestion
        /// <param name="queueUrl">The URL of the queue to delete a message from.</param>
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
+        /// <param name="receiptHandle">The receipt handle of the message to delete</param>
```
martincostello, line 40:
```suggestion
        /// <param name="receiptHandle">The receipt handle of the message to delete.</param>
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
+        /// <param name="cancellationToken">A cancellation token to cancel this operation.</param>
+        /// <returns>A <see cref="Task"/> will complete when the message has been deleted, or the task has faulted.</returns>
+        public Task DeleteMessageAsync(string queueUrl, string receiptHandle, CancellationToken cancellationToken);
```
martincostello, line 43:
```suggestion
        Task DeleteMessageAsync(string queueUrl, string receiptHandle, CancellationToken cancellationToken);
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
+
+        /// <summary>
+        /// Tags a queue with one or more key=value pairs.
+        /// </summary>
+        /// <param name="queueUrl">The URL of the queue to delete a message from</param>
+        /// <param name="tags">A <see cref="Dictionary{string, string}"/> of tags to tag this queue with.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel this operation.</param>
+        /// <returns>A <see cref="Task"/> will complete when the queue has been tagged, or the task has faulted.</returns>
+        public Task TagQueueAsync(string queueUrl, Dictionary<string, string> tags, CancellationToken cancellationToken);
```
martincostello, line 52:
```suggestion
        Task TagQueueAsync(string queueUrl, Dictionary<string, string> tags, CancellationToken cancellationToken);
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
+        /// <summary>
+        /// Downloads messages from a queue.
+        /// </summary>
+        /// <param name="queueUrl">The URL of the queue to delete a message from</param>
+        /// <param name="maxNumOfMessages">The maximum number of messages to try to download. Values larger than 10 will default to 10. </param>
+        /// <param name="secondsWaitTime">The number of seconds to wait for messages to be available before returning.</param>
+        /// <param name="attributesToLoad">A list of attributes to retrieve for the downloaded messages.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel this operation.</param>
+        /// <returns>A <see cref="Task"/> will complete when messages have been received, or the task has faulted.</returns>
+        public Task<IList<Message>> ReceiveMessagesAsync(
```
martincostello, line 63:
```suggestion
        Task<IList<Message>> ReceiveMessagesAsync(
```

src/JustSaying/AwsTools/MessageHandling/ISqsQueue.cs
```diff
@@ -30,8 +34,52 @@ public interface ISqsQueue : IInterrogable
...
+
+        /// <summary>
+        /// Updates the visibility timeout of a message to prevent it from being re-handled until a future time.
+        /// </summary>
+        /// <param name="queueUrl">The URL of the queue to delete a message from</param>
+        /// <param name="receiptHandle">The receipt handle of the message that will be updated</param>
+        /// <param name="visibilityTimeoutInSeconds">The number of seconds until this message will be visible to other consumers</param>
+        /// <param name="cancellationToken">A cancellation token to cancel this operation.</param>
+        /// <returns></returns>
+        public Task ChangeMessageVisibilityAsync(
```
martincostello, line 78:
```suggestion
        Task ChangeMessageVisibilityAsync(
```

src/JustSaying/AwsTools/MessageHandling/SqsQueueReader.cs
```diff
@@ -31,49 +31,37 @@ internal IQueueMessageContext ToMessageContext(Message message)
         internal async Task<IList<Message>> GetMessagesAsync(
             int maximumCount,
             TimeSpan waitTime,
-            IEnumerable<string> requestMessageAttributeNames,
+            IList<string> requestMessageAttributeNames,
```
martincostello, line 34: Could just use `ICollection<T>`?

src/JustSaying/Extensions/AmazonSqsClientExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            var result = await client.ReceiveMessageAsync(new ReceiveMessageRequest(queueUrl)
+                {
+                    AttributeNames = attributesToLoad.ToList(),
+                    WaitTimeSeconds = secondsWaitTime,
+                    MaxNumberOfMessages = maxNumOfMessages
+                },
+                cancellationToken).ConfigureAwait(false);
+
+            return result?.Messages ?? new List<Message>();
```
martincostello, line 32: Use `Array.Empty<T>`?

src/JustSaying/Extensions/AmazonSqsClientExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using Amazon.SQS;
+using Amazon.SQS.Model;
+
+namespace JustSaying.Extensions
+{
+    public static class AmazonSqsClientExtensions
+    {
+        public static async Task TagQueueAsync(this IAmazonSQS client, string queueUrl, Dictionary<string, string> tags, CancellationToken cancellationToken)
+        {
+            await client.TagQueueAsync(new TagQueueRequest()
```
martincostello, line 14: As public should null-check the client in both these methods. Does this extensions type need to be public though?

martincostello (commented):
This looks good.

I wonder if with the extra CancellationToken usages if anywhere needs changing to more gracefully needs to handle if it fires and causes a cancellation mid operation? (Say between two API calls related to a single message batch, like a receive and then a delete)

src/JustSaying/Extensions/AmazonSqsClientExtensions.cs
gkinsman: No it probably doesn't need to be - will make this class internal as its just to make it easier to call these methods without having to know about or construct AWS request types.

gkinsman:
> This looks good.
> 
> I wonder if with the extra CancellationToken usages if anywhere needs changing to more gracefully needs to handle if it fires and causes a cancellation mid operation? (Say between two API calls related to a single message batch, like a receive and then a delete)

I think it's an interesting question, although this is already possible today through app/machine crashes between the message being handled and the dispatcher deleting the message. Threading the cancellation token perhaps makes it more likely to happen on shutdown, but there should be no change to behaviour: if the bus is stopped after a message is handled but before it's deleted, it'll be handled again by another instance.

src/JustSaying/AwsTools/MessageHandling/SqsQueueReader.cs
gkinsman: Done!

tests/JustSaying.UnitTests/Messaging/Channels/Fakes/InMemoryServiceResolver.cs
```diff
@@ -0,0 +1,60 @@
...
+using JustSaying.Messaging.Middleware.MessageContext;
+using JustSaying.Messaging.Middleware.PostProcessing;
+using JustSaying.Messaging.Monitoring;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Xunit.Abstractions;
+
+namespace JustSaying.UnitTests.Messaging.Channels.Fakes
+{
+    public class InMemoryServiceResolver : IServiceResolver, IHandlerResolver
```
martincostello, line 16: Doesn't this just duplicate the built-in implementation we have that wraps `IServiceResolver` already?

src/JustSaying/AwsTools/MessageHandling/SqsQueueReader.cs
```diff
@@ -31,49 +31,36 @@ internal IQueueMessageContext ToMessageContext(Message message)
...
-            var visibilityRequest = new ChangeMessageVisibilityRequest
-            {
-                QueueUrl = _sqsQueue.Uri.ToString(),
-                ReceiptHandle = receiptHandle,
-                VisibilityTimeout = (int)timeout.TotalSeconds,
-            };
-
-            await _sqsQueue.Client.ChangeMessageVisibilityAsync(visibilityRequest, cancellationToken).ConfigureAwait(false);
+            await _sqsQueue.ChangeMessageVisibilityAsync(
+                _sqsQueue.Uri.ToString(),
```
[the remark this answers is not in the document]
gkinsman: I think we could go either way as we don't need the url encoding of AbsoluteUri, and the fact that AbsoluteUri asserts this is an absolute Uri is kinda superfluous. ToString seems less surprising?

src/JustSaying/Fluent/ServiceResolver/ServiceBuilderServiceResolver.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public ServiceBuilderServiceResolver(ServicesBuilder builder)
+        {
+            _builder = builder;
+
+        }
+
+        public T ResolveService<T>() where T : class
+        {
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
gkinsman: It was intentional as this service resolver doesn't support optional resolutions, but instead of throwing an `InvalidOperationException` here it's probably cleaner to just implement ResolveService like this. Thoughts?
```csharp
public T ResolveService<T>() where T : class
{
    return ResolveOptionalService<T>() ??
        throw new InvalidOperationException(
            $"Service type {typeof(T).FullName} isn't available from this service resolver.");
}
```

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Threading.Tasks;
+using Amazon.Runtime;
+using Amazon.SQS;
+using JustSaying.Messaging.MessageProcessingStrategies;
+using JustSaying.Messaging.Monitoring;
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.Messaging.Middleware.Backoff
+{
+    public class BackoffMiddleware : MiddlewareBase<HandleMessageContext, bool>
```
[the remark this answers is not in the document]
gkinsman: Would `BackoffPolicyMiddleware` be a bit gentler? 🙂

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+        public BackoffMiddleware(IMessageBackoffStrategy backoffStrategy, ILoggerFactory loggerFactory, IMessageMonitor monitor)
+        {
+            _backoffStrategy = backoffStrategy;
+            _monitor = monitor;
+            _logger = loggerFactory.CreateLogger<BackoffMiddleware>();
+        }
+
+        protected override async Task<bool> RunInnerAsync(HandleMessageContext context, Func<CancellationToken, Task<bool>> func, CancellationToken stoppingToken)
+        {
+            bool handlingSucceeded = false;
```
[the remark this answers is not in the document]
gkinsman: I don't think we need to null check these in middlewares, as they're passed to it from the framework - they'll never be null.

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace JustSaying.Messaging.Middleware.Backoff
+{
+    public static class BackoffMiddlewareBuilderExtensions
+    {
+        public static HandlerMiddlewareBuilder UseBackoff(this HandlerMiddlewareBuilder builder)
+        {
+            if (builder == null) throw new ArgumentNullException(nameof(builder));
+
+            var backoffStrategy = builder.ServiceResolver.ResolveOptionalService<IMessageBackoffStrategy>();
+            if (backoffStrategy == null) return builder;
```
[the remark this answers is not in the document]
gkinsman: Yeah it's a fair question. We could instead do the check in `UseDefaults` to determine whether to call this, and then throw if we can't find an instance of `IMessageBackoffStrategy`? I tended towards the way it is but don't have strong feelings.

```csharp
builder.UseMessageContextAccessor();
if (builder.ServiceResolver.ResolveOptionalService<IMessageBackoffStrategy>() != null)
{
    builder.UseBackoff();
}
builder.UseErrorHandler();
builder.Use<LoggingMiddleware>();
builder.UseStopwatch(handlerType);
builder.Use<SqsPostProcessorMiddleware>();
builder.UseHandler<TMessage>();
```

src/JustSaying/Messaging/Middleware/ErrorHandling/ErrorHandlerMiddleware.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class ErrorHandlerMiddleware : MiddlewareBase<HandleMessageContext, bool>
+    {
+        private readonly IMessageMonitor _monitor;
+
+        public ErrorHandlerMiddleware(IMessageMonitor monitor)
+        {
+            _monitor = monitor;
+        }
+
+        protected override async Task<bool> RunInnerAsync(HandleMessageContext context, Func<CancellationToken, Task<bool>> func, CancellationToken stoppingToken)
```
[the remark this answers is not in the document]
gkinsman: Yeah again I don't think we need them for these as the fx passes them in.

src/JustSaying/Fluent/ISubscriptionBuilder`1.cs
```diff
@@ -20,24 +20,36 @@ public interface ISubscriptionBuilder<out T>
         /// Configures the middleware pipeline for this subscription.
         /// Any middleware configured here will be wrapped around a handler and metrics middleware.
         /// </summary>
-        /// <param name="middlewareConfiguration"></param>
+        /// <param name="middlewareConfiguration">A configuration action that provides an API
+        /// to override the default middleware behaviour. By default, <see cref="HandlerMiddlewareBuilderExtensions.UseDefaults{TMessage}"/> applies a set of
+        /// default middlewares to add metrics, error handling, completion handling, context setting, and logging.
+        /// Below is an example of how to add your own middlewares around the defaults.</param>
```
martincostello, line 26: Not sure if "Below..." would make sense when viewed in an Intellisense context.

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddleware.cs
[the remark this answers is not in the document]
martincostello: Genuinely not sure, it just sounded funny in my head.

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddleware.cs
```diff
@@ -11,17 +11,27 @@
...
         private readonly IMessageBackoffStrategy _backoffStrategy;
         private readonly ILogger<BackoffMiddleware> _logger;
         private readonly IMessageMonitor _monitor;
 
+        /// <summary>
+        /// Constructs a <see cref="BackoffMiddleware"/> with a given backoff strategy and logger/monitor.
+        /// </summary>
+        /// <param name="backoffStrategy">A <see cref="IMessageBackoffStrategy"/> to use to determine how long to delay message redelivery when a handler returns false or throws.</param>
+        /// <param name="loggerFactory"></param>
+        /// <param name="monitor"></param>
```
martincostello:
```suggestion
        /// <param name="loggerFactory">The logger factory to use.</param>
        /// <param name="monitor">The message monitor to use.</param>
```

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddleware.cs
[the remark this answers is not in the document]
martincostello: Not a hill I'm going to die on, just suggested it as analyzers will flag for `public` and `protected` as potentially a derived type could use it incorrectly and pass a null (if it's a public type).

src/JustSaying/Messaging/Middleware/MessageContext/MessageContextAccessorMiddleware.cs
```diff
@@ -0,0 +1,31 @@
...
+        public MessageContextAccessorMiddleware(IMessageContextAccessor messageContextAccessor)
+        {
+            _messageContextAccessor = messageContextAccessor;
+        }
+
+        protected override async Task<bool> RunInnerAsync(HandleMessageContext context, Func<CancellationToken, Task<bool>> func, CancellationToken stoppingToken)
+        {
+            try
+            {
+                _messageContextAccessor.MessageContext = new MessageHandling.MessageContext(context.RawMessage, context.QueueUri, context.MessageAttributes);
```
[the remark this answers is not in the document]
gkinsman: sure, just curious if its for any particular reason or just a consistency thing? 👍

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddlewareBuilderExtensions.cs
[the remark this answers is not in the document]
martincostello: I don't have strong feelings on a solution, it just seems that if something was misconfigured, if it didn't throw or something, then it would be less surprising/fail-fast than if it just silently looked like it had done something.

src/JustSaying/Messaging/Middleware/MessageContext/MessageContextAccessorMiddleware.cs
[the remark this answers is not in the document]
martincostello: Just that if the constructor threw, then the assignment would fail, so setting it back to null would be redundant as it would already be null.

src/JustSaying/Fluent/ISubscriptionBuilder`1.cs
gkinsman: Yeah fair - at least on rider it's not particularly useful. I think the example is clear enough, no need to restate it.

image

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddlewareBuilderExtensions.cs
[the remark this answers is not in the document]
gkinsman: That's a fair point I think - the failure behaviour shouldn't really be nothing. I thought about a log but it would log by default if no back off was registered, which is the default.

I also experimented with a `UseIf` function that conditionally added it to the pipe, but it stuffs with the ordering and makes things a bit too unpredictable I think, so pulled it out.

Perhaps another option might be to require the user to add the backoff themselves? Instead of auto adding it if we find an IMessageBackoffStrategy, we could flip it so the user has to do `builder.UseBackoff(IMessageBackoffStrategy strategy)`. Feels a bit service locatorey to me though as you'd need to pull it out or create it separately beforehand.

tests/JustSaying.UnitTests/Messaging/Channels/Fakes/FakeMessageDeleter.cs
```diff
@@ -0,0 +1,14 @@
+using System.Threading;
+using System.Threading.Tasks;
+using JustSaying.Messaging.Channels.Context;
+
+namespace JustSaying.UnitTests.Messaging.Channels.Fakes
+{
+    public class FakeMessageDeleter : IMessageDeleter
```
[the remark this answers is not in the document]
gkinsman: I admittedly tend towards fakes instead of mocks in general, as fakes only need to be created once. In this case we don't really care about tracking actions (we track them on the ISqsQueue instead), so it's simple enough that a mock would require no setup true. Happy to change it, I don't mind too much 🙂

tests/JustSaying.UnitTests/Messaging/Channels/Fakes/InMemoryServiceResolver.cs
gkinsman: Assuming you're referring to `ServiceProviderResolver`, it takes an already built `IServiceProvider` which would mean tests would need to create a service collection, register stuff, then create a ServiceProviderResolver from it. Totally doable, but this one streamlines it a bit by providing a configuration action for overrides, and registering some defaults. The ServiceProviderResolver ctor is also internal so can't be called from tests unless we change it.

We could create a static class with defaults so you could do something like 
```
var sp = new ServiceCollection().AddTestDefaults().BuildServiceProvider()
var resolver= new ServiceProviderResolver(sp)
```
which is pretty much what this does.

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddleware.cs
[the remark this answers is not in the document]
gkinsman: Hmm I wonder if we should seal these middlewares instead. I wouldn't think deriving from them would ever make sense - the extensibility point is to add a new middleware.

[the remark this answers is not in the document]
gkinsman: I've opted to seal them to avoid the need for this. Middlewares shouldn't be inherited 👍

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddlewareBuilderExtensions.cs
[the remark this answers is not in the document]
gkinsman: So I've been going around a bit on this one, given its the last bit of the puzzle 😅. I'm leaning a bit towards removing `IMessageBackoffStrategy` from DI and instead requiring the user to pass it in to the middleware, like so:

```csharp
builder.UseBackoff(IMessageBackoffStrategy strategy)
```

And I think this has a few benefits:
- It would be possible to use a different backoff strategy for different subscriptions, whereas at the moment there's only one in DI.
- It'd mean we can leave the call to `UseBackoff` out of the defaults, and users would opt-in as above.
- We can check for a null IMBS in UseBackoff

thoughts?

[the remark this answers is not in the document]
martincostello: Making this opt-in seems fair. We could always have it as a default in JustSayingStack.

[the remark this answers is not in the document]
slang25: Yeah I like this suggestion of opting in at the middleware builder layer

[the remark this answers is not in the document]
gkinsman: I've PR'd this approach here to make the change clear: https://github.com/gkinsman/JustSaying/pull/153.

There is a downside, which is that we'd now have to always receive the `ApproximateReceiveCount` attribute as before we only fetched it if an IMBS was registered.

Thoughts @slang25 @martincostello?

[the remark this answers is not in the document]
martincostello: Just thinking out loud: could there be an internal config-y/options-y type that could be queried to control whether or not to fetch it, then adding back-off would add that in and make it get configured to be fetched at the appropriate place?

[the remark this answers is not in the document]
gkinsman: Ah that's an interesting idea - perhaps a set of receive attributes that is configurable. That would open the doors to future middlewares that wanted to use attributes other than just approx receive count.

I think if we configure it on the `ISqsQueue` then it could be used by the receive buffer, and set by the backoff middleware - there might be some trickery but will see if it falls out.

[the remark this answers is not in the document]
gkinsman: How would we feel about merging this as-is and following up with the optimisation to prevent always loading the approx count? Cautious that this is already new-concept heavy 🙂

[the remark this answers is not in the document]
martincostello: Works for me.

martincostello (dismissed):

src/JustSaying/Messaging/Middleware/Backoff/BackoffMiddlewareBuilderExtensions.cs
```diff
@@ -13,14 +13,13 @@ public static class BackoffMiddlewareBuilderExtensions
         /// a <see cref="BackoffMiddleware"/> and add it to the pipeline.
         /// </summary>
         /// <param name="builder">The <see cref="HandlerMiddlewareBuilder"/> to add the middleware to.</param>
+        /// <param name="backoffStrategy">The <see cref="IMessageBackoffStrategy"/> to use to determine message visibility timeouts.</param>
         /// <returns>The current <see cref="HandlerMiddlewareBuilder"/>.</returns>
         /// <exception cref="ArgumentNullException">When the <see cref="HandlerMiddlewareBuilder"/> is null.</exception>
-        public static HandlerMiddlewareBuilder UseBackoff(this HandlerMiddlewareBuilder builder)
+        public static HandlerMiddlewareBuilder UseBackoff(this HandlerMiddlewareBuilder builder, IMessageBackoffStrategy backoffStrategy)
```
martincostello, line 19: Nit: null check the back-off strategy.

gkinsman: Ah thanks, missed this! 👍

martincostello: Did you forget to push the fix?

gkinsman: 🤦‍♂️ I pushed it to my other branch instead of here hah. Done now, sorry about that!

martincostello (approved):

PR #935, opened 2021-09-04
Update getting-started.md

jasonsultana:
Specify that the extension method is defined in a separate package to the main JustSaying nuget package, which might not be obvious at first glance.

I thought it was also a good idea to include that the package is in pre-release, since prerelease packages aren't shown in the nuget package manager window by default.


slang25:
Thanks @jasonsultana for taking the time to fix up our getting started section. We're close to getting the next release out, however until we do I can see this saving many people some time.

PR #938, opened 2021-09-28
Try localstack 0.12.18

slang25:
We've pinned localstack to 0.12.15 as there were issues in subsequent releases causing our functional tests to fail.
Trying 0.12.18 as there are fixes to SQS which may be related to what we were seeing.


gkinsman:
darn 😢

slang25:
💩

PR #947, opened 2021-11-08
.NET 6

slang25:


tests/JustSaying.IntegrationTests/Fluent/AwsTools/WhenUsingABasicThrottle.cs
```diff
@@ -12,131 +8,131 @@
...
-        [InlineData(1000)]
-        public async Task Messages_Are_Throttled_But_Still_Delivered(int throttleMessageCount)
-        {
-            // Arrange
-            ILoggerFactory loggerFactory = OutputHelper.ToLoggerFactory();
-            IAwsClientFactory clientFactory = CreateClientFactory();
+    [AwsTheory]
+    [InlineData(100)]
+    [InlineData(1000)]
+#pragma warning disable xUnit1008
```
martincostello, line 28: I wonder if there's a bug here worth logging with xunit that derived classes aren't considered as part of the rule?

samples/src/JustSaying.Sample.Restaurant.OrderingApi/Program.cs
```diff
@@ -1,47 +1,120 @@
...
         {
-            return Host.CreateDefaultBuilder(args)
-                .UseSerilog()
-                .ConfigureWebHostDefaults((builder) => builder.UseStartup<Startup>());
-        }
-    }
+            app.Logger.LogInformation("Order received for {description}", order.Description);
+
+            // Save order to database generating OrderId
+            var orderId = new Random().Next(1, 100);
```
martincostello, line 98:
```suggestion
            var orderId = Random.Shared.Next(1, 100);
```

.editorconfig
```diff
@@ -32,6 +32,7 @@ csharp_new_line_before_members_in_object_initializers=false
 csharp_new_line_before_open_brace=all
 csharp_preferred_modifier_order=public, private, protected, internal, new, abstract, virtual, sealed, override, static, readonly, extern, unsafe, volatile, async:suggestion
 csharp_space_after_keywords_in_control_flow_statements=true
+csharp_style_namespace_declarations = file_scoped
```
martincostello, line 35: TIL, you need to add `:warning` for `dotnet format` to enforce/apply it.

src/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
```diff
@@ -8,10 +8,10 @@
   <ItemGroup>
     <ProjectReference Include="..\JustSaying\JustSaying.csproj" />
   </ItemGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="1.1.0" />
-  </ItemGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.0.0" />
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="6.0.0" />
```
martincostello, line 12: Do we want to force this to require .NET 6?

src/JustSaying.Extensions.DependencyInjection.StructureMap/JustSaying.Extensions.DependencyInjection.StructureMap.csproj
```diff
@@ -12,6 +12,9 @@
     <ProjectReference Include="..\JustSaying\JustSaying.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="structuremap" Version="4.6.0" />
+    <PackageReference Include="StructureMap" Version="4.6.0" />
   </ItemGroup>
+  <ItemGroup>
+      <Using Remove="System.Net.Http" />
```
martincostello, line 18: Indenting sadness 😄

src/JustSaying/JustSaying.csproj
```diff
@@ -13,10 +13,13 @@
   <ItemGroup>
     <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.0" />
     <PackageReference Include="AWSSDK.SQS" Version="3.7.0" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
```
martincostello, line 17: Same here with forcing use of 6 for JustSaying 7 - feels it would make migration harder.

slang25: Yeah it would force package upgrades on people.

We used to try hard to avoid taking new versions of `Microsoft.Extensions.Logging.Abstractions` because `net461` folk would get a netstandard package explosion, today I think we can accept that particular issue.

I guess it all comes down to preference, do we want evergreen dependencies (last major version), or do we want to target the lowest compatible version of each package. I like the evergreen approach as there is no doubt when looking at the package that it will work well with the latest major versions.

I'm finding my attitude shifting over the last fews years as .NET has matured. What's your thoughts?

martincostello: I think right now to get this shipped and make internal adoption more likely to be successful without additional engineering effort, we could multi-target to `netstandard2.0`, with additional TFMs any of for `netcoreapp3.1`, `net5.0` and/or `net6.0`.

Each TFM could set its own baseline (2.0.0, 3.1.0, 5.0.0, 6.0.0) etc., then in a post-7.0.0 release we could drop 3.1 and 5.0.

I think if JustSaying was internal-only only supporting `net6.0`+ would be a good carrot for migrating from 3.1 and 5.0 sooner rather than later, but I feel with the number of people who've asked about when 7 will be released from outsite JET, it might be disappointing/rug-pulling to last-minute also require .NET 6.

slang25: As it stands this still supports the TFMs you are talking about, the packages that are updated target:
- .NETFramework 4.6.1
- .NETStandard 2.0
- .NETStandard 2.1
- .NET 6.0

I wasn't planning on updating the JustSaying package itself to additionally target .NET 6, as the benefits would be almost insignificant (at least as I understand it)

martincostello: Right OK.

My _understanding_ though (not sure where I've seen it written down and/or if it's correct) is that you should stick with packages that match your TFM (i.e. a `net5.0` app wouldn't reference `6.0.x` packages). If that's the case, even if we did go with "latest" for each one, we'd still have different "channels" of packages referenced in the package references.

My guess would be that it's so if there's anything that needs the new runtime that a package is relying on internally in its own code doesn't run into issues.

slang25: I remember there being some being the same concerns amongst the community about going to higher package versions that the current targeting runtime, and there being clarification on Twitter that it was fine, and that's why they multi-target these packages (can I find that tweet? Nope 😆 )

I think there are at least 5 approaches:
- Update the SDK don't change target TFMs or packages (maybe use C# 10 "syntax-ony" features)
- Add the .NET 6 TFM, keep the packages on the lowest compatible versions (well, for technical reasons it should be at least netstandard2.0 for .NET Core 2 and up)
- Keep the TFMs as they are and upgrade the packages (we are here at the moment)
- Add the .NET 6 TFM and update the packages
- Add the .NET 6 TFM and conditionally reference packages that align with that "channel"

That last option I dislike, but I don't really have a good reason for it.

martincostello: I think my _personal_ order of preference would be: 5, 1, 4, 2, 3

Maybe regardless of the preferred option, maybe we should keep this PR as the C# 10 PR and back out the package updates to the library (not the tests, benchmarks etc.), then we do a follow up to do whatever version changes we want? Then they'd be a bit cleaner to track back to the PR/diff at a later date?

slang25: I like that plan 👍

martincostello (approved):
Swish!

PR #958, opened 2021-11-10
Register Middlewares as Transient to avoid clobber

gkinsman:
I noticed this issue due to a failing test in JustSayingStack.

Most of our tests only establish a single subscription, which works fine. If there are multiple subscriptions however, then the second will resolve the same instance of `LoggingMiddleware` or `SqsPostProcessorMiddleware` (the only middlewares that are resolved from DI). As those instances were already used to build the previous pipeline, they are already part of an existing pipeline, and therefore have `_next` values that link them to the next step. When the second builder comes along and resolves the same middleware, it redirects the `_next` to the second subscription. The end result is that the same handler is invoked for all subscriptions. Bad!

This PR does the following to fix this:
- Adds a test to ensure this doesn't happen 
- Changes the middleware registrations to be transient
- Emits an InvalidOperationException when constructing middlewares if the `_next` property has already been populated, which indicates that the middleware has been incorrectly registered and the single instance is already part of an existing pipeline. If there's only one subscription this won't throw, so it still works for apps that only have a single sub and single instance middlewares.

### DIFF
--- a/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
+++ b/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
@@ -42,7 +42,13 @@ public sealed class HandlerMiddlewareBuilder
     public HandlerMiddlewareBuilder Use<TMiddleware>() where TMiddleware : MiddlewareBase<HandleMessageContext, bool>
     {
         var newMiddleware = ServiceResolver.ResolveService<TMiddleware>();
-        if (newMiddleware.HasNext) throw new InvalidOperationException("Middlewares must be registered as Transient.");
+        if (newMiddleware.HasNext)
+        {
+            throw new InvalidOperationException(
+                @"Middlewares must be registered into your DI container such that each resolution creates a new instance.
+For StructureMap use Transient(), and for Microsoft.Extensions.DependencyInjection, use AddTransient().
+Please check the documentation for your container for more details.");
+        }
 
         _middlewares.Add(() => newMiddleware);
         return this;
--- a/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
+++ b/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
@@ -38,10 +38,11 @@ public sealed class HandlerMiddlewareBuilder
     /// </summary>
     /// <typeparam name="TMiddleware">The type of the middleware to add.</typeparam>
     /// <returns>The current HandlerMiddlewareBuilder.</returns>
+    /// <exception cref="InvalidOperationException">When the middleware is not registered as Transient, an exception will be thrown if the resolved middleware is already part of a pipeline.</exception>
     public HandlerMiddlewareBuilder Use<TMiddleware>() where TMiddleware : MiddlewareBase<HandleMessageContext, bool>
     {
         var newMiddleware = ServiceResolver.ResolveService<TMiddleware>();
-        if (newMiddleware.HasNext) throw new InvalidOperationException("Middlewares must be registered as Transient");
+        if (newMiddleware.HasNext) throw new InvalidOperationException("Middlewares must be registered as Transient.");
 
         _middlewares.Add(() => newMiddleware);
         return this;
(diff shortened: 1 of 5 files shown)


src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
```diff
@@ -38,9 +38,13 @@ public HandlerMiddlewareBuilder(IHandlerResolver handlerResolver, IServiceResolv
     /// </summary>
     /// <typeparam name="TMiddleware">The type of the middleware to add.</typeparam>
     /// <returns>The current HandlerMiddlewareBuilder.</returns>
+    /// <exception cref="InvalidOperationException">When the middleware is not registered as Transient, an exception will be thrown if the resolved middleware is already part of a pipeline.</exception>
     public HandlerMiddlewareBuilder Use<TMiddleware>() where TMiddleware : MiddlewareBase<HandleMessageContext, bool>
     {
-        _middlewares.Add(() => ServiceResolver.ResolveService<TMiddleware>());
+        var newMiddleware = ServiceResolver.ResolveService<TMiddleware>();
+        if (newMiddleware.HasNext) throw new InvalidOperationException("Middlewares must be registered as Transient.");
```
martincostello, line 45: Is calling this "Transient" maybe a leaking implementation detail, or is it the right catch-all term?

src/JustSaying/Messaging/Middleware/MiddlewareBase`2.cs
```diff
@@ -10,6 +10,8 @@ public abstract class MiddlewareBase<TContext, TOut>
         return this;
     }
 
+    public bool HasNext => _next != null;
```
martincostello, line 13: I wonder if this should just be internal?

tests/JustSaying.IntegrationTests/Fluent/Subscribing/WhenSubscribingToMultipleTopics.cs
```diff
@@ -0,0 +1,48 @@
+using System.Threading.Tasks;
+using JustSaying.Messaging.MessageHandling;
+using JustSaying.TestingFramework;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+using Xunit.Abstractions;
+
+namespace JustSaying.IntegrationTests.Fluent.Subscribing
```
martincostello, line 8: As Stu did the big PR to add it yesterday, should use file-scoped namespace.

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
gkinsman: Yeah it's a good question - the point here is that we're indicating to users that they've registered a custom middleware incorrectly. Autofac uses `InstancePerDependency`, while StructureMap/MSDI uses `Transient`, so I thought the latter might be more widely used?

src/JustSaying/Messaging/Middleware/MiddlewareBase`2.cs
gkinsman: Ah yeah this is fair, good point.

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
martincostello: Maybe we could be a little bit more verbose in the exception message? I'm not against the way it is, just wondering if there's a different wording we could use to convey the idea and it still fitting other DI containers' concepts/terminology.

gkinsman: How about:


Middlewares must be registered into your DI container so that each resolution creates a new instance. For StructureMap use `Transient()`, and for Microsoft.Extensions.DependencyInjection, use `AddTransient`. Please check the documentation for your container for more details.

martincostello: I like it.

martincostello (approved):

Issue #1004, opened 2022-04-28
Support late binding of topic names


gkinsman:
JustSaying currently has no support for deciding which topic to publish to at publish time. Target topics are determined on startup through configuration.

We want to support deciding the topic to be published to at publish time to support publishing to tenanted topics.

Possible API:
`WithTopicCustomiser(Func<Message, string originalTopic, string resultTopic> customiser)`

 Problems:
- How do we support creating `SnsTopicByName`'s at publish time. Currently they're created on startup to ensure that infrastructure exists before publishing. 
  - Support creation of these at publish time
    - Was attempted in JustSayingStack, had potential for deadlocks
    - Used by LambdaWorker
  - Require user to specify possible topics at startup, and fail at publish time if not registered
- Extend support of infrastructureless mode to add templated ARN's for topic providers.
  - Assumes infra already exists, don't need to create infra on the fly

PR #1010, opened 2022-05-13
Support dynamic publishers

gkinsman:
This PR adds a new mechanism to support publishing to tenanted topics using the solution outlined in #1004. 

It adds a new `TopicPublicationBuilder<T> WithTopicName(Func<Message, string> topicNameCustomizer)` overload to `TopicPublicationBuilder<T>` that allows users to dynamically create a topic at publish time instead of statically defining them on startup.

### DIFF
--- a/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
+++ b/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
@@ -63,7 +63,7 @@ internal sealed class DynamicMessagePublisher : IMessagePublisher
         _logger.LogDebug("Executing startup task for topic {TopicName}", topicName);
         await config.StartupTask(cancellationToken).ConfigureAwait(false);
 
-        _publisherCache.TryAdd(topicName, config.Publisher);
+        _ = _publisherCache.TryAdd(topicName, config.Publisher);
 
         _logger.LogDebug("Publishing message on newly created topic {TopicName}", topicName);
         await config.Publisher.PublishAsync(message, metadata, cancellationToken).ConfigureAwait(false);
--- a/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
+++ b/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
@@ -1,0 +1,71 @@
+using System.Collections.Concurrent;
+using JustSaying.Messaging;
+using JustSaying.Messaging.Interrogation;
+using JustSaying.Models;
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.Fluent;
+
+internal class DynamicMessagePublisher : IMessagePublisher
+{
+    private readonly ConcurrentDictionary<string, IMessagePublisher> _publisherCache = new();
+    private readonly Func<Message, string> _topicNameCustomizer;
+    private readonly Func<string, StaticPublicationConfiguration> _staticConfigBuilder;
+
+    private readonly ConcurrentDictionary<string, SemaphoreSlim> _topicCreationLocks = new();
+    private readonly ILogger<DynamicMessagePublisher> _logger;
+
+    public DynamicMessagePublisher(Func<Message, string> topicNameCustomizer, Func<string, StaticPublicationConfiguration> staticConfigBuilder, ILoggerFactory loggerFactory)
+    {
+        _topicNameCustomizer = topicNameCustomizer;
+        _staticConfigBuilder = staticConfigBuilder;
+        _logger = loggerFactory.CreateLogger<DynamicMessagePublisher>();
+    }
+
+    public InterrogationResult Interrogate()
+    {
+        return new InterrogationResult(new
+        {
+
+        });
+    }
+
+    public Task StartAsync(CancellationToken stoppingToken)
+    {
+        return Task.CompletedTask;
+    }
+
+    public async Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
+    {
+        var topicName = _topicNameCustomizer(message);
+        if (_publisherCache.TryGetValue(topicName, out var publisher))
+        {
+            await publisher.PublishAsync(message, metadata, cancellationToken).ConfigureAwait(false);
+            return;
+        }
+
+        var lockObj = _topicCreationLocks.GetOrAdd(topicName, _ => new SemaphoreSlim(1, 1));
+
+        _logger.LogDebug("Publisher for topic {TopicName} not found, waiting on creation lock", topicName);
+        await lockObj.WaitAsync(cancellationToken).ConfigureAwait(false);
+        if (_publisherCache.TryGetValue(topicName, out var thePublisher))
+        {
+            _logger.LogDebug("Lock re-entrancy detected, returning existing publisher");
+            await thePublisher.PublishAsync(message, metadata, cancellationToken).ConfigureAwait(false);
+            return;
+        }
+
+        _logger.LogDebug("Lock acquired to init topic {TopicName}", topicName);
+        var config = _staticConfigBuilder(topicName);
+        _logger.LogDebug("Executing startup task for topic {Top ame}", topicName);
+        await config.StartupTask(cancellationToken).ConfigureAwait(false);
+
+        _publisherCache.TryAdd(topicName, config.Publisher);
+
+        _logger.LogDebug("Publishing message on newly created topic {TopicName}", topicName);
+        await config.Publisher.PublishAsync(message, metadata, cancellationToken).ConfigureAwait(false);
+    }
+
+    public Task PublishAsync(Message message, CancellationToken cancellationToken)
+        => PublishAsync(message, null, cancellationToken);
+}
--- a/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
+++ b/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
@@ -58,7 +58,7 @@ internal sealed class DynamicMessagePublisher : IMessagePublisher
             return;
         }
 
-        _logger.LogDebug("Lock acquired to init topic {TopicName}", topicName);
+        _logger.LogDebug("Lock acquired to initialize topic {TopicName}", topicName);
         var config = _staticConfigBuilder(topicName);
         _logger.LogDebug("Executing startup task for topic {TopicName}", topicName);
         await config.StartupTask(cancellationToken).ConfigureAwait(false);
--- a/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
+++ b/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
@@ -24,9 +24,12 @@ internal class DynamicMessagePublisher : IMessagePublisher
 
     public InterrogationResult Interrogate()
     {
+        var pairs = _publisherCache.Keys.OrderBy(x => x)
+            .ToDictionary(x => x, x => _publisherCache[x].Interrogate());
+
         return new InterrogationResult(new
         {
-
+            Publishers = pairs
         });
     }
 
--- a/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
+++ b/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
@@ -6,7 +6,7 @@ using Microsoft.Extensions.Logging;
 
 namespace JustSaying.Fluent;
 
-internal class DynamicMessagePublisher : IMessagePublisher
+internal sealed class DynamicMessagePublisher : IMessagePublisher
 {
     private readonly ConcurrentDictionary<string, IMessagePublisher> _publisherCache = new();
     private readonly Func<Message, string> _topicNameCustomizer;
--- a/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
+++ b/src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
@@ -60,7 +60,7 @@ internal sealed class DynamicMessagePublisher : IMessagePublisher
 
         _logger.LogDebug("Lock acquired to init topic {TopicName}", topicName);
         var config = _staticConfigBuilder(topicName);
-        _logger.LogDebug("Executing startup task for topic {Top ame}", topicName);
+        _logger.LogDebug("Executing startup task for topic {TopicName}", topicName);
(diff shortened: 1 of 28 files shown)


src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    public Task StartAsync(CancellationToken stoppingToken)
+    {
+        return Task.CompletedTask;
+    }
+
+    public async Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
+    {
+        var topicName = _topicNameCustomizer(message);
+        if (_publisherCache.ContainsKey(topicName))
```
martincostello, line 45: Use `TryGetValue()`, not `ContainsKey()` + `[]`.

martincostello, line 45: Should this be `ConcurrentDictionary` too to prevent a race?

src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            await _publisherCache[topicName].PublishAsync(message, metadata, cancellationToken);
+            return;
+        }
+
+        var lockObj = _topicCreationLocks.GetOrAdd(topicName, _ => new SemaphoreSlim(1, 1));
```
martincostello, line 51: `SemaphoreSlim` implements `IDisposable`, so we'd need a way to tidy things up.

src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
```diff
@@ -0,0 +1,97 @@
...
+
+
+        _logger.LogDebug("Publisher for topic {TopicName} not found, waiting on creation lock", topicName);
+        await lockObj.WaitAsync(cancellationToken);
+        if (_publisherCache.ContainsKey(topicName))
```
martincostello, line 56: `TryGetValue()`

src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        _logger.LogDebug("Lock acquired to init topic {TopicName}", topicName);
+        var config = _staticConfigBuilder(topicName);
+        _logger.LogDebug("Executing startup task for topic {TopicName}", topicName);
+        await config.StartupTask(cancellationToken);
+
+        _publisherCache.Add(topicName, config.Publisher);
+
+        _logger.LogDebug("Publishing message on newly created topic {TopicName}", topicName);
+        await _publisherCache[topicName].PublishAsync(message, metadata, cancellationToken);
```
martincostello, line 71: You already have the object so don't need to go back into the dictionary.

```suggestion
        await config.Publisher.PublishAsync(message, metadata, cancellationToken);
```

src/JustSaying/Fluent/PublishConfig/StaticPublicationConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+            bus.SerializationRegister,
+            loggerFactory,
+            bus.Config.MessageSubjectProvider)
+        {
+            MessageResponseLogger = bus.Config.MessageResponseLogger,
+        };
+
+        var snsTopic = new SnsTopicByName(
+            readConfiguration.TopicName,
+            proxy.GetAwsClientFactory().GetSnsClient(RegionEndpoint.GetBySystemName(region)),
```
martincostello, line 54: Should they share the client, rather than have one each?

src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        });
+    }
+
+    public Task StartAsync(CancellationToken stoppingToken)
+    {
+        return Task.CompletedTask;
+    }
+
+    public async Task PublishAsync(Message message, PublishMetadata metadata, CancellationToken cancellationToken)
```
martincostello, line 42: A number of missing `ConfigureAwait(false)` calls.

src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
gkinsman: We're not accessing `AvailableWaitHandle` so we don't need to call `Dispose` as it would have no effect 👍

src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
gkinsman: Ah yeah fair - I thought it'd be ok as it'd only be concurrent readers on different keys, but other threads might be adding additional publishers so probably best to be on the safe side 👍 .

src/JustSaying/Fluent/PublishConfig/StaticPublicationConfiguration.cs
gkinsman: It's an interesting one! I don't see why not - seems like a waste. We do currently have a different client per publisher for other topics, but we should probably be sharing them more. Will do this 👍

gkinsman: Done 👍

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -0,0 +1,74 @@
+using System.Collections.Concurrent;
+using JustSaying.Messaging;
+using JustSaying.Messaging.Interrogation;
+using JustSaying.Models;
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.Fluent;
+
+internal class DynamicMessagePublisher : IMessagePublisher
```
martincostello, line 9:
```suggestion
internal sealed class DynamicMessagePublisher : IMessagePublisher
```

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        _topicNameCustomizer = topicNameCustomizer;
+        _staticConfigBuilder = staticConfigBuilder;
+        _logger = loggerFactory.CreateLogger<DynamicMessagePublisher>();
+    }
+
+    public InterrogationResult Interrogate()
+    {
+        var pairs = _publisherCache.Keys.OrderBy(x => x)
+            .ToDictionary(x => x, x => _publisherCache[x].Interrogate());
```
martincostello, line 28: Would this operation be thread-safe between accessing the keys and getting the value from the cache?

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -0,0 +1,74 @@
...
+        _logger.LogDebug("Publisher for topic {TopicName} not found, waiting on creation lock", topicName);
+        await lockObj.WaitAsync(cancellationToken).ConfigureAwait(false);
+        if (_publisherCache.TryGetValue(topicName, out var thePublisher))
+        {
+            _logger.LogDebug("Lock re-entrancy detected, returning existing publisher");
+            await thePublisher.PublishAsync(message, metadata, cancellationToken).ConfigureAwait(false);
+            return;
+        }
+
+        _logger.LogDebug("Lock acquired to init topic {TopicName}", topicName);
```
martincostello, line 61:
```suggestion
        _logger.LogDebug("Lock acquired to initialize topic {TopicName}", topicName);
```

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -0,0 +1,74 @@
...
+        var config = _staticConfigBuilder(topicName);
+        _logger.LogDebug("Executing startup task for topic {Top ame}", topicName);
```
martincostello, line 63:
```suggestion
        _logger.LogDebug("Executing startup task for topic {TopicName}", topicName);
```

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -0,0 +1,74 @@
...
+        await config.StartupTask(cancellationToken).ConfigureAwait(false);
+
+        _publisherCache.TryAdd(topicName, config.Publisher);
```
martincostello, line 66:
```suggestion
        _ = _publisherCache.TryAdd(topicName, config.Publisher);
```

src/JustSaying/Fluent/PublishConfig/DynamicPublicationConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Xml.Linq;
+using JustSaying.AwsTools;
+using JustSaying.AwsTools.QueueCreation;
+using JustSaying.Messaging;
+using JustSaying.Models;
+using Microsoft.Extensions.Logging;
+
+namespace JustSaying.Fluent;
+
+internal class DynamicPublicationConfiguration : ITopicPublisher
```
martincostello, line 11:
```suggestion
internal sealed class DynamicPublicationConfiguration : ITopicPublisher
```

src/JustSaying/Fluent/PublishConfig/ITopicPublisher.cs
```diff
@@ -0,0 +1,9 @@
+using JustSaying.Messaging;
+
+namespace JustSaying.Fluent;
+
+internal interface ITopicPublisher
+{
+    public Func<CancellationToken, Task> StartupTask { get; }
+    public IMessagePublisher Publisher { get; }
```
martincostello, line 8:
```suggestion
    Func<CancellationToken, Task> StartupTask { get; }
    MessagePublisher Publisher { get; }
```

src/JustSaying/Fluent/PublishConfig/StaticPublicationConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+using JustSaying.AwsTools.MessageHandling;
+using JustSaying.AwsTools.QueueCreation;
+using JustSaying.Messaging;
+using Microsoft.Extensions.Logging;
+
+#pragma warning disable CS0618
+
+namespace JustSaying.Fluent;
+
+internal class StaticPublicationConfiguration : ITopicPublisher
```
martincostello, line 15:
```suggestion
internal sealed class StaticPublicationConfiguration : ITopicPublisher
```

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -37,6 +38,12 @@ internal TopicPublicationBuilder()
     /// </summary>
     private string TopicName { get; set; } = string.Empty;
 
+    /// <summary>
+    /// Function that will produce a topic name dynamically from a Message at publish time.
+    /// If the topic doesn't exist, it'll be created at that point.
```
martincostello, line 43:
```suggestion
    /// If the topic doesn't exist, it will be created at that point.
```

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -126,12 +133,29 @@ public TopicPublicationBuilder<T> WithTag(string key, string value)
...
-    public TopicPublicationBuilder<T> WithName(string name)
+    public TopicPublicationBuilder<T> WithTopicName(string name)
     {
         TopicName = name ?? throw new ArgumentNullException(nameof(name));
         return this;
     }
 
+    /// <summary>
+    /// Configures the name of the topic by calling this func at publish time to determine the name of the topic.
+    /// If the topic doesn't exist, it will be created on first publish.
```
martincostello, line 144:
```suggestion
    /// If the topic does not exist, it will be created on first publish.
```

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -126,12 +133,29 @@ public TopicPublicationBuilder<T> WithTag(string key, string value)
...
+    /// </summary>
+    /// <param name="topicNameCustomizer">Function that will be called at publish time to determine the name of the target topic for this <see cref="T"/>.
+    /// <para>
+    /// For example: WithTopicName(msg => $"{msg.Tenant}-mymessage") with msg.Tenant of [uk, au] would
```
martincostello, line 148:
```suggestion
    /// For example: <c>WithTopicName(msg => $"{msg.Tenant}-mymessage")</c> with <c>msg.Tenant</c> of <c>["uk", "au"]</c> would
```

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -126,12 +133,29 @@ public TopicPublicationBuilder<T> WithTag(string key, string value)
...
+    /// create topics "uk-mymessage" and "au-mymessage" when a message is published with those tenants.
```
martincostello, line 149:
```suggestion
    /// create topics <c>"uk-mymessage"</c> and <c>"au-mymessage"</c> when a message is published with those tenants.
```

src/JustSaying/Fluent/TopicPublicationBuilder`1.cs
```diff
@@ -126,12 +133,29 @@ public TopicPublicationBuilder<T> WithTag(string key, string value)
...
+    /// </para>
+    /// </param>
+    /// <returns></returns>
```
martincostello, line 152: Missing doc.

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
gkinsman: No it wouldn't be, although worst case there might be a new publisher that isn't captured in the `Keys` so it wouldn't be emitted. We could add a lock mechanism around it which would block publishers during interrogation, but given it's just interrogation I'm not sure it's worth it? The member accesses themselves are all thread safe so it's just a behaviour thing.

What do you think?

martincostello (approved):

PR #1016, opened 2022-05-18
Use stable localstack tag

martincostello:
Use a stable tag for localstack to fix CI failures.


martincostello:
As far as I can tell, `latest` is literally that - the most recent commit in their default branch. While using it is low-maintenance, it brings the risk of in-development changes breaking us.

gkinsman:
Ahh that looks to have done the trick. Should we depend on the `0.14` tag so that we automatically pick up major releases?

PR #1023, opened 2022-06-08
Move ErrorHandlerMiddleware down next to HandlerInvocationMiddleware, and ensure exceptions are logged

gkinsman:
This PR moves the `ErrorHandlerMiddleware` next to the `HandlerInvocationMiddleware`, so that the first thing that happens after executing user code is handling an exception if one is thrown, and adding it to the `HandleMessageContext` via `SetException`.

Then, it adds the `HandledException` property to the log context so that the error is visible to users. This should have been happening before but the test was lying to us, as it had a bug in it:
image

### DIFF
--- a/tests/JustSaying.IntegrationTests/Logging/LogContextTests.cs
+++ b/tests/JustSaying.IntegrationTests/Logging/LogContextTests.cs
@@ -122,7 +122,12 @@ public class LogContextTests : IntegrationTestBase
             handleMessage.ShouldNotBeNull();
 
             handleMessage.LogLevel.ShouldBe(level);
-            Assert.Equal(exceptionMessage, handleMessage.Exception?.Message);
+
+            if (exceptionMessage != null)
+            {
+                handleMessage.Exception.ShouldNotBeNull();
+                handleMessage.Exception.Message.ShouldBe(exceptionMessage);
+            }
 
             var propertyMap = new Dictionary<string, object>(handleMessage.Properties);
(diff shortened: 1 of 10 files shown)


tests/JustSaying.IntegrationTests/Logging/LogContextTests.cs
```diff
@@ -122,7 +122,7 @@ public async Task HandleMessageFromQueueLogs_ShouldHaveContext(bool handlerShoul
             handleMessage.ShouldNotBeNull();
 
             handleMessage.LogLevel.ShouldBe(level);
-            handleMessage.Exception?.Message.ShouldBe(exceptionMessage);
+            Assert.Equal(exceptionMessage, handleMessage.Exception?.Message);
```
martincostello, line 125:
```suggestion
            handleMessage.Exception.ShouldNotBeNull();
            handleMessage.Exception.Message.ShouldBe(exceptionMessage);
```

gkinsman: This isn't always the case, as its a data driven test, and the exception should only not be null if an exception message is supplied to be thrown. I think we could check to see if the message isnt null and check that way though, to keep things shouldly.

gkinsman: Fixed 👍

martincostello (approved):

PR #1036, opened 2022-07-07
Fix GitHub package publish

martincostello:
Try and fix the publishing to GitHub Packages by cribbing from [webhooks.net](https://github.com/octokit/webhooks.net/blob/fcb76e93d1ade4ced717bdc0c0463c9596591700/.github/workflows/build.yml#L58-L78).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,10 @@ on:
     branches: [main]
   workflow_dispatch:
 
+permissions:
+  contents: read
+  packages: write
+
 env:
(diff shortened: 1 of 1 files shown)


martincostello:
It worked, yet didn't 🤔

https://github.com/justeat/JustSaying/runs/7239660290?check_suite_focus=true#step:8:30

martincostello:
Caused by this: https://docs.microsoft.com/en-us/nuget/release-notes/nuget-6.1#known-issues

PR #1039, opened 2022-07-11
Use TimeSpan for KmsDataKeyReusePeriod

### LINKED ISSUE #829: Encryption uses string for kmsDataKeyReusePeriodSeconds when the docs specify integer
In V7 we allow users to specify `KmsDataKeyReusePeriodSeconds` as a string when configuring encryption.

The SQS docs however state that the value should be an integer: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html. 

I'd guess that if this works currently, it's because SQS attempts to cast the json string to an integer, but we should fix our API to allow specifying a timespan that we convert to the number of seconds.

slang25:
Fixes #829


src/JustSaying/AwsTools/JustSayingConstants.cs
```diff
@@ -71,5 +71,5 @@ public static class JustSayingConstants
     /// <summary>
     /// Default length of time, in seconds, for which Amazon SQS can reuse a data key to encrypt/decrypt messages before calling AWS KMS again
```
martincostello, line 72:
```suggestion
    /// Default length of time for which Amazon SQS can reuse a data key to encrypt/decrypt messages before calling AWS KMS again.
```

src/JustSaying/AwsTools/QueueCreation/ServerSideEncryption.cs
```diff
@@ -7,14 +7,10 @@ public class ServerSideEncryption
...
-        KmsDataKeyReusePeriodSeconds = JustSayingConstants.DefaultAttributeEncryptionKeyReusePeriodSecond;
+        KmsDataKeyReusePeriod = JustSayingConstants.DefaultAttributeEncryptionKeyReusePeriod;
     }
 
-    [JsonProperty("kmsMasterKeyId")]
-    [JsonPropertyName("kmsMasterKeyId")]
     public string KmsMasterKeyId { get; set; }
 
-    [JsonProperty("kmsDataKeyReusePeriodSeconds")]
-    [JsonPropertyName("kmsDataKeyReusePeriodSeconds")]
```
slang25, line 18: It turns out that we don't actually serialize this type 🤷‍♂️

src/JustSaying/AwsTools/QueueCreation/SqsBasicConfiguration.cs
```diff
@@ -50,6 +50,12 @@ public void Validate()
                 $"Invalid configuration. {nameof(DeliveryDelay)} must be between {JustSayingConstants.MinimumDeliveryDelay} and {JustSayingConstants.MaximumDeliveryDelay}.");
         }
 
+        if (ServerSideEncryption.KmsDataKeyReusePeriod.TotalSeconds is > 86400 or < 60)
```
martincostello, line 53: _Maybe_ would be more readable comparing `KmsDataKeyReusePeriod` to `TimeSpan.FromSeconds(1)` and `TimeSpan.FromHours(24)` (or `TimeSpan.FromDays(1)`)?

PR #1059, opened 2022-09-14
Remove JustSaying specific attributes types

slang25:
Historically we added our own `MessageAttributesValue` type, which was basically a copy of the SNS/SQS implementations (which are also identical, so we have 3 copies of the same type in the mix).

This was done so that `JustSaying.Models` package had no dependency on the AWS SDK.

Now that we've moved message attributes to a `PublisherMetadata` type for publishing, and `MessageContext` for subscribing, we could use the SDK types in those places as they don't belong to the `JustSaying.Models` package.

This is primarily to avoid confusion when looking at the message context `Message` property, which has a `MessageAttributes` property that will always be empty, this is because AWS gives us all of the attributes in the message body, but not in the wrapping XML payload (you have to explicitly ask for it in the request). Rather than ask for all attributes in the receive message request, this change populates the property from the message body.

After finishing this change, I actually think it might be cleaner to use our own type on both publishing and consuming, and document the quirk with `MessageAttributes` on the message context. We could even add an option or change the default to request all message attributes, which also solves it.


src/JustSaying/Messaging/MessageSerialization/IMessageSerialiser.cs
```diff
@@ -1,13 +1,13 @@
...
+using Message = JustSaying.Models.Message;
 
 namespace JustSaying.Messaging.MessageSerialization;
 
 public interface IMessageSerializer
 {
     string GetMessageSubject(string sqsMessage);
 
-    MessageAttributes GetMessageAttributes(string message);
+    Dictionary<string, MessageAttributeValue> GetMessageAttributes(string message);
```
martincostello, line 10: Should we use `IDictionary<K, V>`, or is it better to keep the concrete type?

src/JustSaying/Messaging/MessageSerialization/MessageWithAttributes.cs
```diff
@@ -22,5 +22,16 @@ public MessageWithAttributes(Message message, MessageAttributes messageAttribute
...
-    public MessageAttributes MessageAttributes { get; }
-}
\ No newline at end of file
+    public Dictionary<string, MessageAttributeValue> MessageAttributes { get; }
+
+    /// <summary>
+    /// Deconstructs the instance into message and attributes.
+    /// </summary>
+    /// <param name="message"></param>
+    /// <param name="attributes"></param>
```
martincostello, line 31: Nit: add text.

src/JustSaying/Messaging/MessageSerialization/SystemTextJsonSerializer.cs
```diff
@@ -55,32 +56,32 @@ public string GetMessageSubject(string sqsMessage)
         }
     }
 
-    public MessageAttributes GetMessageAttributes(string message)
+    public Dictionary<string, MessageAttributeValue> GetMessageAttributes(string message)
     {
-        var jsonDocument = JsonDocument.Parse(message);
+        var attributes = new Dictionary<string, MessageAttributeValue>();
 
+        var jsonDocument = JsonDocument.Parse(message);
```
martincostello, line 63:
```suggestion
        using var jsonDocument = JsonDocument.Parse(message);
```

PR #1062, opened 2022-09-16
Message context docs

slang25:


docs/subscriptions/message-context.md
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+### Attributes
+
+On the `MessageContext` type, there are the following places to get attributes:
+
+**Context.MessageAttributes**
+
+Message attributes are user defined attributes, and the attributes here are retrieved from the message body.
```
martincostello, line 44:
```suggestion
Message attributes are user-defined attributes, and the attributes here are retrieved from the message body.
```

docs/subscriptions/message-context.md
```diff
@@ -0,0 +1,52 @@
...
+
+**Context.Message.MessageAttributes**
+
+`Message` is the AWS SQS SDK type, the `MessageAttributes` here are the ones that present in the `MessageAttributes` XML payload that the message is wrapped in. When using JustSaying, this property will always be empty as we populate `Context.MessageAttributes` from the message body, so don't request to include any message attributes in the reponse object.
```
martincostello, line 48:
```suggestion
`Message` is the AWS SQS SDK type, the `MessageAttributes` here are the ones that are present in the `MessageAttributes` XML payload that the message is wrapped in. When using JustSaying, this property will always be empty as we populate `Context.MessageAttributes` from the message body, so don't request to include any message attributes in the response object.
```

docs/subscriptions/message-context.md
```diff
@@ -0,0 +1,52 @@
...
+
+**Context.Message.Attributes**
+
+These are the "system attributes" for the message, in JustSaying we request for the `ApproximateReceiveCount` system attribute to be included, so this should be only present attribute here, and will always be available.
```
martincostello, line 52:
```suggestion
These are the _"system attributes"_ for the message; in JustSaying we request for the `ApproximateReceiveCount` system attribute to be included, so this should be the only present attribute here, and will always be available.
```

docs/subscriptions/message-context.md
```diff
@@ -0,0 +1,52 @@
+# Message Context
+
+When creating a handler for a message, you provide an `IHandler<T>` implementation that will process the JustSaying message. However, sometimes you want to access the raw SQS message itself, i.e. an [Amazon.SQS.Model.Message](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/SQS/TMessage.html) instance. This can be useful for accessing non-domain concepts, such as message attributes like `ApproximateReceiveCount` or the message receipt handle.
+
+JustSaying allows you to get access to this through the type `MessageContext`, which you can access by injecting an `IMessageContextAccessor` into your handler.
+
+`IMessageContextAccessor` internally stores the `MessageContext` for the current message being processed in an `AsyncLocal<MessageContext>`, so the appropriate instance will be available anywhere in the logical call stack while handling a message.
```
hwoodiwiss, line 7: Is it worth linking to the docs on AsyncLocal? 

```suggestion
`IMessageContextAccessor` internally stores the `MessageContext` for the current message being processed in an [`AsyncLocal<MessageContext>`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1), so the appropriate instance will be available anywhere in the logical call stack while handling a message.
```

martincostello (approved):

docs/SUMMARY.md
```diff
@@ -19,6 +19,7 @@
 * [Middleware](subscriptions/middleware/README.md)
   * [Custom Middleware](subscriptions/middleware/custom-middleware.md)
 * [Dead Letter Queues](subscriptions/dead-letter-queues.md)
+* [Message Context](messaging-configur/message-context.md)
```
hwoodiwiss, line 22:
```suggestion
* [Message Context](messaging-configuration/message-context.md)
```

PR #1080, opened 2022-10-23
Feature/improve arn parsing and IAM Policy testability

PeterKneale:
- I noticed `SnsPolicy.cs` was using regex to parse the source ARN instead of using Amazons supplied `ARN.Parse(..)` method.
- Before changing this I separated the two responsibilities of `generating` and `applying` the policy
- This allowed me to add an `approval test` against the IAM policy generated
- Then I made the actual change and introduced `ARN.Parse(..)` and observed no regressions.
- I also added a mechanism to validate the ARN I saw elsewhere.

### DIFF
--- a/src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
+++ b/src/JustSaying/AwsTools/MessageHandling/SnsPolicy.cs
@@ -1,5 +1,3 @@
-using System.Text.Json;
-using System.Text.RegularExpressions;
 using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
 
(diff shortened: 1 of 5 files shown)


PeterKneale:
Is this sort of small improvement okay to add without a ticket - Please let me know if I overstep :)

PeterKneale:
I can do the same for `SqsPolicy` if desired

slang25 (approved):
Nice!

slang25:
Thanks for this @PeterKneale, these improvements are always appreciated 🙂. We don't mind skipping issues when there isn't a design discussion to be had.

PR #1081, opened 2022-10-23
Improve arn parsing and testability of IAM SQS policy generation

PeterKneale:
I noticed `SqsPolicy.cs` was using string manipulation to parse the source ARN instead of using Amazons supplied `ARN.Parse(..) `method.
Before changing this I separated the two responsibilities of `generating` and `applying` the policy
This allowed me to add an approval test against the IAM policy generated
Then I made the actual change and introduced `ARN.Parse(..)` and observed no regressions.
Added an additional test that covers the `wildcard topic` scenario.

### DIFF
--- a/src/JustSaying/AwsTools/MessageHandling/SqsPolicyBuilder.cs
+++ b/src/JustSaying/AwsTools/MessageHandling/SqsPolicyBuilder.cs
@@ -1,0 +1,45 @@
+﻿using Amazon;
+
+namespace JustSaying.AwsTools.MessageHandling;
+
+internal static class SqsPolicyBuilder{
+
+    public static string BuildPolicyJson(SqsPolicyDetails policyDetails)
+    {
+        var sid = Guid.NewGuid().ToString().Replace("-", "");
+
+        var resource = policyDetails.QueueArn;
+
+        var topicArnWildcard = string.IsNullOrWhiteSpace(policyDetails.SourceArn)
+            ? "*"
+            : CreateTopicArnWildcard(policyDetails.SourceArn);
+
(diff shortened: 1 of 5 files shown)


src/JustSaying/AwsTools/MessageHandling/SqsPolicyBuilder.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Amazon;
+
+namespace JustSaying.AwsTools.MessageHandling;
+
+internal static class SqsPolicyBuilder{
+
+    public static string BuildPolicyJson(SqsPolicyDetails policyDetails)
```
hwoodiwiss, line 7: Similar to @adammorr's request on #1082, could this be made an extension method for SqsPolicyDetails?

hwoodiwiss (changes requested):
Thanks for the contribution! 😄 
Just one change, and it's essentially the same as what's been suggested in your other PR.

src/JustSaying/AwsTools/MessageHandling/SqsPolicyBuilder.cs
PeterKneale: @hwoodiwiss I've updated https://github.com/justeat/JustSaying/pull/1082 to make them both extension methods.
I've left this PR to just introduce the approval tests and seperation of the policy generation

hwoodiwiss: Great stuff!

hwoodiwiss (approved):
lgtm!

Issue #1095, opened 2022-11-08
Add support for Batch operation


lillo42:
Hi,

Today AWS has support for batch operation by sending 10 by 10 messages for [SNS ](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-sns-supports-publishing-batches-messages-single-api-request/) and [SQS](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html), and I think it would be nice have support for it. Today I'm doing it send message by message; I guess, if we have it, I could reduce my AWS call and speed up my API/Lambda.

martincostello:
Resolved by #1335.

PR #1097, opened 2022-11-10
Update SQS & SNS package

lillo42:
Update SQS & SNS packages to the latest version

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,8 +6,8 @@
     <NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <AwsSdkSnsVersion>3.7.0</AwsSdkSnsVersion>
-    <AwsSdkSqsVersion>3.7.0</AwsSdkSqsVersion>
+    <AwsSdkSnsVersion>3.7.100.14</AwsSdkSnsVersion>
+    <AwsSdkSqsVersion>3.7.100.14</AwsSdkSqsVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="coverlet.msbuild" Version="3.2.0" PrivateAssets="All" />
--- a/src/JustSaying/JustSaying.csproj
+++ b/src/JustSaying/JustSaying.csproj
@@ -11,8 +11,8 @@
     <InternalsVisibleTo Include="JustSaying.UnitTests" Key="$(JustSayingKey)" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.0" />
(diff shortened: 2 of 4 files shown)


lillo42:
it's necessary to add support to send messages in batch https://github.com/justeat/JustSaying/issues/1095

martincostello:
I think this would be better incorporated into a pull request to actually add batch send functionality if we add one. Otherwise a consuming application can just update the dependency versions themselves.

adammorr:
I believe @hwoodiwiss was going to take a look at the batch publishing but I could be wrong

Directory.Build.props
```diff
@@ -6,8 +6,8 @@
     <NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <AwsSdkSnsVersion>3.7.0</AwsSdkSnsVersion>
-    <AwsSdkSqsVersion>3.7.0</AwsSdkSqsVersion>
+    <AwsSdkSnsVersion>3.7.100.14</AwsSdkSnsVersion>
```
slang25, line 9: I think here we are looking for the lowest version that has the binary compatibility we want, so here `3.7.100`

martincostello: LambdaWorker supports batch sends, and that depends on:

```xml
<PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.3" />
<PackageVersion Include="AWSSDK.SQS" Version="3.7.0" />
```

src/JustSaying/JustSaying.csproj
```diff
@@ -11,8 +11,8 @@
     <InternalsVisibleTo Include="JustSaying.UnitTests" Key="$(JustSayingKey)" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.0" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.7.0" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="$(AwsSdkSnsVersion)" />
+    <PackageReference Include="AWSSDK.SQS" Version="$(AwsSdkSqsVersion)" />
```
martincostello, line 15: These used a specific version rather than the variable so they could be pinned independently of them being increased for the test projects.

Directory.Build.props
martincostello: It would make sense that batching support arrived in both SDKs in the same version. As Stu says, we should use the earliest possible version that works for the use cases, not whatever happens to be the latest one right now.

lillo42:
@martincostello OK, I'll close this one and open another when I have the batch implemented, and I'll revert the changes on csproj

lillo42:
@hwoodiwiss sure, no problem. Also I'm already working on this 😄

PR #1098, opened 2022-11-14
GH-1095 Add support to batch operation

lillo42:
[GH-1095](https://github.com/justeat/JustSaying/issues/1095) This PR adds support to batch operation to SNS & SQS

- Updating AWS SDK (SNS & SQS) to the latest version
- Creating a new interface called `IMessageBatchPublisher` (avoiding breaking changes)
- Creating batch loggers & handler exception filter

Consideration:
Because we need to chuck the batch message and the current supported doesn't have the `Chuck `method, I've copied the current implement from [dotnet repository](https://github.com/dotnet/runtime/blob/aa0550291bc7aea587a8e5d97777cd828d9a2f01/src/libraries/System.Linq/src/System/Linq/Chunk.cs)

TODO:
- [x] Add Unit Test
- [x] Add Integration Test
- [x] Add Batch sample


src/JustSaying/Messaging/MessageBatchPublisherExtensions.cs
```diff
@@ -1,29 +0,0 @@
-using System.ComponentModel;
-using JustSaying.Models;
-
-namespace JustSaying.Messaging;
-
-[EditorBrowsable(EditorBrowsableState.Never)]
-public static class MessageBatchPublisherExtensions
```
martincostello: Do you _have_ to delete this?

It's a binary breaking change and would otherwise require us to release JustSaying v8, and we've already just done a large breaking change with v7.

Directory.Build.props
```diff
@@ -6,8 +6,8 @@
     <NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <AwsSdkSnsVersion>3.7.0</AwsSdkSnsVersion>
-    <AwsSdkSqsVersion>3.7.0</AwsSdkSqsVersion>
+    <AwsSdkSnsVersion>3.7.100</AwsSdkSnsVersion>
+    <AwsSdkSqsVersion>3.7.100</AwsSdkSqsVersion>
```
martincostello, line 10: As noted in the previous PR, will this work with just a minimum of `3.7.3`?

src/JustSaying/AwsTools/MessageHandling/MessageBatchResponse.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Net;
+using Amazon.Runtime;
+using JustSaying.Messaging;
+
+namespace JustSaying.AwsTools.MessageHandling;
+
+public class MessageBatchResponse
+{
+    public IEnumerable<string> SuccessfulMessageIds { get; set; }
+    public IEnumerable<string> FailedMessageIds { get; set; }
```
martincostello, line 10: Prefer `ICollection<T>` or `IList<T>` as it guards against things like LINQ queries being assigned that then get enumerated multiple times.

src/JustSaying/AwsTools/MessageHandling/SqsMessagePublisher.cs
```diff
@@ -112,4 +114,121 @@ public InterrogationResult Interrogate()
...
+            }
+
+            if (response != null)
+            {
+                using (_logger.BeginScope(new Dictionary<string, string>
+                {
+                    ["AwsRequestId"] = response.ResponseMetadata?.RequestId
+                }))
+                {
+                    if (response.Successful.Count > 0)
```
martincostello, line 153: To remove the need to redundant enumeration of the response if logging isn't enabled at this level for this logger:

```suggestion
                    if (response.Successful.Count > 0 && _logger.IsEnabled(LogLevel.Information))
```

src/JustSaying/AwsTools/MessageHandling/SqsMessagePublisher.cs
```diff
@@ -112,4 +114,121 @@ public InterrogationResult Interrogate()
...
+
+                    if (response.Failed.Count > 0)
+                    {
+                        _logger.LogError(
+                            "Fail to published batch of {MessageCount} to {DestinationType} '{MessageDestination}'.",
+                            response.Failed.Count,
+                            "Queue",
+                            request.QueueUrl);
+
+                        foreach (var message in response.Failed)
```
martincostello, line 180: Same here for the warning level.

src/JustSaying/AwsTools/MessageHandling/SqsMessagePublisher.cs
```diff
@@ -112,4 +114,121 @@ public InterrogationResult Interrogate()
...
+                    }
+                }
+            }
+
+            if (MessageBatchResponseLogger != null)
+            {
+                var responseData = new MessageBatchResponse
+                {
+                    SuccessfulMessageIds = response?.Successful.Select(x => x.MessageId),
+                    FailedMessageIds = response?.Failed.Select(x => x.Id),
```
martincostello, line 199: As noted in the previous comment, this ensures the queries are materialised once.

```suggestion
                    SuccessfulMessageIds = response?.Successful.Select(x => x.MessageId).ToArray(),
                    FailedMessageIds = response?.Failed.Select(x => x.Id).ToArray(),
```

src/JustSaying/AwsTools/MessageHandling/SqsMessagePublisher.cs
```diff
@@ -112,4 +114,121 @@ public InterrogationResult Interrogate()
...
+    {
+        if (QueueUrl is null)
+        {
+            throw new PublishException("Queue URL was null, perhaps you need to call `StartAsync` on the `IMessagePublisher` before publishing.");
+        }
+
+        var size = metadata?.BatchSize ?? 10;
+        size = Math.Min(size, 10);
+
+        foreach (var chuck in messages.Chunk(size))
```
martincostello, line 131: Typo: `chunk` not `chuck`.

src/JustSaying/AwsTools/MessageHandling/SqsMessagePublisher.cs
```diff
@@ -112,4 +114,121 @@ public InterrogationResult Interrogate()
...
+                var entry = new SendMessageBatchRequestEntry
+                {
+                    Id = message.UniqueKey(),
+                    MessageBody = GetMessageInContext(message)
+                };
+
+                if (metadata?.Delay != null)
+                {
+                    entry.DelaySeconds = (int)metadata.Delay.Value.TotalSeconds;
+                }
```
martincostello, line 225: I _think_ you'd be able to simplify this to:

```suggestion
                if (metadata?.Delay is { } delay)
                {
                    entry.DelaySeconds = (int)delay.TotalSeconds;
                }
```

src/JustSaying/Extensions/ChunkExtensions.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System.Diagnostics;
+
+namespace JustSaying.Extensions;
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
martincostello, line 6: Presumably we only need this if targeting below .NET 6 so this could be wrapped in a hash-define?

src/JustSaying/Extensions/ChunkExtensions.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System.Diagnostics;
+
+namespace JustSaying.Extensions;
```
martincostello, line 3: You could put this into the same namespace as the original code, then with a hash define we can just not include it at all and use the built-in version.

src/JustSaying/Fluent/AccountAddressProvider.cs
```diff
@@ -106,10 +109,14 @@ public Uri GetQueueUriByConvention<T>()
     /// <returns>The <see cref="Uri"/> for this queue.</returns>
     public Uri GetQueueUri(string queueName)
     {
-        var hostname = _regionEndpoint.GetEndpointForService("sqs").Hostname;
-        return new UriBuilder("https", hostname)
+        var endpoint = new AmazonSQSEndpointProvider().ResolveEndpoint(new SQSEndpointParameters
```
martincostello, line 112: Can we store the provider in a static readonly field, rather than make a new one every time the method is called?

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -71,4 +72,47 @@ public async Task PublishAsync(Message message, PublishMetadata metadata, Cancel
...
+                continue;
+            }
+
+            _logger.LogDebug("Lock acquired to initialize topic {TopicName}", topicName);
+            var config = _staticConfigBuilder(topicName);
+            _logger.LogDebug("Executing startup task for topic {TopicName}", topicName);
+            await config.StartupTask(cancellationToken).ConfigureAwait(false);
+
+
+            _ = _batchPublisherCache.TryAdd(topicName, config.BatchPublisher);
```
martincostello, line 110: What happens here if `TryAdd()` returns false?

src/JustSaying/Fluent/QueueAddress.cs
```diff
@@ -79,11 +81,12 @@ public static QueueAddress FromArn(string queueArn)
         if (!Arn.TryParse(queueArn, out var arn)) throw new ArgumentException("Must be a valid ARN.", nameof(queueArn));
         if (!string.Equals(arn.Service, "sqs", StringComparison.OrdinalIgnoreCase)) throw new ArgumentException("Must be an ARN for an SQS queue.", nameof(queueArn));
 
-        var hostname = RegionEndpoint.GetBySystemName(arn.Region)
-            .GetEndpointForService("sqs")
-            .Hostname;
+        var endpoint = new AmazonSQSEndpointProvider().ResolveEndpoint(new SQSEndpointParameters
```
martincostello, line 84: Same comment as before regarding the static field.

src/JustSaying/Fluent/TopicAddressPublicationBuilder`1.cs
```diff
@@ -42,6 +43,22 @@ public TopicAddressPublicationBuilder<T> WithExceptionHandler(Func<Exception, Me
         return this;
     }
 
+    /// <summary>
+    /// Configures an exception handler to use.
+    /// </summary>
+    /// <param name="exceptionBatchHandler">A delegate to invoke if an exception is thrown while publishing in batch.</param>
```
martincostello, line 49:
```suggestion
    /// <param name="exceptionBatchHandler">A delegate to invoke if an exception is thrown while publishing a batch.</param>
```

src/JustSaying/IPublishConfiguration.cs
```diff
@@ -8,5 +8,6 @@ public interface IPublishConfiguration
     int PublishFailureReAttempts { get; set; }
     TimeSpan PublishFailureBackoff { get; set; }
     Action<MessageResponse, Message> MessageResponseLogger { get; set; }
+    Action<MessageBatchResponse, IEnumerable<Message>> MessageBatchResponseLogger { get; set; }
```
martincostello, line 11: This is a binary-breaking change. Can it be broken out to not be so?

src/JustSaying/JustSaying.csproj
```diff
@@ -11,8 +11,8 @@
     <InternalsVisibleTo Include="JustSaying.UnitTests" Key="$(JustSayingKey)" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.0" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.7.0" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.100" />
+    <PackageReference Include="AWSSDK.SQS" Version="3.7.100" />
```
martincostello, line 15: 3.7.3?

```suggestion
    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.100" />
    <PackageReference Include="AWSSDK.SQS" Version="3.7.100" />
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -108,8 +110,27 @@ public void AddMessageMiddleware<T>(string queueName, HandleMessageMiddleware mi
         }
 
         _publishersByType[typeof(T)] = messagePublisher;
+        if(messagePublisher is IMessageBatchPublisher batchPublisher)
```
martincostello, line 113:
```suggestion
        if (messagePublisher is IMessageBatchPublisher batchPublisher)
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -108,8 +110,27 @@ public void AddMessageMiddleware<T>(string queueName, HandleMessageMiddleware mi
...
+        {
+            _batchPublishersByType[typeof(T)] = batchPublisher;
+        }
     }
 
+    public void AddMessageBatchPublisher<T>(IMessageBatchPublisher messageBatchPublisher) where T : Message
+    {
+        if (Config.PublishFailureReAttempts == 0)
+        {
+            _log.LogWarning("You have not set a re-attempt value for publish failures. If the publish location is 'down' you may lose messages.");
```
martincostello, line 123:
```suggestion
            _log.LogWarning("You have not set a re-attempt value for publish failures. If the publish location is not available you may lose messages.");
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -108,8 +110,27 @@ public void AddMessageMiddleware<T>(string queueName, HandleMessageMiddleware mi
...
+        }
+
+        _batchPublishersByType[typeof(T)] = messageBatchPublisher;
+        if(messageBatchPublisher is IMessagePublisher messagePublisher)
```
martincostello, line 127:
```suggestion
        if (messageBatchPublisher is IMessagePublisher messagePublisher)
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -289,4 +310,86 @@ public void Dispose()
...
+    public Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken)
+    {
+        if (!_busStarted && _startupTasks.Count > 0)
+        {
+            throw new InvalidOperationException("There are pending startup tasks that must be executed by calling StartAsync before messages may be published.");
+        }
+
+        var tasks = (from @group in messages.GroupBy(x => x.GetType())
+            let publisher = GetBatchPublishersForMessageType(@group.Key)
+            select PublishAsync(publisher, @group.ToList(), metadata, 0, @group.Key, cancellationToken))
```
martincostello, line 326: Please use the not-SQL syntax to match the rest of the code base.

src/JustSaying/Messaging/MessagePublisherExtensions.cs
```diff
@@ -34,4 +34,34 @@ await publisher.PublishAsync(message, metadata, CancellationToken.None)
...
     }
-}
\ No newline at end of file
+
+    public static Task PublishAsync(this IMessagePublisher publisher, IEnumerable<Message> messages)
+        => publisher.PublishAsync(messages, null, CancellationToken.None);
+    public static Task PublishAsync(this IMessagePublisher publisher, IEnumerable<Message> messages, CancellationToken cancellationToken)
+        => publisher.PublishAsync(messages, null, cancellationToken);
+    public static Task PublishAsync(this IMessagePublisher publisher, IEnumerable<Message> messages, PublishBatchMetadata metadata)
+        => publisher.PublishAsync(messages, metadata, CancellationToken.None);
```
martincostello, line 43:
```suggestion
    public static Task PublishAsync(this IMessagePublisher publisher, IEnumerable<Message> messages)
        => publisher.PublishAsync(messages, null, CancellationToken.None);

    public static Task PublishAsync(this IMessagePublisher publisher, IEnumerable<Message> messages, CancellationToken cancellationToken)
        => publisher.PublishAsync(messages, null, cancellationToken);

    public static Task PublishAsync(this IMessagePublisher publisher, IEnumerable<Message> messages, PublishBatchMetadata metadata)
        => publisher.PublishAsync(messages, metadata, CancellationToken.None);
```

src/JustSaying/MessagingConfig.cs
```diff
@@ -21,6 +21,7 @@ public MessagingConfig()
     public int PublishFailureReAttempts { get; set; }
     public TimeSpan PublishFailureBackoff { get; set; }
     public Action<MessageResponse, Message> MessageResponseLogger { get; set; }
+    public Action<MessageBatchResponse, IEnumerable<Message>> MessageBatchResponseLogger { get; set; }
```
martincostello, line 24: I think this would probably be better with `ICollection<T>` or `IList<T>` as well so consumers can access the count easily (as we already know it internally).

src/JustSaying/Messaging/MessageBatchPublisherExtensions.cs
lillo42: I've just created this file, but I deleted it.
We don't had this file before

src/JustSaying/IPublishConfiguration.cs
gkinsman: The response logger is legacy and its job is better performed by middleware now, so would suggest leaving this out and maybe even marking the old response logger as obsolete 👍.

Directory.Build.props
lillo42: For SQS, there is no 3.7.3 😢 

Is it OK to have SQS & SNS in different versions?

```c#
    <AwsSdkSnsVersion>3.7.3</AwsSdkSnsVersion>
    <AwsSdkSqsVersion>3.7.0</AwsSdkSqsVersion>
```

martincostello: SQS [3.7.1.43](https://www.nuget.org/packages/AWSSDK.SQS/3.7.1.43) came out on the same day as [3.7.3](https://www.nuget.org/packages/AWSSDK.SimpleNotificationService/3.7.3) for SNS.

SQS [3.7.2](https://www.nuget.org/packages/AWSSDK.SQS/3.7.2) then came out a week later.

We should use whichever the earliest version that has the batch functionality is. You might have to experiment to find which one it is.

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
lillo42: Looking at the code, this shouldn't happen since we have a semaphore that allows 1 per time, but let's imagine that happens; I guess we are going to 2 instances of `IMessageBatchPublisher`. GC will collect the newest one after some time.

src/JustSaying/Fluent/AccountAddressProvider.cs
lillo42: Since I've changed the package to 3.7.3 & 3.7.2, It's not necessary, I've rollback to the old code

src/JustSaying/Fluent/AccountAddressProvider.cs
```diff
@@ -109,12 +106,8 @@ public Uri GetQueueUriByConvention<T>()
...
     public Uri GetQueueUri(string queueName)
     {
-        var endpoint = new AmazonSQSEndpointProvider().ResolveEndpoint(new SQSEndpointParameters
-        {
-            Region = _regionEndpoint.SystemName
-        });
-
-        return new UriBuilder(endpoint.URL)
+        var hostname = _regionEndpoint.GetEndpointForService("sqs").Hostname;
+        return new UriBuilder("https", hostname)
```
martincostello, line 110: https://learn.microsoft.com/dotnet/api/system.uri.urischemehttps

```suggestion
        return new UriBuilder(Uri.UriSchemeHttps, hostname)
```

src/JustSaying/Fluent/QueueAddress.cs
```diff
@@ -81,14 +79,13 @@ public static QueueAddress FromArn(string queueArn)
...
-        var endpoint = new AmazonSQSEndpointProvider().ResolveEndpoint(new SQSEndpointParameters
-        {
-            Region = arn.Region
-        });
+        var hostname = RegionEndpoint.GetBySystemName(arn.Region)
+            .GetEndpointForService("sqs")
+            .Hostname;
 
-        var queueUrl = new UriBuilder(endpoint.URL)
+        var queueUrl = new UriBuilder("https", hostname)
```
martincostello, line 86:
```suggestion
        var queueUrl = new UriBuilder(Uri.UriSchemeHttps, hostname)
```

src/JustSaying/Extensions/ChunkExtensions.cs
slang25: Perhaps we should start to include `net6.0` in the built TFMs, in which case @hwoodiwiss suggested we conditionally include this file for non-net6 tfms.

martincostello: I think that's already been done.

slang25: Ah I see, the `#if !NET6_0_OR_GREATER` thing, yeah this will always evaluate to `true`, I'm suggesting adding `net6.0` as a TFM, it increases the package size a little, but we can start to use the newer stuff when it's available.

src/JustSaying/Messaging/IMessageBatchPublisher.cs
```diff
@@ -0,0 +1,11 @@
+﻿using JustSaying.Messaging.Interrogation;
+using JustSaying.Models;
+
+namespace JustSaying.Messaging;
+
+public interface IMessageBatchPublisher : IInterrogable, IStartable
+{
+    Task PublishAsync(IEnumerable<Message> messages, CancellationToken cancellationToken);
+    Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken);
```
hwoodiwiss, line 9: Asking to start the discussion, should we cover the use-case where a user needs to have custom metadata per message being sent?

@slang25 and I chatted about this earlier and thought it might be best to leave it for now, and add it later if it's specifically requested.

slang25:
This is looking good @lillo42 Sorry we've gone a bit quiet here, do you think this PR is ready?

src/JustSaying/Messaging/IMessageBatchPublisher.cs
```diff
@@ -5,7 +5,7 @@ namespace JustSaying.Messaging;
 
 public interface IMessageBatchPublisher : IInterrogable, IStartable
 {
-    Task PublishAsync(IEnumerable<Message> messages, CancellationToken cancellationToken);
-    Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken);
+    Task PublishAsync(IEnumerable<Message> messages, CancellationToken cancellationToken = default);
+    Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken = default);
```
martincostello, line 9: I think this counts as a breaking change.

samples/src/JustSaying.Sample.Restaurant.OrderingApi/Program.cs
```diff
@@ -74,10 +74,7 @@
     builder.Services.AddHostedService<BusService>();
 
     builder.Services.AddEndpointsApiExplorer();
-    builder.Services.AddSwaggerGen(c =>
-    {
-        c.SwaggerDoc("v1", new OpenApiInfo { Title = "Restaurant Ordering API", Version = "v1" });
-    });
+    builder.Services.AddSwaggerGen(c => { c.SwaggerDoc("v1", new OpenApiInfo { Title = "Restaurant Ordering API", Version = "v1" }); });
```
martincostello, line 77:
```suggestion
    builder.Services.AddSwaggerGen(c => c.SwaggerDoc("v1", new OpenApiInfo { Title = "Restaurant Ordering API", Version = "v1" }));
```

src/JustSaying/Messaging/IMessageBatchPublisher.cs
lillo42: The IMessageBatchPublisher.cs is a new interface(created on this PR), also I guess the optional parameter is a syntax sugar

lillo42:
@slang25 I think the PR is ready for review now

src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs
```diff
@@ -26,8 +29,9 @@ public class SnsMessagePublisher : IMessagePublisher, IInterrogable
         IMessageSerializationRegister serializationRegister,
         ILoggerFactory loggerFactory,
         IMessageSubjectProvider messageSubjectProvider,
-        Func<Exception, Message, bool> handleException = null)
-        : this(client, serializationRegister, loggerFactory, messageSubjectProvider, handleException)
+        Func<Exception, Message, bool> handleException = null,
+        Func<Exception, IReadOnlyCollection<Message>, bool> handleBatchException = null)
```
martincostello, line 33: This (and the change below) is a binary-breaking change as it adds new optional parameters.

If we can avoid needing a v8 that would be preferred, so instead you'd need to add a completely new constructor overload.

src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs
```diff
@@ -150,4 +157,109 @@ public virtual InterrogationResult Interrogate()
...
         });
     }
+
+    public Task PublishAsync(IEnumerable<Message> messages, CancellationToken cancellationToken)
+        => PublishAsync(messages, null, cancellationToken);
+
+    public async Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken)
+    {
+        var size = metadata?.BatchSize ?? 10;
+        size = Math.Min(size, 10);
```
martincostello, line 167: Let's make the `10` a named constant, rather than a repeated magic number.

src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs
```diff
@@ -150,4 +157,109 @@ public virtual InterrogationResult Interrogate()
...
+            {
+                response = await Client.PublishBatchAsync(request, cancellationToken).ConfigureAwait(false);
+            }
+            catch (AmazonServiceException ex)
+            {
+                if (!ClientExceptionHandler(ex, chunk))
+                {
+                    throw new PublishBatchException($"Failed to publish batch of messages to SNS. Topic ARN: '{request.TopicArn}'.", ex);
+                }
+            }
```
martincostello, line 183: We should probably log something here.

src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs
```diff
@@ -150,4 +157,109 @@ public virtual InterrogationResult Interrogate()
...
+
+            if (response is { })
+            {
+                using (_logger.BeginScope(new Dictionary<string, string>
+                       {
+                           ["AwsRequestId"] = response.ResponseMetadata?.RequestId
+                       }))
+                {
```
martincostello, line 191: Could simplify this to something like:

```csharp
using scope = _logger.BeginScope(new Dictionary<string, string> { ["AwsRequestId"] = response.ResponseMetadata?.RequestId })
```

src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs
```diff
@@ -150,4 +157,109 @@ public virtual InterrogationResult Interrogate()
...
+        return new PublishBatchRequest
+        {
+            TopicArn = Arn,
+            PublishBatchRequestEntries = message.Select(message => new PublishBatchRequestEntry
+                {
+                    Id = message.UniqueKey(),
+                    Subject = _messageSubjectProvider.GetSubjectForType(message.GetType()),
+                    Message = _serializationRegister.Serialize(message, serializeForSnsPublishing: true),
+                    MessageAttributes = BuildMessageAttributes(metadata)
+                })
```
martincostello, line 261: This would be better to do manually in a loop as there's a lot of methods being called here, which makes it more difficult to debug if it's a lambda inside a `Select()`.

src/JustSaying/AwsTools/MessageHandling/SqsMessagePublisher.cs
```diff
@@ -112,4 +114,121 @@ public InterrogationResult Interrogate()
...
     }
+
+    public Task PublishAsync(IEnumerable<Message> messages, CancellationToken cancellationToken)
+        => PublishAsync(messages, null, cancellationToken);
+
+    public async Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken)
+    {
+        if (QueueUrl is null)
+        {
+            throw new PublishException("Queue URL was null, perhaps you need to call `StartAsync` on the `IMessagePublisher` before publishing.");
```
martincostello, line 125:
```suggestion
            throw new PublishException($"Queue URL was null. Perhaps you need to call the ${nameof(IMessagePublisher.StartAsync)} method on the ${nameof(IMessagePublisher)} before publishing.");
```

src/JustSaying/AwsTools/MessageHandling/SqsMessagePublisher.cs
```diff
@@ -112,4 +114,121 @@ public InterrogationResult Interrogate()
...
+                    ex);
+            }
+
+            if (response != null)
+            {
+                using (_logger.BeginScope(new Dictionary<string, string>
+                {
+                    ["AwsRequestId"] = response.ResponseMetadata?.RequestId
+                }))
+                {
```
martincostello, line 152: Similar comment to above.

src/JustSaying/Fluent/MessagingConfigurationBuilder.cs
```diff
@@ -334,4 +390,27 @@ public IMessagingConfig Build()
 
         return config;
     }
-}
\ No newline at end of file
+
+    public IPublishBatchConfiguration BuildPublishBatchConfiguration()
```
martincostello, line 394: As this is public, needs `///` documentation.

src/JustSaying/JustSayingBus.cs
```diff
@@ -289,4 +334,93 @@ public void Dispose()
...
     }
+
+    public Task PublishAsync(IEnumerable<Message> messages, CancellationToken cancellationToken)
+        => PublishAsync(messages, null, cancellationToken);
+
+    public Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken)
+    {
+        if (!_busStarted && _startupTasks.Count > 0)
+        {
+            throw new InvalidOperationException("There are pending startup tasks that must be executed by calling StartAsync before messages may be published.");
```
martincostello, line 345: Use `nameof()` for the method name.

src/JustSaying/JustSayingBus.cs
```diff
@@ -289,4 +334,93 @@ public void Dispose()
...
+        }
+
+        var tasks = new List<Task>();
+        foreach (IGrouping<Type, Message> group in messages.GroupBy(x => x.GetType()))
+        {
+            IMessageBatchPublisher publisher = GetBatchPublishersForMessageType(group.Key);
+            tasks.Add(PublishAsync(publisher, group.ToList(), metadata, 0, group.Key, cancellationToken));
+        }
+
+        return Task.WhenAll(tasks);
```
martincostello, line 355: Is there an overload of this that supports passing the `CancellationToken` through?

src/JustSaying/JustSayingBus.cs
```diff
@@ -289,4 +334,93 @@ public void Dispose()
...
+    private async Task PublishAsync(
+        IMessageBatchPublisher publisher,
+        List<Message> messages,
+        PublishBatchMetadata metadata,
+        int attemptCount,
+        Type messageType,
+        CancellationToken cancellationToken)
+    {
+        var batchSize = metadata?.BatchSize ?? 10;
+        batchSize = Math.Min(batchSize, 10);
```
martincostello, line 385: +1 magic number.

martincostello:
@lillo42 Are you going to come back to this? I've just found a need for this myself.

martincostello:
I'm going to close this as it's been dormant for nearly a year now, and shortly open a new PR based on this code as a starting point.

PR #1134, opened 2023-02-26
Add dotnet outdated

martincostello:
- Add `dotnet-outdated-tool` to also update .NET NuGet packages, if there are any relevant updates, when the .NET SDK is updated.
- Change the update schedule to be less frequent.

### DIFF
--- a/.github/workflows/update-dotnet-sdk.yml
+++ b/.github/workflows/update-dotnet-sdk.yml
@@ -1,8 +1,14 @@
 name: update-dotnet-sdk
 
+env:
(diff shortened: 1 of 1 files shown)


slang25:
Very cool. It might be overkill, but dotnet-outdated now also supports outputting Markdown which could be useful for displaying what's changed:
https://github.com/dotnet-outdated/dotnet-outdated/pull/343

slang25 (approved):

martincostello:
Ah nice. I'll have a play with that later today in one of my personal projects and see how I get on.

martincostello:
@slang25 Looking at the PR, am I correct that it only supports _one_ output formatter?

PR #1144, opened 2023-04-01
GH-1139: Use AWSSDK Policy Creation

### LINKED ISSUE #1139: Remove `SqsPolicy` from JustSaying
I found that during the subscription creation we're rewritting Policy attribute. And this one is done in a very strict maner - all existing policy statements are being deleted. See usage [here](https://github.com/justeattakeaway/JustSaying/blob/c3b98d536535f49d5fc3657ef4789aba9fc8a963/src/JustSaying/AwsTools/QueueCreation/AmazonQueueCreator.cs#L68) and implementation [here](https://github.com/justeattakeaway/JustSaying/blob/c3b98d536535f49d5fc3657ef4789aba9fc8a963/src/JustSaying/AwsTools/MessageHandling/SqsPolicy.cs#L10).

The main thing is that we don't need this operation at all because inside `IAmazonSimpleNotificationService.SubscribeQueueAsync` is well implemented policy updater that will keep existing statements & add new based on the TopicArn. See the [link](https://github.com/aws/aws-sdk-net/blob/4c0f4e67589c21a04e3be4282c9d17c41c4cd99b/sdk/src/Services/SimpleNotificationService/Custom/_async/AmazonSimpleNotificationServiceClient.Extensions.cs#L132) to AWS SDK. 

Suggested solution: remove custom policy updater from `JustSaying` because
1) it's an unnecessary API call 
2) we don't want to override existing statements inside policy 

Possible solution: allow `JustSaying` to set their custom policy over policy that is created via AWS SDK (but not override it)

hwoodiwiss:
Removes the SqsPolicyBuilder in favour of AWSSDK inbuilt policy creation.

fixes #1139


hwoodiwiss:
I don't _think_ this change is breaking, as the only use case I could find for using the SQS ploicy builder was after we had just created a topic by name, and direcly passed the ARN of that new topic to the builder, so I can't think of a way that we would have generated wildcard policies.

tests/JustSaying.IntegrationTests/Fluent/AwsTools/Approvals/WhenEnsuringATopicExistsPolicyIsCreated.Then_The_Iam_Policy_Is_Created.approved.txt
```diff
@@ -8,10 +8,10 @@
                 "AWS" : "*"
             },
             "Action"    : "sqs:SendMessage",
-            "Resource"  : "",
+            "Resource"  : "arn:aws:sqs:us-east-1:000000000000:<unique-name>",
             "Condition" : {
                 "ArnLike" : {
-                    "aws:SourceArn" : "*"
+                    "aws:SourceArn" : "arn:aws:sns:us-east-1:000000000000:<unique-name>-topic"
```
hwoodiwiss, line 14: I could be wrong, but I _think_ the us-east-1 here is an localstack thing

slang25: It is, localstack lets you override it by specifying a `DEFAULT_REGION` env var

hwoodiwiss: I'll give it a go with the region set just to make sure nothing weird is happening with the region being sent

slang25 (approved):
It looks good 🙂

tests/JustSaying.IntegrationTests/Fluent/AwsTools/Approvals/WhenEnsuringATopicExistsPolicyIsCreated.Then_The_Iam_Policy_Is_Created.approved.txt
hwoodiwiss: Yeah, looks good

image

PR #1165, opened 2023-05-05
ISSUE-1155 Update getting-started.md

bjornsy-tl:
Adding information about IMessageReceiveToggle, from https://github.com/justeattakeaway/JustSaying/pull/1164

https://github.com/justeattakeaway/JustSaying/issues/1155

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -139,6 +139,6 @@ public class BusService : BackgroundService
 ```
 
 * Note here that we're calling `StartAsync` on both the `IMessagingBus`, and `IMessagePublisher`. Starting the `IMessagePublisher` ensures that all publish infrastructure is created and ready, while starting the `IMessagingBus` will firstly ensure that all subscription infrastructure is created and ready, and then begin consuming messages.
-* The only way to stop the bus is to cancel the cancellation token, which will tear down everything; however you can effectively pause listening on the bus by
-using the methods on `IMessageReceiveToggle` to control the receiving of all messages, even while the bus is running.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -139,5 +139,6 @@ public class BusService : BackgroundService
 ```
 
 * Note here that we're calling `StartAsync` on both the `IMessagingBus`, and `IMessagePublisher`. Starting the `IMessagePublisher` ensures that all publish infrastructure is created and ready, while starting the `IMessagingBus` will firstly ensure that all subscription infrastructure is created and ready, and then begin consuming messages.
-* The only way to stop the bus is to cancel the cancellation token, which will tear down everything.
+* The only way to stop the bus is to cancel the cancellation token, which will tear down everything; however you can effectively pause listening on the bus by
+using the methods on a singleton instance of `MessageReceiveController` to control the receiving of all messages, even while the bus is running.
```
martincostello, line 143: This should point the user to the interface in DI, rather than the concrete type. Otherwise if the implementation were changed in DI, using the concrete type wouldn't do anything because the instance actually being used by the internals would be something else.

martincostello:
I think it would be better to just include this change in #1164. They need to be in sync with each other, and we wouldn't merge this without it.

bjornsy-tl:
> I think it would be better to just include this change in #1164. They need to be in sync with each other, and we wouldn't merge this without it.

Maybe I'm missing something, but I couldn't see this file in the `main` branch, hence the separate PR for docs

martincostello:
Sorry - my bad, I'd forgotten we have that separate 😄

slang25 (approved):
Thanks for following up with this. We should have `7.0.2` available on nuget.org imminently 🙂

PR #1239, opened 2023-09-05
Add actionlint

martincostello:
Add GitHub Actions workflow to lint workflows with actionlint.

### DIFF
--- a/.github/workflows/lint-actions.yml
+++ b/.github/workflows/lint-actions.yml
@@ -1,4 +1,4 @@
-name: lint
+name: lint-actions
 
 on:
   push:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,39 @@
+name: lint
```
hwoodiwiss, line 1: This doesn't give a great title in the status checks
image

I think the following would be better, as it would give us `actionlint/lint`
```suggestion
name: actionlint

```

martincostello: I'd left it generic so we could add other linters in the future, but I've changed it to `lint-actions`.

hwoodiwiss (approved):

PR #1295, opened 2023-11-19
Update localstack to 3.0.0

slang25:
This version includes support for the SQS JSON protocol

### DIFF
--- a/tests/JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
+++ b/tests/JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
@@ -12,8 +12,8 @@
     <ProjectReference Include="..\JustSaying.TestingFramework\JustSaying.TestingFramework.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.SimpleNotificationService" />
-    <PackageReference Include="AWSSDK.SQS" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" VersionOverride="3.7.300.5" />
+    <PackageReference Include="AWSSDK.SQS" VersionOverride="3.7.300.5" />
     <PackageReference Include="GitHubActionsTestLogger" />
     <PackageReference Include="MartinCostello.Logging.XUnit" />
     <PackageReference Include="MELT.Xunit" />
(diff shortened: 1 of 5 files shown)


tests/JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
```diff
@@ -12,8 +12,8 @@
     <ProjectReference Include="..\JustSaying.TestingFramework\JustSaying.TestingFramework.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AWSSDK.SimpleNotificationService" />
-    <PackageReference Include="AWSSDK.SQS" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" VersionOverride="3.7.300.5" />
+    <PackageReference Include="AWSSDK.SQS" VersionOverride="3.7.300.5" />
```
martincostello, line 16: TIL - does this work with dependabot?

tests/JustSaying.UnitTests/Messaging/Channels/Fakes/FakeAmazonSqs.cs
```diff
@@ -48,6 +49,11 @@ public Task<string> AuthorizeS3ToSendMessageAsync(string queueUrl, string bucket
         return Task.FromResult(new AddPermissionResponse());
     }
 
+    public Task<CancelMessageMoveTaskResponse> CancelMessageMoveTaskAsync(CancelMessageMoveTaskRequest request, CancellationToken cancellationToken = new CancellationToken())
```
martincostello, line 52: We should change all these optional parameters to use the simplified syntax:

```suggestion
    public Task<CancelMessageMoveTaskResponse> CancelMessageMoveTaskAsync(CancelMessageMoveTaskRequest request, CancellationToken cancellationToken = default)
```

tests/JustSaying.IntegrationTests/JustSaying.IntegrationTests.csproj
slang25: Good question, I don't know, would be good to find out. There is also a feature that I can't find where you define versions in groups, and you can say in the package reference "I want the version from group A", can't remember how that works, so doing this for now.

martincostello: I guess we can merge it and if it messes stuff up we can try something else.

martincostello (approved):

PR #1308, opened 2024-01-02
Di aws config

luthervd:
This adds the ability to use `AWSSDK.Extensions.NETCore.Setup` package. This should result in less boilerplate code checking for test/dev environments. We can use the profile or config files to drive the behaviour we want rather than baking it into code.


Directory.Packages.props
```diff
@@ -3,13 +3,15 @@
...
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="1.1.0" Condition=" '$(TargetFramework)' == 'net461' " />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.0.0" Condition=" '$(TargetFramework)' == 'netstandard2.0' " />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageVersion Include="StructureMap" Version="4.6.0" />
     <PackageVersion Include="System.Text.Json" Version="4.6.0" />
     <PackageVersion Include="System.Threading.Channels" Version="4.5.0" />
+    <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="3.7.300" />
```
martincostello, line 14: Nit: maintain sorting.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/GlobalUsings.cs
martincostello, line 1: Please use `<Using Include="..." />` MSBuild items, rather than this file.

Example:

https://github.com/justeattakeaway/JustSaying/blob/b626a63ee8ea22d7348f263806c356430308e798/tests/JustSaying.UnitTests/JustSaying.UnitTests.csproj#L24-L28

Directory.Packages.props
martincostello, line 14: Please add both packages to [dependabot.yml](https://github.com/justeattakeaway/JustSaying/blob/main/.github/dependabot.yml) as an ignored dependency so we "pin" the versions the library depends on.

src/JustSaying.Extensions.DependencyInjection.AwsCore/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
+using Amazon.SimpleNotificationService;
+using Amazon.SQS;
+using JustSaying;
+using JustSaying.AwsTools;
+using JustSaying.Extensions.DependencyInjection.AwsCore;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using System.ComponentModel;
+
+namespace Microsoft.Extensions.DependencyInjection;
```
martincostello, line 10:
```suggestion
namespace Microsoft.Extensions.DependencyInjection;

```

src/JustSaying.Extensions.DependencyInjection.AwsCore/AwsConfigResolvingClientFactory.cs
```diff
@@ -0,0 +1,39 @@
+using Amazon;
+using Amazon.SimpleNotificationService;
+using Amazon.SQS;
+using JustSaying.AwsTools;
+using Microsoft.Extensions.DependencyInjection;
+
+
+namespace JustSaying.Extensions.DependencyInjection.AwsCore;
```
martincostello, line 8:
```suggestion
using Microsoft.Extensions.DependencyInjection;

namespace JustSaying.Extensions.DependencyInjection.AwsCore;
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/AwsConfigResolvingClientFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+
+internal class AwsConfigResolvingClientFactory : IAwsClientFactory
+{
+
```
martincostello, line 12:
```suggestion
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/AwsConfigResolvingClientFactory.cs
```diff
@@ -0,0 +1,39 @@
+using Amazon;
+using Amazon.SimpleNotificationService;
+using Amazon.SQS;
+using JustSaying.AwsTools;
+using Microsoft.Extensions.DependencyInjection;
+
+
+namespace JustSaying.Extensions.DependencyInjection.AwsCore;
+
+internal class AwsConfigResolvingClientFactory : IAwsClientFactory
```
martincostello, line 10:
```suggestion
internal sealed class AwsConfigResolvingClientFactory : IAwsClientFactory
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/AwsConfigResolvingClientFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+
+    private readonly IServiceProvider _serviceProvider;
+    public AwsConfigResolvingClientFactory(IServiceProvider provider)
```
martincostello, line 14:
```suggestion
    private readonly IServiceProvider _serviceProvider;

    public AwsConfigResolvingClientFactory(IServiceProvider provider)
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/AwsConfigResolvingClientFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        _serviceProvider = provider;
+    }
+
+    public IAmazonSimpleNotificationService GetSnsClient(RegionEndpoint region)
+    {
+        var sns = _serviceProvider.GetRequiredService<IAmazonSimpleNotificationService>();
+        if(string.IsNullOrEmpty(sns.Config.ServiceURL) && sns.Config.RegionEndpoint != region)
+        {
+            throw new ArgumentException($"Configured region {sns.Config.RegionEndpoint.DisplayName} does not match the region argument {region.DisplayName}");
```
martincostello, line 24:
```suggestion
            throw new ArgumentException(nameof(region), $"Configured region {sns.Config.RegionEndpoint.DisplayName} does not match the region argument {region.DisplayName}.");
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/AwsConfigResolvingClientFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+        return sns;
+    }
+
+    public IAmazonSQS GetSqsClient(RegionEndpoint region)
+    {
+        var sqs = _serviceProvider.GetRequiredService<IAmazonSQS>();
+        if (string.IsNullOrEmpty(sqs.Config.ServiceURL) && sqs.Config.RegionEndpoint != region)
+        {
+            throw new ArgumentException($"Configured region {sqs.Config.RegionEndpoint.DisplayName} does not match the region argument {region.DisplayName}");
```
martincostello, line 34:
```suggestion
            throw new ArgumentException(nameof(region), $"Configured region {sqs.Config.RegionEndpoint.DisplayName} does not match the region argument {region.DisplayName}.");
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <param name="configure">A delegate to a method to use to configure JustSaying.</param>
+    /// <returns>
+    /// The <see cref="IServiceCollection"/> specified by <paramref name="services"/>.
+    /// </returns>
+    /// <exception cref="ArgumentNullException">
+    /// <paramref name="services"/> is <see langword="null"/>.
+    /// </exception>
+    public static void AddJustSyaingWithAwsConfig(this IServiceCollection services, IConfiguration configuration, Action<MessagingBusBuilder> builderConfig)
+    {
+        AddJustSayingWithAwsConfig(services, configuration,(builder,_) => builderConfig.Invoke(builder));
```
martincostello, line 31:
```suggestion
        AddJustSayingWithAwsConfig(services, configuration, (builder, _) => builderConfig.Invoke(builder));
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <param name="services">The <see cref="IServiceCollection"/> to add JustSaying services to.</param>
+    /// <param name="configuration">The <see cref="IConfiguration"/> used to setup AWS configuration.</param>
+    /// <param name="configure">A delegate to a method to use to configure JustSaying.</param>
+    /// <returns>
+    /// The <see cref="IServiceCollection"/> specified by <paramref name="services"/>.
+    /// </returns>
+    /// <exception cref="ArgumentNullException">
+    /// <paramref name="services"/> is <see langword="null"/>.
+    /// </exception>
+    public static void AddJustSyaingWithAwsConfig(this IServiceCollection services, IConfiguration configuration, Action<MessagingBusBuilder> builderConfig)
```
martincostello, line 29: `builderConfig` needs to be checked for `null`.

src/JustSaying.Extensions.DependencyInjection.AwsCore/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <param name="services">The <see cref="IServiceCollection"/> to add JustSaying services to.</param>
+    /// <param name="configuration">The <see cref="IConfiguration"/> used to setup AWS configuration.</param>
+    /// <param name="configure">A delegate to a method to use to configure JustSaying.</param>
+    /// <returns>
+    /// The <see cref="IServiceCollection"/> specified by <paramref name="services"/>.
+    /// </returns>
+    /// <exception cref="ArgumentNullException">
+    /// <paramref name="services"/> is <see langword="null"/>.
+    /// </exception>
+    public static void AddJustSayingWithAwsConfig(this IServiceCollection serviceCollection, IConfiguration configuration, Action<MessagingBusBuilder, IServiceProvider> builderConfig)
```
martincostello, line 46: Parameters need testing for `null`.

src/JustSaying.Extensions.DependencyInjection.AwsCore/JustSaying.Extensions.DependencyInjection.AwsCore.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\JustSaying.Extensions.DependencyInjection.Microsoft\JustSaying.Extensions.DependencyInjection.Microsoft.csproj" />
+    <ProjectReference Include="..\JustSaying\JustSaying.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.SimpleNotificationService" />
+    <PackageReference Include="AWSSDK.SQS" />
```
martincostello, line 13: These shouldn't be needed as they're transitively referenced in JustSaying.

```suggestion
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -0,0 +1,106 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Configuration;
+using JustSaying.AwsTools;
+using Amazon;
+
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
+
+namespace TestProject1;
```
martincostello, line 9: Proper namespace.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -0,0 +1,106 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Configuration;
+using JustSaying.AwsTools;
+using Amazon;
+
+
```
martincostello, line 6: Rather than me comment on them individually, remove all the double new-lines.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -0,0 +1,106 @@
...
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
martincostello, line 7: Why does this need to be disabled?

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -0,0 +1,106 @@
...
+            .AddJsonFile("appsettings.json")
+            .Build();
+        var serviceCollection = new ServiceCollection();
+        serviceCollection.AddJustSyaingWithAwsConfig(config, (_) => { });
+        var provider = serviceCollection.BuildServiceProvider();
+        var clientFactory = provider.GetRequiredService<IAwsClientFactory>();
+        var snsClient = clientFactory.GetSnsClient(RegionEndpoint.EUWest1);
+        Assert.True(snsClient.Config.ServiceURL.Equals("http://test.test/") == true);
+        Assert.True(snsClient.Config.UseHttp);
+    }
```
martincostello, line 28: Break the tests up into arrange, act and assert blocks. Otherwise it's just a wall of code.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -0,0 +1,106 @@
...
+
+
+    [Fact]
+    public void ShouldReadFromDefaultProfile()
+    {
+        Environment.SetEnvironmentVariable("AWS_PROFILE", null);
+        var bin = AppContext.BaseDirectory;
+        Environment.SetEnvironmentVariable("AWS_CONFIG_FILE", $"{bin}/config");
+        Environment.SetEnvironmentVariable("AWS_SHARED_CREDENTIALS_FILE", $"{bin}/credentials");
```
martincostello, line 37: Use `Path.Join()` so that the correct directory separators are used on Windows.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -0,0 +1,106 @@
...
+        var config = new ConfigurationBuilder()
+            .AddEnvironmentVariables()
```
martincostello, line 39: Rather than set environment variables and need to disable parallelism and reset the variables at the end of the tests (which also won't happen if it fails), use [an in-memory configuration source](https://learn.microsoft.com/dotnet/core/extensions/configuration-providers#memory-configuration-provider) instead.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <NoWarn>$(NoWarn);CA1034;CA1051;CA1054;CA1707;CA2000;CA2007</NoWarn>
+    <TargetFramework>net8.0</TargetFramework>
+    <Deterministic>false</Deterministic>
+  </PropertyGroup>
+  <ItemGroup>
+    <Content Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.SimpleNotificationService" VersionOverride="3.7.300.5" />
+    <PackageReference Include="AWSSDK.SQS" VersionOverride="3.7.300.5" />
```
martincostello, line 12:
```suggestion
    <PackageReference Include="AWSSDK.SimpleNotificationService" VersionOverride="3.7.300.28" />
    <PackageReference Include="AWSSDK.SQS" VersionOverride="3.7.300.28" />
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" />
+    <PackageReference Include="GitHubActionsTestLogger" />
+    <PackageReference Include="MELT" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" />
+    <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="Polly.Core" />
+    <PackageReference Include="xunit" />
+    <PackageReference Include="xunit.runner.visualstudio" />
+  </ItemGroup>
```
martincostello, line 23: I feel you don't need all of these and there's an element of copy-pasta here. Remove the references that aren't actually needed.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests.csproj
```diff
@@ -0,0 +1,44 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <NoWarn>$(NoWarn);CA1034;CA1051;CA1054;CA1707;CA2000;CA2007</NoWarn>
+    <TargetFramework>net8.0</TargetFramework>
+    <Deterministic>false</Deterministic>
+  </PropertyGroup>
+  <ItemGroup>
+    <Content Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
```
martincostello, line 8:
```suggestion
    <Content Include="appsettings.json;config;credentials;xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <None Update="appsettings.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="config">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="credentials">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
martincostello, line 43:
```suggestion
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
luthervd: I have moved the needed env vars up to the constructor so they are only set once. SOme of the tests were doing the wrong thing. Hopefully they read better now. Fixed the Path issue

.github/dependabot.yml
```diff
@@ -14,13 +14,19 @@ updates:
     - "justeattakeaway/justsaying-maintainers"
   open-pull-requests-limit: 10
   ignore:
+  - dependency-name: AWSSDK.Extensions.NETCore.Setup
+    versions:
+    - "> 3.7.300"
```
martincostello, line 19: We can just simplify this to "always ignore" by removing the versions.

```suggestion
  - dependency-name: AWSSDK.Extensions.NETCore.Setup
```

.github/dependabot.yml
```diff
@@ -14,13 +14,19 @@ updates:
...
   - dependency-name: AWSSDK.SimpleNotificationService
     versions:
     - "> 3.3.100.1, < 3.8"
   - dependency-name: AWSSDK.SQS
     versions:
     - "> 3.3.100.1, < 3.8"
   - dependency-name: Microsoft.Bcl.AsyncInterfaces
+  - dependency-name: Microsoft.Extensions.Configuration
+    versions:
+    - "> 8.0.0"
```
martincostello, line 29:
```suggestion
  - dependency-name: Microsoft.Extensions.Configuration
```

src/JustSaying.Extensions.DependencyInjection.AwsCore/IServiceCollectionExtensions.cs
```diff
@@ -26,9 +27,24 @@ public static class IServiceCollectionExtensions
...
+
+        if(configuration is null)
+        {
+            throw new ArgumentNullException(nameof(configuration));
+        }
+
+        if(builderConfig is null)
+        {
+            throw new ArgumentNullException(nameof(builderConfig));
+        }
```
martincostello, line 45:
```suggestion
        if (services is null)
        {
            throw new ArgumentNullException(nameof(services));
        }

        if (configuration is null)
        {
            throw new ArgumentNullException(nameof(configuration));
        }

        if (builderConfig is null)
        {
            throw new ArgumentNullException(nameof(builderConfig));
        }
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
martincostello: Using an in-memory source would still be preferable as it means we aren't changing global process state in the tests.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -31,76 +41,88 @@ public void ShouldReadConfigFromFile()
...
         var provider = serviceCollection.BuildServiceProvider();
         var clientFactory = provider.GetRequiredService<IAwsClientFactory>();
-        var snsClient = clientFactory.GetSnsClient(RegionEndpoint.EUWest1);
+        var snsClient = clientFactory.GetSnsClient(RegionEndpoint.EUWest2);
+
+        //Assert
         Assert.True(string.IsNullOrEmpty(snsClient.Config.ServiceURL), snsClient.Config.ServiceURL);
-        Assert.True(snsClient.Config.RegionEndpoint == RegionEndpoint.EUWest1);
+        Assert.True(snsClient.Config.RegionEndpoint == RegionEndpoint.EUWest2);
         Assert.False(snsClient.Config.UseHttp);
```
martincostello, line 59: Use Shouldy for better assert messages in all these tests (that's what the other tests use).

Also, always use _"Equals(a, b)"_ type constructs not _"True(a == b)"_ ones as you get better error messages if the test fails (_"Expected a but got b"_ instead of _"Expected true but got false"_).

So, for example:

```csharp
snsClient.Config.ServiceURL.ShouldBeNull(); // or empty, whichever is the correct expectation
snsClient.Config.RegionEndpoint.ShouldBe(RegionEndpoint.EUWest2);
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -31,76 +41,88 @@ public void ShouldReadConfigFromFile()
...
+        //Arrange
         var config = new ConfigurationBuilder()
-            .AddEnvironmentVariables()
+            .AddInMemoryCollection(new Dictionary<string, string>
+            {
+                {"AWS:Profile", "local"},
+                {"AWS:ProfilesLocation", _credentialsFile},
+                {"AWS:Region", "us-east-1"},
+                {"AWS:ServiceURL", "http://profile.test/"},
+                {"AWS:UseHttp", "true"}
```
martincostello, line 74: Newer C# syntax to initialise dictionaries is slightly less verbose and nicer to read IMHO:

```suggestion
                ["AWS:Profile"] = "local",
                ["AWS:ProfilesLocation"] = _credentialsFile,
                ["AWS:Region"] = "us-east-1",
                ["AWS:ServiceURL"] = "http://profile.test/",
                ["AWS:UseHttp"] = "true",
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -31,76 +41,88 @@ public void ShouldReadConfigFromFile()
...
+        //Act
         var provider = serviceCollection.BuildServiceProvider();
         var clientFactory = provider.GetRequiredService<IAwsClientFactory>();
         var snsClient = clientFactory.GetSnsClient(RegionEndpoint.USEast1);
-        Assert.True(snsClient.Config.ServiceURL.Equals("http://profile.test/") == true);
-    }
 
+        //Assert
+        Assert.True(snsClient.Config.RegionEndpoint == null);
+        Assert.True(snsClient.Config.ServiceURL.Equals("http://profile.test/") == true, snsClient.Config.ServiceURL);
```
martincostello, line 87:
```suggestion
        snsClient.Config.RegionEndpoint.ShouldBeNull();
        snsClient.Config.ServiceURL.ShouldBe("http://profile.test/");
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -31,76 +41,88 @@ public void ShouldReadConfigFromFile()
...
+
+        //Act
         var provider = serviceCollection.BuildServiceProvider();
+        var clientFactory = provider.GetRequiredService<IAwsClientFactory>();
+        var sqsClient = clientFactory.GetSqsClient(RegionEndpoint.EUWest2);
+        var snsClient = clientFactory.GetSnsClient(RegionEndpoint.EUWest2);
 
+        //Assert
+        Assert.True(sqsClient.Config.RegionEndpoint == RegionEndpoint.EUWest2);
+        Assert.True(snsClient.Config.RegionEndpoint == RegionEndpoint.EUWest2);
```
martincostello, line 108: Same assertion twice - incorrect second assertion, or just duplicated?

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
```diff
@@ -31,76 +41,88 @@ public void ShouldReadConfigFromFile()
...
-        serviceCollection.AddJustSyaingWithAwsConfig(config, (_) => { });
+        serviceCollection.AddJustSayingWithAwsConfig(config, (_) => { });
+
+        //Act
         var provider = serviceCollection.BuildServiceProvider();
         var clientFactory = provider.GetRequiredService<IAwsClientFactory>();
-        Assert.Throws<ArgumentException>(() => clientFactory.GetSnsClient(RegionEndpoint.USEast1));
+
+        //Assert
+        Assert.Throws<ArgumentException>(() => clientFactory.GetSnsClient(RegionEndpoint.USEast2));
```
martincostello, line 126: Exception for Shouldy here - keep xunit's Assert as it has specific overloads for validating these types of exceptions, but change to use the one that validates the argument name, so:

```suggestion
        Assert.Throws<ArgumentException>("region", () => clientFactory.GetSnsClient(RegionEndpoint.USEast2));
```

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests.csproj
```diff
@@ -5,40 +5,21 @@
...
-    <PackageReference Include="Polly.Core" />
     <PackageReference Include="xunit" />
     <PackageReference Include="xunit.runner.visualstudio" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\JustSaying.Extensions.DependencyInjection.AwsCore\JustSaying.Extensions.DependencyInjection.AwsCore.csproj" />
-    <ProjectReference Include="..\JustSaying.TestingFramework\JustSaying.TestingFramework.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <Using Include="Shouldly" />
```
martincostello: This needs putting back as should be using Shouldly like the other tests.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
luthervd: Hey yeah I tried that out and the AWS SDK didn't seem to work without those values being set as environment variables. I will dig into the code for the SDK (if I can find it) and check they don't just read direct from env vars.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
luthervd: One is for the SNS client the other is for the SQS client. Will double check the assignments.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
martincostello: Cool - if that's the way it works (even though `IConfiguation` is passed to them) then that's fine to do that if there's no other way. If you do keep that though, you should make this class implement `IDisposable` and reset the environment variables in the `Dispose()` method so the state is reset between tests.

gkinsman: Hiya! Have been following this with some interest, and thought I'd chime in, but don't want to step on any toes :-). 

I do wonder if it's worth adopting this package if it doesn't follow the rules of loading from configuration sources as is expected in dotnet. 
In my experience this kind of configuration management leads to really brittle test suites that are slow due to lack of parallelization and flaky as tests aren't properly isolated due to the reliance on environmental configuration. The just saying suite should always run in parallel :).

If it's an issue in the upstream library it might be worth raising a bug there and waiting on the fix?

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
martincostello: Ah my bad, I just misread it.

tests/JustSaying.Extensions.DependencyInjection.AwsCore.Tests/ClientFactoryTests.cs
luthervd: Thanks @gkinsman  and @martincostello for the input. I reviewed the tests again and it think the tests relying on env vars weren't adding any value. They were just testing that the AWS dotnet library does what it says it should. I think it seemed a bit pointless testing there code for them here so removed it in the end. This has enabled just using an in memory collection.

build.ps1
```diff
@@ -14,6 +14,7 @@ $sdkFile = Join-Path $solutionPath "global.json"
 $libraryProjects = @(
     (Join-Path $solutionPath "src" "JustSaying" "JustSaying.csproj"),
     (Join-Path $solutionPath "src" "JustSaying.Models" "JustSaying.Models.csproj"),
+    (Join-Path $solutionPath "src" "JustSaying.Extensions.DependencyInjection.AwsCore" "JustSaying.Extensions.DependencyInjection.AwsCore.csproj"),
```
slang25, line 17: I like this PR, it's a very handy addition. Everything is looking good-to-go, I'm just not happy yet with the package name.

I'm struggling with the name mostly because our naming convention is already a mouthful and the package from AWS is badly named, so I don't think we have an obvious nice option.

The `NETCore` part of the upstream AWS library is now out-of-date, so should just be `Net`, but `Net` in namespaces typically means network, so the AWS package should probably just be named `AWSSDK.Extensions.Setup` in hindsight.

I was thinking of something like:
* `JustSaying.Extensions.DependencyInjection.AwsExtensions`
or given that it's not a typical DI package, and it's just piggy-backing off of the `.Microsoft` package, perhaps we just go with:
* `JustSaying.AwsExtensionsSetup`

I'm not too excited by those suggestions, so would love to hear other options.

martincostello: Of the two options above, I think I'd go with `JustSaying.Extensions.DependencyInjection.AwsExtensions` to fit with the other two extension packages.

martincostello:
Once we're decided on a name, this will need rebasing and having the public API analysers added.

slang25 (approved):

PR #1328, opened 2024-01-31
HandlerInvocationMiddleware will resolve services via the IHandlerResolver

mgoodfellow:
# Why?

At the moment when the subscription pipeline is built the handlers are all resolved and cached. This causes an issue with other containers, such as SimpleInjector as we are unable to setup a scope for resolving injected instances within the handler itself.

# What does this do?

Instead of resolving all the handlers when the pipeline is built, we leverage the IHandlerResolver to resolver handlers on demand.

# Example of use

When using SimpleInjector, I could setup a middleware to run earlier in the pipeline to initialise a scope:

```csharp
    public class SimpleInjectorScopeMiddleware : MiddlewareBase<HandleMessageContext, bool>
    {
        private readonly Container _container;

        public SimpleInjectorScopeMiddleware(Container container)
        {
            _container = container;
        }

        protected override async Task<bool> RunInnerAsync(
            HandleMessageContext context,
            Func<CancellationToken, Task<bool>> func,
            CancellationToken stoppingToken)
        {
            await using (AsyncScopedLifestyle.BeginScope(_container))
            {
                return await func(stoppingToken);
            }
        }
    }


    // Somewhere else for convenience
    public static class MiddlewareExtensions
    {
        public static HandlerMiddlewareBuilder UseSimpleInjectorScope(this HandlerMiddlewareBuilder builder)
        {
            builder.Use<SimpleInjectorScopeMiddleware>();

            return builder;
        }
    }
```

This could then be added to a subscription as such:

```csharp
                        x.ForTopic<TestMessage>(
                            cfg =>
                            {
                                cfg.WithMiddlewareConfiguration(m => { m.UseSimpleInjectorScope(); });
                            });
```

Now when receiving a message, my handler will be resolved in scope and I can use injected dependencies correctly.

# Alternatives

There are other ways to solve this problem, the main one would be to allow us to swap in our own `HandlerInvocationMiddleware` - then we could do the scoping in there. However the docs state, and the code seems to reflect that we are unable to provide our own version. If we could, we could set this up as a global middleware and we wouldn't need to set this scoped middleware up on every subscription which would be great.

Another alternative is for us to use a library like Mediator, and every `IHandlerAsync<>` is simply a shell which invokes a mediator pipeline which then sets up our scoping. However, this seems really messy and a bleeding of concerns. In addition it would require us to make sweeping changes across our codebase to convert from v6 to v7.

As an aside, in v6 we define a wrapper when we instantiate the handler during setup which sets up the scope and then resolves the handler inside, effectively acting like a middleware, but this is no longer possible. However for reference:

```csharp
public class HandlerResolverWrapper<T> : IHandlerAsync<T>
{
    private readonly Container _container;

    public HandlerResolverWrapper(Container container)
    {
        _container = container;
    }

    public async Task<bool> Handle(T message)
    {
        await using (AsyncScopedLifestyle.BeginScope(_container))
        {
            try
            {
                var handler = _container.GetInstance<IHandlerAsync<T>>();

                return await handler.Handle(message);
            }
            catch (ActivationException ex)
            {
                Log.Logger.Error(ex, "Failed to find a handler for {messageType}", message.GetType().Name);

                return false;
            }
        }
    }
}
```

And it is configured as such:

```csharp
        stack.WithSqsTopicSubscriber()
                .IntoQueue(null)
                .WithMessageHandler(new HandlerResolverWrapper<TestMessage>(container));
```

Maybe I'm missing something completely obvious, but I have been playing with v7 and this has been a huge blocker for working with SimpleInjector which we use extensively.

If we could resolve this issue using my approach here (or similar), where the handlers are resolved on demand, it would also give cool middleware functionality, such as defining a subscription to have a UoW applied to it.

Order of middlewares would be:
* Scope (setup scope for handler resolution)
* UoW (pull session from container, begin a transaction)
* [internal middlewares, to resolving]
* UoW (handle commit/rollback)
* Scope (close down scope and resolved instances)

This PR isn't ready for merge, it needs a clean up, however from what I can see the `UseHandler<>` I modified was there solely for testing, so making this change would not produce any breaking changes. 

Would love to hear thoughts and start a discussion on this.

Many thanks!

Mike

### DIFF
--- a/src/JustSaying/Messaging/Middleware/Handle/HandlerInvocationMiddleware.cs
+++ b/src/JustSaying/Messaging/Middleware/Handle/HandlerInvocationMiddleware.cs
@@ -8,10 +8,8 @@ namespace JustSaying.Messaging.Middleware;
 /// This middleware is responsible for resolving a message handler and calling it.
 /// </summary>
 /// <typeparam name="T">The type of the message that the message handler handles.</typeparam>
-public sealed class HandlerInvocationMiddleware<T>(Func<HandlerResolutionContext, IHandlerAsync<T>> handlerResolver) : MiddlewareBase<HandleMessageContext, bool> where T : Message
+public sealed class HandlerInvocationMiddleware<T>(IHandlerResolver handlerResolver) : MiddlewareBase<HandleMessageContext, bool> where T : Message
 {
-    private readonly Func<HandlerResolutionContext, IHandlerAsync<T>> _handlerResolver = handlerResolver ?? throw new ArgumentNullException(nameof(handlerResolver));
-
     protected override async Task<bool> RunInnerAsync(
         HandleMessageContext context,
         Func<CancellationToken, Task<bool>> func,
@@ -23,7 +21,7 @@ public sealed class HandlerInvocationMiddleware<T>(Func<HandlerResolutionContext
 
         var resolutionContext = new HandlerResolutionContext(context.QueueName);
 
-        IHandlerAsync<T> handler = _handlerResolver(resolutionContext);
+        var handler = handlerResolver.ResolveHandler<T>(resolutionContext);
 
         return await handler.Handle(context.MessageAs<T>()).ConfigureAwait(false);
     }
--- a/src/JustSaying/Messaging/Middleware/Handle/HandlerInvocationMiddleware.cs
+++ b/src/JustSaying/Messaging/Middleware/Handle/HandlerInvocationMiddleware.cs
@@ -6,9 +6,9 @@ namespace JustSaying.Messaging.Middleware;
 
 /// <summary>
 /// This middleware is responsible for recalling a previously resolved message handler and calling it.
+/// This class is obsolete and will be removed in a future version. Please use HandlerResolveInvocationMiddleware instead.
 /// </summary>
 /// <typeparam name="T">The type of the message that the message handler handles.</typeparam>
-[Obsolete("This class is obsolete and will be removed in a future version. Please use HandlerResolveInvocationMiddleware instead.")]
 public sealed class HandlerInvocationMiddleware<T>(Func<HandlerResolutionContext, IHandlerAsync<T>> handlerResolver) : MiddlewareBase<HandleMessageContext, bool> where T : Message
 {
     private readonly Func<HandlerResolutionContext, IHandlerAsync<T>> _handlerResolver = handlerResolver ?? throw new ArgumentNullException(nameof(handlerResolver));
--- a/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
+++ b/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
@@ -95,7 +95,7 @@ Please check the documentation for your container for more details.");
                 $"Handler middleware has already been specified for {typeof(TMessage).Name} on this queue.");
         }
 
-        _handlerMiddleware = new HandlerInvocationMiddleware<TMessage>(handlerResolver.ResolveHandler<TMessage>);
+        _handlerMiddleware = new HandlerInvocationMiddleware<TMessage>(handlerResolver);
 
         return this;
     }
--- a/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
+++ b/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
@@ -95,7 +95,7 @@ Please check the documentation for your container for more details.");
                 $"Handler middleware has already been specified for {typeof(TMessage).Name} on this queue.");
         }
 
-        _handlerMiddleware = new HandlerInvocationMiddleware<TMessage>(handlerResolver);
+        _handlerMiddleware = new HandlerResolveInvocationMiddleware<TMessage>(handlerResolver);
 
         return this;
     }
--- a/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs
+++ b/src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs
@@ -12,7 +12,7 @@ namespace JustSaying.Messaging.Middleware;
 public static class HandlerMiddlewareBuilderExtensions
 {
     /// <summary>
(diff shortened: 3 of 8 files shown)


src/JustSaying/Messaging/Middleware/Handle/HandlerInvocationMiddleware.cs
```diff
@@ -8,10 +8,8 @@ namespace JustSaying.Messaging.Middleware;
 /// This middleware is responsible for resolving a message handler and calling it.
 /// </summary>
 /// <typeparam name="T">The type of the message that the message handler handles.</typeparam>
-public sealed class HandlerInvocationMiddleware<T>(Func<HandlerResolutionContext, IHandlerAsync<T>> handlerResolver) : MiddlewareBase<HandleMessageContext, bool> where T : Message
+public sealed class HandlerInvocationMiddleware<T>(IHandlerResolver handlerResolver) : MiddlewareBase<HandleMessageContext, bool> where T : Message
```
martincostello, line 11: Just skim reading at this point rather than _properly_ reviewing the change (I'll do that later), but this would be a binary breaking change (as would the other changes to any `public` member), and I don't think we'd want to have to rev to v8 right now.

This would need doing in a way that is opt-in additional functionality, not a change to the existing API surface.

mgoodfellow: Thanks for alerting me to this!

Correct me if I'm wrong, but my latest commit will resolve this issue - we don't change existing public methods, and leave in place the old public classes, and instead mark them obsolete.

We then use new methods and classes for the newer functionality

src/JustSaying/Messaging/Middleware/Handle/HandlerInvocationMiddleware.cs
```diff
@@ -5,9 +5,10 @@
 namespace JustSaying.Messaging.Middleware;
 
 /// <summary>
-/// This middleware is responsible for resolving a message handler and calling it.
+/// This middleware is responsible for recalling a previously resolved message handler and calling it.
 /// </summary>
 /// <typeparam name="T">The type of the message that the message handler handles.</typeparam>
+[Obsolete("This class is obsolete and will be removed in a future version. Please use HandlerResolveInvocationMiddleware instead.")]
```
martincostello, line 11: I would prefer to avoid doing this, as otherwise it's a build-breaking change for people using `TreatWarningsAsErrors=true`.

This feels more like a _differnent_ way to achieve something, rather than a _wrong_ way.

We could obsolete it later if we wanted to do a full-on replace.

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilder.cs
```diff
@@ -95,7 +95,7 @@ public HandlerMiddlewareBuilder Use(Func<HandleMessageMiddleware> middlewareFact
                 $"Handler middleware has already been specified for {typeof(TMessage).Name} on this queue.");
         }
 
-        _handlerMiddleware = new HandlerInvocationMiddleware<TMessage>(handlerResolver.ResolveHandler<TMessage>);
+        _handlerMiddleware = new HandlerResolveInvocationMiddleware<TMessage>(handlerResolver);
```
martincostello, line 98: Could this cause any behavioural changes to existing integrations?

It might be safer to have this be an opt-in feature, and then do one or the other based on that option's value.

src/JustSaying/Messaging/Middleware/Handle/HandlerMiddlewareBuilderExtensions.cs
```diff
@@ -22,6 +22,7 @@ public static class HandlerMiddlewareBuilderExtensions
     /// <exception cref="ArgumentNullException">
     /// <paramref name="builder"/> or <paramref name="handler"/> is <see langword="null"/>.
     /// </exception>
+    [Obsolete("This method is obsolete and will be removed in a future version. Please use UseHandler(IHandlerResolver) instead.")]
```
martincostello, line 25: Same comment here regarding making it obsolete.

src/JustSaying/Messaging/Middleware/Handle/HandlerResolveInvocationMiddleware.cs
```diff
@@ -0,0 +1,28 @@
...
+// ReSharper disable once CheckNamespace
+namespace JustSaying.Messaging.Middleware;
+
+/// <summary>
+/// This middleware is responsible for resolving a message handler and calling it.
+/// </summary>
+/// <typeparam name="T">The type of the message that the message handler handles.</typeparam>
+public sealed class HandlerResolveInvocationMiddleware<T>(IHandlerResolver handlerResolver) : MiddlewareBase<HandleMessageContext, bool> where T : Message
+{
+    protected override async Task<bool> RunInnerAsync(
```
martincostello, line 13:
```suggestion
    /// <inheritdoc />
    protected override async Task<bool> RunInnerAsync(
```

mgoodfellow:
Thanks for taking the time to look at this @martincostello - this is the package I'm looking at open sourcing at the moment which would be dependant on the changes - it's very much a work in progress which has come about from the last few days finally getting around to look at upgrading our infrastructure to v7 of JustSaying: https://github.com/Audiu/JustSaying.Extensions.DependencyInjection.SimpleInjector

I hope that a SimpleInjector DI library to complement the other DI examples would be of some use to other projects.

mgoodfellow:
Hey @martincostello 

Turns out we can scrap this entire PR. The docs are wrong, just found a sample by accident which shows how to setup middleware. Further more, _the defaults can be replaced_

https://justeat.gitbook.io/justsaying/subscriptions/middleware

image

This is incorrect, this setup will not handle messages. Turns out this was my issue.

https://github.com/justeattakeaway/JustSaying/tree/main/samples/src/JustSaying.Sample.Middleware - this actual shows how to do it properly.

Therefore in my case:

```csharp
                 x.ForTopic<TestMessage>(
                            cfg =>
                            {
                                cfg.WithMiddlewareConfiguration(m =>
                                {
                                    m.UseSimpleInjectorScope();
                                    m.UseDefaults<TestMessage>(typeof(TestMessageHandler)); // Add default middleware pipeline
                                });
                            });
```

This now handles messages correctly and also correctly wraps in a scope. Sorry I was misled by the docs and then couldn't see the wood for the trees.

In reality, this PR should be for a docs update, nothing more 😂

PR #1330, opened 2024-01-31
Add public API analyzer

### LINKED ISSUE #1329: Add Public API analyzers
**Is your feature request related to a problem? Please describe.**

Pull requests, such as #1328, may be opened that change the public API surface area in a way that break compatibility that might be unintentional. Ideally contributors should be aware of this at the time of editing code locally, rather than relying on reviewers to notice. Then either they change their design, press ahead as an intentional change we can discuss, or they don't notice and the build breaks and raises awareness.

**Describe the solution you'd like**

Adopt the use of `Microsoft.CodeAnalysis.PublicApiAnalyzers`, [like in httpclient-interception](https://github.com/justeattakeaway/httpclient-interception/blob/0e6a4c1d13c6933f23f27c157319a1f0e44a142d/Directory.Packages.props#L10C30-L10C71).

martincostello:
Add public API analyzer and configure public API baselines.

Resolves #1329.


slang25 (approved):
Very nice

PR #1335, opened 2024-02-01
Support SNS batch publishing

martincostello:
Add support for batch publishing messages to SNS.

Supersedes #1098.

## TODO

- [x] ~~Review original feedback from #1098 and address.~~
- [x] ~~Check test coverage.~~
- [x] ~~Rebase on #1330 and add updated baselines.~~
- [x] ~~Update version to 7.2.0.~~


Directory.Packages.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <ItemGroup Label="Libraries">
-    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.0" />
-    <PackageVersion Include="AWSSDK.SQS" Version="3.7.0" />
+    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.300" />
+    <PackageVersion Include="AWSSDK.SQS" Version="3.7.300" />
```
martincostello, line 4: These are the minimum versions that support .NET 8 and AoT, so I think they're worth raising the baseline too as part of getting the batch publishing support.

src/JustSaying/Fluent/AccountAddressProvider.cs
```diff
@@ -106,10 +106,12 @@ public Uri GetQueueUriByConvention<T>()
     /// <returns>The <see cref="Uri"/> for this queue.</returns>
     public Uri GetQueueUri(string queueName)
     {
+#pragma warning disable CS0618 // Type or member is obsolete
         var hostname = _regionEndpoint.GetEndpointForService("sqs").Hostname;
```
martincostello, line 110: This is obsolete, but using the new way is difficult to do for libraries without breaking compatibility or making things harder for consumers, so keeping using the old way.

src/JustSaying/Fluent/TopicAddressPublisher.cs
martincostello, line 1: This was internal and just syntactic sugar, so I just removed it.

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -62,6 +65,48 @@ public async Task PublishAsync(Message message, PublishMetadata metadata, Cancel
...
+
+            var lockObj = _topicCreationLocks.GetOrAdd(topicName, _ => new SemaphoreSlim(1, 1));
+            _logger.LogDebug("Publisher for topic {TopicName} not found, waiting on creation lock", topicName);
+            await lockObj.WaitAsync(cancellationToken).ConfigureAwait(false);
+            if (_batchPublisherCache.TryGetValue(topicName, out publisher))
+            {
+                _logger.LogDebug("Lock re-entrancy detected, returning existing publisher");
+                publisherTask.Add(publisher.PublishAsync(batchMessages, metadata, cancellationToken));
+                continue;
+            }
```
martincostello, line 96: I think there's a bug here, which is why that one test is failing.

If the batch contains messages that would be sent to different SNS topics based on the payload via the customisation delegate, then that is skipped and all the messages [end up in the same topic as the first message in the batch](https://github.com/justeattakeaway/JustSaying/actions/runs/7745312137/job/21120996624?pr=1335#step:6:369).

martincostello: Fixed by 11a93f01703404c8f2e3658c16e4e68e46e12164, though the test is still failing for some other reason.

src/JustSaying/Fluent/AccountAddressProvider.cs
martincostello: This will need revisiting with AWSSDK v4 as this has been removed completely.

martincostello:
This should be ready to test at your convenience now @hwoodiwiss.

martincostello:
@slang25 Do we want to make this part of 8? If so, I'll update the version number again before merging.

hwoodiwiss (changes requested):
I've reviewed down to src/JustSaying/IPublishBatchConfiguration.cs, I'll try and get back to this soon. Just a few minor changes so far.

src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs
```diff
@@ -10,38 +10,47 @@
...
 
     public SnsMessagePublisher(
-        string topicArn,
         IAmazonSimpleNotificationService client,
         IMessageSerializationRegister serializationRegister,
         ILoggerFactory loggerFactory,
         IMessageSubjectProvider messageSubjectProvider,
         Func<Exception, Message, bool> handleException = null)
-        : this(client, serializationRegister, loggerFactory, messageSubjectProvider, handleException)
+        : this(null, client, serializationRegister, loggerFactory, messageSubjectProvider, handleException)
```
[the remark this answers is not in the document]
martincostello: This is an artifact of me removing the primary constructor.

If you look at the code in main, you'll see that the `Arn` property will have a value of `null` as there's no way to pass it in unless you use the new constructor:

https://github.com/justeattakeaway/JustSaying/blob/dc7603d56be30bddf1ef7434e53966f46d91fdb4/src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs#L13-L18

This change is semantically the same, just swapping which constructor calls which.

src/JustSaying/AwsTools/MessageHandling/SnsMessagePublisher.cs
```diff
@@ -134,11 +141,122 @@ private static MessageAttributeValue BuildMessageAttributeValue(Messaging.Messag
...
+                }
+            }
+
+            if (response is { })
+            {
+                using var scope = _logger.BeginScope(new Dictionary<string, string> { ["AwsRequestId"] = response.ResponseMetadata?.RequestId });
+
+                if (response.Successful.Count > 0 && _logger.IsEnabled(LogLevel.Information))
+                {
+                    _logger.LogInformation(
```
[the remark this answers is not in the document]
martincostello: Did you have a log filter configured on JustSaying messages?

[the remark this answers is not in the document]
hwoodiwiss: No, and in the previous version, we do get the logs for individual message publishes. I'll have another look tomorrow.

src/JustSaying/JustSayingBus.cs
```diff
@@ -296,9 +342,102 @@ public InterrogationResult Interrogate()
...
 
+    /// <inheritdoc/>
     public void Dispose()
     {
         _startLock?.Dispose();
         _loggerFactory?.Dispose();
     }
+
+    /// <inheritdoc/>
+    public Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetadata metadata, CancellationToken cancellationToken)
```
hwoodiwiss, line 353: I wonder if we should make this generic to limit messaage batch publishing to one type at a time?

To me this gives the appearance of atomicity but we don't actually guarantee that. 
(Ie if I PublishAsync([MessageType1, MessageType2]...) our interface gives me the false sense that if publishing for MessageType1 fails, publishing for MessageType2 will also fail.)

src/JustSaying/MessagingBusBuilder.cs
```diff
@@ -229,6 +229,32 @@ public IMessagePublisher BuildPublisher()
...
+        var publishBatchConfiguration = MessagingConfig.BuildPublishBatchConfiguration();
+        ILoggerFactory loggerFactory = ServiceResolver.ResolveService<ILoggerFactory>();
+
+        JustSayingBus bus = CreateBus(config, loggerFactory, publishBatchConfiguration);
+        IAwsClientFactoryProxy proxy = CreateFactoryProxy();
+
+        if (PublicationsBuilder != null)
+        {
+            PublicationsBuilder.Configure(bus, proxy, loggerFactory);
+        }
```
hwoodiwiss, line 253: Can this be simplified to
```suggestion
        PublicationsBuilder?.Configure(bus, proxy, loggerFactory);
```
?

hwoodiwiss (changes requested):
Just a few more spelling bits, and a comment on the API surface

src/JustSaying/JustSayingBus.cs
```diff
@@ -121,8 +147,27 @@ public void AddMessagePublisher<T>(IMessagePublisher messagePublisher) where T :
...
+        {
+            _batchPublishersByType[typeof(T)] = batchPublisher;
+        }
+    }
+
+    public void AddMessageBatchPublisher<T>(IMessageBatchPublisher messageBatchPublisher) where T : Message
+    {
+        if (PublishBatchConfiguration.PublishFailureReAttempts == 0)
+        {
+            _log.LogWarning("You have not set a re-attempt value for publish failures. If the publish location is not available you may lose messages.");
```
hwoodiwiss, line 160:
```suggestion
            _log.LogWarning("You have not set a re-attempt value for batch publish failures. If the publish location is not available you may lose messages.");
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -296,9 +342,102 @@ public InterrogationResult Interrogate()
...
+        }
+
+        return Task.WhenAll(tasks);
+    }
+
+    private IMessageBatchPublisher GetBatchPublishersForMessageType(Type messageType)
+    {
+        if (_publishersByType.Count == 0)
+        {
+            const string errorMessage = "Error publishing message, no publishers registered. Has the bus been started?";
```
hwoodiwiss, line 371:
```suggestion
            const string errorMessage = "Error publishing message batch, no publishers registered. Has the bus been started?";
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -296,9 +342,102 @@ public InterrogationResult Interrogate()
...
+            _log.LogError(errorMessage);
+            throw new InvalidOperationException(errorMessage);
+        }
+
+        if (!_batchPublishersByType.TryGetValue(messageType, out var publisher))
+        {
+            _log.LogError("Error publishing message. No publishers registered for message type '{MessageType}'.", messageType);
+            throw new InvalidOperationException($"Error publishing message, no publishers registered for message type '{messageType}'.");
```
hwoodiwiss, line 379:
```suggestion
            _log.LogError("Error publishing message batch. No publishers registered for message type '{MessageType}'.", messageType);
            throw new InvalidOperationException($"Error publishing message batch, no publishers registered for message type '{messageType}'.");
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -296,9 +342,102 @@ public InterrogationResult Interrogate()
...
+            }
+            catch (Exception ex)
+            {
+                if (attemptCount >= PublishBatchConfiguration.PublishFailureReAttempts)
+                {
+                    _monitor.IssuePublishingMessage();
+
+                    _log.LogError(
+                        ex,
+                        "Failed to publish a message of type '{MessageType}'. Halting after attempt number {PublishAttemptCount}.",
```
hwoodiwiss, line 414:
```suggestion
                        "Failed to publish a message batch of type '{MessageType}'. Halting after attempt number {PublishAttemptCount}.",
```

src/JustSaying/JustSayingBus.cs
```diff
@@ -296,9 +342,102 @@ public InterrogationResult Interrogate()
...
+                        messageType,
+                        attemptCount);
+
+                    throw;
+                }
+
+                _log.LogWarning(
+                    ex,
+                    "Failed to publish a message of type '{MessageType}'. Retrying after attempt number {PublishAttemptCount} of {PublishFailureReattempts}.",
```
hwoodiwiss, line 423:
```suggestion
                    "Failed to publish a message batch of type '{MessageType}'. Retrying after attempt number {PublishAttemptCount} of {PublishFailureReattempts}.",
```

hwoodiwiss (commented):
I don't think these are worth blocking over, I just think it would be clearer if we disambiguated the log messages for batches.

tests/JustSaying.IntegrationTests/Fluent/Publishing/WhenAMessageIsPublishedToAQueue.cs
```diff
@@ -40,4 +42,47 @@ await WhenAsync(
...
             });
     }
-}
\ No newline at end of file
+
+    [AwsTheory]
+    [InlineData(10, 10)]
+    [InlineData(10, 100)]
+    [InlineData(5, 100)]
+    public async Task Then_Multiple_Messages_Is_Handled(int maxBatchSize, int batchSize)
```
hwoodiwiss, line 50:
```suggestion
    public async Task Then_Multiple_Messages_Are_Handled(int maxBatchSize, int batchSize)
```

tests/JustSaying.IntegrationTests/Fluent/Publishing/WhenAMessageIsPublishedToATopic.cs
```diff
@@ -40,4 +42,47 @@ await WhenAsync(
...
             });
     }
-}
\ No newline at end of file
+
+    [AwsTheory]
+    [InlineData(10, 10)]
+    [InlineData(10, 20)]
+    [InlineData(5, 10)]
+    public async Task Then_Multiple_Messages_Is_Handled(int maxBatchSize, int batchSize)
```
hwoodiwiss, line 50:
```suggestion
    public async Task Then_Multiple_Messages_Are_Handled(int maxBatchSize, int batchSize)
```

hwoodiwiss (commented):
A few more spellings

src/JustSaying/JustSayingBus.cs
hwoodiwiss: If we do want to keep the multiple message type publishing, I think we should cover that use-case in tests.

martincostello: I've added tests for this. I kept the method non-generic (`IEnumerable<Message>`) to match the non-batch publishing method so there was symmetry of usage. Otherwise if someone had this:

```csharp
foreach (var message in messages)
{
    await publisher.PublishAsync(message);
}
```

they couldn't just trivially change to:

```csharp
await publisher.PublishAsync(messages);
```

In that scenario someone would need to handle partial publishing anyway. Otherwise the caller needs to publish the specific concrete type and do their own grouping themselves, which is _probably_ going to need further refactoring in their app to use batching compared to a simple swap-over.

hwoodiwiss (approved):

hwoodiwiss:
Do we still want to bump the major version for this, or do we want to have this as the last big thing for v7, as it should be backwards-compatible.

slang25:
A big part of me wants to put this in to a V8. The compression work is almost done.

martincostello:
@slang25 What do you want to do with this PR? Is it ready to merge, or does it need to wait on something else?

martincostello:
Unless there's any objections in the next 24 hours, I'm going to merge this tomorrow and it'll just sit in main until whenever we do another release.

PR #1348, opened 2024-02-17
Support Native AOT

### LINKED ISSUE #1347: Native AOT Compatibility
**Is your feature request related to a problem? Please describe.**
I would like to be able to include JustSaying in applications that are compiled to native binaries via .NET's Native AOT feature.

**Describe the solution you'd like**
There will be a number of places that will need `DynamicallyAccessedMembersAttribute` added, but I think there are also parts, especially around serialization, where the API will need to be changed so that generic types can be fully cascaded.

**Describe alternatives you've considered**
The alternative at this point is to not use JustSaying and use the AWS libraries directly, but loose out on some of  the benefits of JustSaying.

hwoodiwiss:
This PR is a work in progress to get Native AoT working. Posting early, as I think it'll be good to get eyes on early.

AoT support as added by using a new generic serializer type and serialization factory to create it.

fixes #1347 (eventually)

Things still to do:
 - [ ] Work out properly how to get JsonSerializerOptions injected, not sure we should be adding a new dependency on `Microsoft.Extensions.Options`
 - [ ] Add better behaviour switching for .NET 8 builds, currenlty, .NET 8 builds will default to AOT compatible behaviour.
 - [ ] Cleanup changes made for local testing
 - [ ] Add general test coverage
 - [ ] Add some sort of testing to validate the case of AOT compiled builds
 - [ ] Bump Version (Major?)


hwoodiwiss:
This might be a bit cleaner to integrate if we look to add .NET 8 as a first-class TFM first

martincostello:
> * Add some sort of testing to validate the case of AOT compiled builds

We use [this test project](https://github.com/App-vNext/Polly/tree/main/test/Polly.AotTest) in Polly. _In theory_ compiling it with the assemblies trim-rooted is enough to flush out issues that the analysers can't detect.

> Bump Version (Major?)

Unless we have to break stuff, I'd class it just as a minor. I'd a minor planned for #1335 once I've validated it internally, so depending on how long both take to complete, we could have either an 8.2 then an 8.3, or ship them together as 8.2.

src/JustSaying.Extensions.DependencyInjection.Microsoft/IServiceCollectionExtensions.cs
```diff
@@ -135,7 +140,11 @@ public static IServiceCollection AddJustSaying(this IServiceCollection services,
         services.TryAddSingleton<IMessageContextAccessor>(serviceProvider => serviceProvider.GetRequiredService<MessageContextAccessor>());
         services.TryAddSingleton<IMessageContextReader>(serviceProvider => serviceProvider.GetRequiredService<MessageContextAccessor>());
 
+#if NET8_0_OR_GREATER
+        services.TryAddSingleton<IMessageSerializationFactory>(sp => new TypedSystemTextJsonSerializationFactory(sp.GetRequiredService<IOptions<JsonSerializerOptions>>().Value));
```
martincostello, line 144: I'd suggest we add a custom holder type for this, then it's more intentional for the user where the serialization options are coming from, in case they want JS to use different defaults compared to _{some other thing they have}_.

src/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
```diff
@@ -4,15 +4,19 @@
...
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Options" Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net8.0'))" />
   </ItemGroup>
   <ItemGroup>
     <Using Remove="System.Net.Http" />
   </ItemGroup>
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net8.0'))">
+    <IsAotCompatible>true</IsAotCompatible>
+  </PropertyGroup>
```
martincostello, line 21: Unless it turns out some of them are redundant, I'd add [all these](https://github.com/App-vNext/Polly/blob/a6b9b33d0ea77c2a559cd58be34e51450b9012c2/src/Polly.Core/Polly.Core.csproj#L14-L20) rather than just this one.

You could potentially put it into the `Directory.Build.props` file and have it turned on automatically for all the library projects once you've gotten them all to 0 warnings.

src/JustSaying/AwsTools/MessageHandling/PublishException.cs
```diff
@@ -17,7 +17,9 @@ public PublishException(string message, Exception inner) : base(message, inner)
     {
     }
 
+#if !NET8_0_OR_GREATER
```
martincostello, line 20: Can drop `[Serializable]` too for .NET 8.

src/JustSaying/AwsTools/QueueCreation/RedrivePolicy.cs
```diff
@@ -23,11 +26,17 @@ protected RedrivePolicy()
     {
     }
 
-    // Cannot use System.Text.Json below as no public parameterless constructor. Change for v7?
-
     public override string ToString()
+#if NET8_0_OR_GREATER
+        => System.Text.Json.JsonSerializer.Serialize(this, JustSayingSerializationContext.Default.RedrivePolicy);
```
martincostello, line 31: `this` _might_ be problematic for AoT in case it's a derived type.

src/JustSaying/Fluent/ServiceResolver/DefaultServiceResolver.cs
```diff
@@ -44,7 +45,22 @@ private object TryResolveService(Type desiredType)
...
         {
+
+#if NET8_0_OR_GREATER
+            if (RuntimeFeature.IsDynamicCodeSupported)
+            {
+                #pragma warning disable IL2026
+                #pragma warning disable IL3050
+                return new NewtonsoftSerializationFactory();
+                #pragma warning restore
+            }
```
martincostello, line 56: Slightly surprised it isn't smart enough to not warn about this.

src/JustSaying/JustSayingSerializationContext.cs
```diff
@@ -0,0 +1,12 @@
...
+using System.Text.Json.Serialization;
+using JustSaying.AwsTools.QueueCreation;
+using JustSaying.Messaging.MessageSerialization;
+
+namespace JustSaying;
+
+[JsonSerializable(typeof(SqsMessageEnvelope))]
+[JsonSerializable(typeof(IReadOnlyCollection<string>))]
+[JsonSerializable(typeof(RedrivePolicy))]
+internal sealed partial class JustSayingSerializationContext : JsonSerializerContext;
```
martincostello, line 11: It's never occured to me to use `;` than just have an empty `{...}` body instead 👍

src/JustSaying/Messaging/MessageSerialization/SystemTextJsonSerializer`1.cs
```diff
@@ -0,0 +1,119 @@
...
+        if (!jsonDocument.RootElement.TryGetProperty("MessageAttributes", out var attributesElement))
+        {
+            return new MessageAttributes();
+        }
+
+        var attributes = new Dictionary<string, MessageAttributeValue>();
+        foreach(var obj in attributesElement.EnumerateObject())
+        {
+            var dataType = obj.Value.GetProperty("Type").GetString();
+            var dataValue = obj.Value.GetProperty("Value").GetString();
```
martincostello, line 74: We should make these defensive against the possibility of the properties not being there so we don't NRE.

src/JustSaying/Messaging/MessageSerialization/SystemTextJsonSerializer`1.cs
```diff
@@ -0,0 +1,119 @@
...
+        if (options == null)
+        {
+            options = new JsonSerializerOptions
+            {
+#if NET8_0_OR_GREATER
+                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+#else
+                IgnoreNullValues = true,
+#endif
+            };
```
martincostello, line 41: Could cache this into a private static as the default, rather than create a new one per instance.

src/JustSaying/Messaging/MessageSerialization/TypedSystemTextJsonSerializationFactory.cs
```diff
@@ -0,0 +1,11 @@
+#if NET8_0_OR_GREATER
+using System.Text.Json;
+using JustSaying.Models;
+
+namespace JustSaying.Messaging.MessageSerialization;
+
+public class TypedSystemTextJsonSerializationFactory(JsonSerializerOptions options) : IMessageSerializationFactory
```
martincostello, line 7: Not the biggest fan of the `Typed` prefix.

src/JustSaying/Extensions/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        foreach (var info in options.TypeInfoResolverChain)
+        {
+            Console.WriteLine(info);
+        }
+
+        var typeInfo = options.GetTypeInfo(typeof(T));
+        if (typeInfo is not JsonTypeInfo<T> genericTypeInfo)
+        {
+            throw new JsonException($"Could not find type info for the specified type {typeof(T).Name}");
```
martincostello, line 20: This could probably be a bit more informative.

src/JustSaying.Extensions.DependencyInjection.Microsoft/JustSaying.Extensions.DependencyInjection.Microsoft.csproj
hwoodiwiss: It looks like they all become redundant if you have `<IsAotCompatible>`

[EnableSingleFileAnalyzer, EnableAotAnalyzer](https://github.com/dotnet/sdk/blob/8f8c41a31fe753bed37e09492147579280b8ee9d/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L76) and [IsTrimmable](https://github.com/dotnet/sdk/blob/8f8c41a31fe753bed37e09492147579280b8ee9d/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets#L24) are enabled if IsAotCompatible is enabled and [EnableTrimAnalyzer](https://github.com/dotnet/sdk/blob/8f8c41a31fe753bed37e09492147579280b8ee9d/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L84) is enabled if IsTrimmable is enabled.

src/JustSaying/Fluent/ServiceResolver/DefaultServiceResolver.cs
hwoodiwiss: I'll check, I think I had the check backwards when I added the `#pragma warning disable`'s

src/JustSaying/Extensions/JsonSerializerOptionsExtensions.cs
hwoodiwiss: It ends up being redundant anyway tbh, `GetTypeInfo(Type)` throws if it can't find the type info anyway.

src/JustSaying/Fluent/ServiceResolver/DefaultServiceResolver.cs
hwoodiwiss: Yeah, even with that fixed, it still warns.

src/JustSaying/AwsTools/QueueCreation/RedrivePolicy.cs
hwoodiwiss: Yeah, that's a good point, I think it'll end up serializing the base fields, but none added by the derived type.
At the moment, `RedrivePolicy` is only used internally, and I can't see a way that an API user could derive it and pass it through to any internal usage, but I could be missing it.

If that is the case, I don't think it would be a bad idea to `internal sealed` it.

hwoodiwiss:
> Unless we have to break stuff, I'd class it just as a minor. I'd a minor planned for https://github.com/justeattakeaway/JustSaying/pull/1335 once I've validated it internally, so depending on how long both take to complete, we could have either an 8.2 then an 8.3, or ship them together as 8.2.

That makes sense. It makes sense to me that we should add the `net8.0` tfm before this, in part just to reduce the overall size/complexity of this change, so maybe we could do:
8.2 - Batch Publish and net8.0 tfm
8.3 - AOT Support if it doesn't require breaking changes.

src/JustSaying/Messaging/MessageSerialization/TypedSystemTextJsonSerializationFactory.cs
hwoodiwiss: Yeah, same, I was at the "get it working at any cost" stage by then. I'll try and think of something a bit sleaker, _or_ merge this back in with the current STJSerializationFactory, and have it use the typed serializer when dynamic code isn't supported.

If we go this way, in the next breaking version, I think it'll make sense to deprecate the non-typed serializer, and to that end, maybe decorate it now.

slang25:
This is looking very nice 🙂 I had looked at this a while back in a local branch and the thing that I was concerned about was the `JsonStringEnumConverter` behaviour, which in an AOT environment will need to be added to the source generated serializer per-enum using `JsonStringEnumConverter<TEnum>`, which is fine but I think something that will be easily missed by consumers.

hwoodiwiss:
> This is looking very nice 🙂 I had looked at this a while back in a local branch and the thing that I was concerned about was the `JsonStringEnumConverter` behaviour, which in an AOT environment will need to be added to the source generated serializer per-enum using `JsonStringEnumConverter<TEnum>`, which is fine but I think something that will be easily missed by consumers.

Yeah, I'd agree, I think this is a known enough sharp edge of AOT compilation, that we could just guard the addition of the default `JsonStringEnumConverter` with `RuntimeFeature.IsDynamicCodeSupported`, I'm not sure if there's a better way that we could surface a warning around this to API consumers.

slang25:
Another thought I had while browsing this change, now might be the time to create separate interfaces and implementation for the "message envelope" and message serializer/deserializer.

src/JustSaying/Messaging/MessageSerialization/JustSayingJsonSerializerOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿#if NET8_0_OR_GREATER
+using System.Text.Json;
+
+namespace JustSaying.Messaging.MessageSerialization;
+
+public class JustSayingJsonSerializerOptions
+{
+    public JsonSerializerOptions SerializerOptions { get; } = new();
```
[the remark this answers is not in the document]
hwoodiwiss: `JsonSerializerOptions.Default` is annotated with `RequiresUnreferencedCode` and `RequiresDynamicCode`. It would also not be configurable after it's first use application-wide.

hwoodiwiss:
> Another thought I had while browsing this change, now might be the time to create separate interfaces and implementation for the "message envelope" and message serializer/deserializer.

Yeah, I had originally thought this change would require doing that anyway but found a way to avoid it. 
I think I'd prefer a v1 of AoT support to not alter the public API, at the cost of ergonomics for the AoT use-case, then we can look at building first-class support as part of the API changes that come for v8.

src/JustSaying/Fluent/ServiceResolver/DefaultServiceResolver.cs
hwoodiwiss: I've found an issue related to this: https://github.com/dotnet/runtime/issues/97273, looks like the supression in runtime checks just didn't make it into .NET 8, not sure if they plan to backport either.

It also looks like there are plans for a wider design proposal to support feature checks suppressing analyzers: https://github.com/dotnet/runtime/issues/96859

src/JustSaying/Messaging/MessageSerialization/NewtonsoftSerializer.cs
```diff
@@ -5,6 +5,10 @@
 
 namespace JustSaying.Messaging.MessageSerialization;
 
+#if NET8_0_OR_GREATER
+[RequiresUnreferencedCode(Constants.SerializationUnreferencedCodeMessage)]
+[RequiresDynamicCode(Constants.SerializationDynamicCodeMessage)]
+#endif
 public class NewtonsoftSerializer : IMessageSerializer
```
hwoodiwiss, line 12: This is an assumption at the moment. Newtonsoft.Json isn't currently annotated for AOT/Trimming compatibility.

src/JustSaying/Extensions/JsonSerializerOptionsExtensions.cs
```diff
@@ -9,18 +9,9 @@ internal static class JsonSerializerOptionsExtensions
 {
     public static JsonTypeInfo<T> GetTypeInfo<T>(this JsonSerializerOptions options)
     {
-        foreach (var info in options.TypeInfoResolverChain)
-        {
-            Console.WriteLine(info);
-        }
-
+        // This is not guarded as we want to throw if the desired type has not been configured for
```
martincostello, line 12: What does it throw though? We wouldn't want, say, a `NullReferenceException` trying to do the cast.

src/JustSaying/Messaging/MessageSerialization/SystemTextJsonSerializer.cs
```diff
@@ -12,7 +12,8 @@ namespace JustSaying.Messaging.MessageSerialization;
 [RequiresUnreferencedCode(Constants.SerializationUnreferencedCodeMessage)]
 [RequiresDynamicCode(Constants.SerializationDynamicCodeMessage)]
 #endif
-public partial class SystemTextJsonSerializer : IMessageSerializer
+[Obsolete("This API is deprecated, and will be removed in a later version. Use SystemTextJsonSerializer<T> instead.")]
```
martincostello, line 15:
```suggestion
[Obsolete("This API is deprecated, and will be removed in a future version. Use SystemTextJsonSerializer<T> instead.")]
```

src/JustSaying/Extensions/JsonSerializerOptionsExtensions.cs
hwoodiwiss: I'll update the comment, it's an argument exception if JsonSerializerOptions can't resolve a JsonTypeInfo for T. I think it'll be more understandable as an error if we just let the default error propagate, and this follows the pattern used by request delegate generators to access the typed type info (Although they know the value of T at compile time)

slang25:
Picking this up in #2183

PR #1427, opened 2024-05-23
Generate SBOM

martincostello:
Generate an SBOM from the build output.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
 
     permissions:
       attestations: write
-      contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
     permissions:
       attestations: write
-      contents: read
+      contents: write
```
martincostello, line 33: This is needed so that when a build happens for a release tag, sbom-action can attach the generated file to the release.

slang25 (approved):

PR #1557, opened 2024-11-06
Extend message attribute handling

### LINKED ISSUE #1556: System.FormatException when MessageAttributes contains type number
**Describe the bug**
When an SQS message contains MessageAttributes with Type Number, an exception of type System.FormatException is thrown when the message is picked up from the queue.

**Steps To reproduce**
Send an SQS message coining the following MessageAttributes:
"MessageAttributes" : {
    "TotalValue" : {"Type":"Number","Value":"100"}
  }

**Expected behaviour**
Message should be picked up with no exception.

**Actual behaviour**
The following exception is thrown:
System.FormatException: The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding characters, or an illegal character among the padding characters.
   at JustSaying.Messaging.MessageSerialization.NewtonsoftSerializer.GetMessageAttributes(String message) in /_/src/JustSaying/Messaging/MessageSerialization/NewtonsoftSerializer.cs:line 54
   at JustSaying.Messaging.MessageSerialization.MessageSerializationRegister.DeserializeMessage(String body) in /_/src/JustSaying/Messaging/MessageSerialization/MessageSerializationRegister.cs:line 45
   at JustSaying.AwsTools.MessageHandling.Dispatch.MessageDispatcher.DeserializeMessage(IQueueMessageContext messageContext, CancellationToken cancellationToken) in /_/src/JustSaying/AwsTools/MessageHandling/Dispatch/MessageDispatcher.cs:line 79

**System information:**
 - OS: Windows 11
 - Library Version 7.0.0 and above
 - .NET version 8

**Additional context**
Both SystemTextJsonSerializer and NewtonsoftSerializer implementation of GetMessageAttributes() handle only String data types. Everything else is considered Binary. But AWS documentation specifies that Message Attributes can have 4 data types: string, number, binary and custom which should be handled.

martincostello:
Handle message attributes of type `Number` or with a custom suffix label.

Resolves #1556.

### DIFF
--- a/src/JustSaying/Messaging/MessageSerialization/NewtonsoftSerializer.cs
+++ b/src/JustSaying/Messaging/MessageSerialization/NewtonsoftSerializer.cs
@@ -17,10 +17,10 @@ public class NewtonsoftSerializer : IMessageSerializer
     public NewtonsoftSerializer(JsonSerializerSettings settings)
     {
         settings ??= new JsonSerializerSettings
-            {
-                NullValueHandling = NullValueHandling.Ignore,
-                Converters = new JsonConverter[] { new Newtonsoft.Json.Converters.StringEnumConverter() }
-            };
+        {
+            NullValueHandling = NullValueHandling.Ignore,
+            Converters = [new Newtonsoft.Json.Converters.StringEnumConverter()]
+        };
 
         _settings = settings;
     }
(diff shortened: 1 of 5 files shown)


src/JustSaying/Messaging/MessageSerialization/NewtonsoftSerializer.cs
```diff
@@ -57,33 +57,30 @@ public MessageAttributes GetMessageAttributes(string message)
             return new MessageAttributes();
         }
 
-        var dict = new Dictionary<string, MessageAttributeValue>();
+        var attributes = new Dictionary<string, MessageAttributeValue>();
```
adammorr, line 60: Is it worth initialising the size of the dictionary with `props.Count()`?

martincostello: Is that "free", or does it have to enumerate the properties twice to count them? If the latter it might outweigh the benefit.

adammorr: I think something like this should work:

```
var attrObj = JObject.Parse(message).Value<JObject>("MessageAttributes");

if (attrObj.Count == 0)
{
   return new MessageAttributes(); 
}

var attributes = new Dictionary<string, MessageAttributeValue>(attrObj.Count);

foreach (var property in attrObj.Properties())...
```

martincostello: Ah, I didn't realise there was a property. I thought you meant LINQ count.

martincostello: I suggest doing that in the next PR which moves this code around anyway, then I don't need to worry about merge conflicts.

adammorr (approved):

src/JustSaying/Messaging/MessageSerialization/NewtonsoftSerializer.cs
martincostello: It's just an `IEnumerble<T>`, so there's no `.Count` property, only the `.Count()` method, so it would need to be enumerated twice.

PR #1525, opened 2024-10-03
Compression and JustSaying 8.0 Changes

slang25:
There is still more to go, but I'm pushing this now to get some feedback

PR #1558, opened 2024-11-06
Avoid double-parsing message payloads

martincostello:
Avoid parsing JSON documents containing the message payloads twice by adding an internal specialization for Newtonsoft.Json and System.Text.Json that deserializes both at the same time from the same parsed representation.

We _could_ decide to make this a public-facing change, but that would be a breaking change, so I figured this was the path of least resistance to make things better.

Builds on top of #1557 (which is how I noticed this in the first place).


src/JustSaying/Messaging/MessageSerialization/SystemTextJsonSerializer.cs
```diff
@@ -110,4 +85,41 @@ public string Serialize(Message message, bool serializeForSnsPublishing, string
...
+    }
+
+    private MessageAttributes GetMessageAttributes(JsonDocument jsonDocument)
+    {
+        if (!jsonDocument.RootElement.TryGetProperty("MessageAttributes", out var attributesElement))
+        {
+            return new MessageAttributes();
+        }
+
+        var attributes = new Dictionary<string, MessageAttributeValue>();
```
martincostello, line 114:
```suggestion
        var attributes = new Dictionary<string, MessageAttributeValue>(props.Count);
```

slang25 (changes requested):
This won't be a thing after #1525

PR #1559, opened 2024-11-06
Revert "Extend message attribute handling"

slang25:
Reverts justeattakeaway/JustSaying#1557

I'd like us to get #1525 over the line first, which is my own fault it's not gone in, but I'm going to revert this PR first because it's going to make more sense to re-introduce on top of #1525 rather than the other way around.


slang25:
Ignore this, I've merged with #1525

PR #1619, opened 2025-01-18
Fix StructureMap Transient Middleware Resolution

hwoodiwiss:
In StructureMap the Transient lifetime is more akin to MEDI's Scoped lifetime, where a single object instance will be created for each top level call to `Container.GetInstance()`.

Currently, the JustSayingRegistry causes a throw at startup if you don't clear and re-register these middleware as AlwaysUnique.

https://structuremap.github.io/object-lifecycle/supported-lifecycles/


martincostello (commented):
Can we write a regression test for this?

martincostello (approved):

PR #1625, opened 2025-01-23
Add Publisher TopicAddress customizer

hwoodiwiss:
This PR adds the required APIs and internals for allowing a publisher to be configured that uses message content to determine the topic ARN to publish to.


samples/src/JustSaying.Sample.Restaurant.OrderingApi/Program.cs
```diff
@@ -1,4 +1,5 @@
 using JustSaying.Messaging;
+using JustSaying.Models;
```
martincostello, line 2: Needed?

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+internal sealed class DynamicAddressMessagePublisher(
+    string topicArnTemplate,
+    Func<string, Message, string> topicAddressCustomizer,
+    Func<string, StaticAddressPublicationConfiguration> staticConfigBuilder,
+    ILoggerFactory loggerFactory) : IMessagePublisher, IMessageBatchPublisher
+{
+    private readonly string _topicArnTemplate = topicArnTemplate;
+    private readonly ConcurrentDictionary<string, IMessagePublisher> _publisherCache = new();
+    private readonly ConcurrentDictionary<string, IMessageBatchPublisher> _batchPublisherCache = new();
+    private readonly ConcurrentDictionary<string, SemaphoreSlim> _topicCreationLocks = new();
```
martincostello, line 18: Are we OK with these never being disposed?

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+        if (_publisherCache.TryGetValue(topicArn, out var publisher))
+        {
+            await publisher.PublishAsync(message, metadata, cancellationToken).ConfigureAwait(false);
+            return;
+        }
+
+        var lockObj = _topicCreationLocks.GetOrAdd(topicArn, _ => new SemaphoreSlim(1, 1));
+
+        _logger.LogDebug("Publisher for topic {TopicArn} not found, waiting on setup lock", topicArn);
+        await lockObj.WaitAsync(cancellationToken).ConfigureAwait(false);
```
martincostello, line 52: Won't this wait forever and then throw an exception when the timeout expires after the first call as we don't use `Release`?

src/JustSaying/Fluent/TopicAddressPublicationBuilder`1.cs
```diff
@@ -60,30 +66,48 @@ public TopicAddressPublicationBuilder<T> WithExceptionHandler(Func<Exception, IR
         return this;
     }
 
+    /// <summary>
+    /// Configures the address of the topic by calling this func at publish time to determine the topic ARN.
```
martincostello, line 70:
```suggestion
    /// Configures the address of the topic by calling this funcion at publish time to determine the topic ARN.
```

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
hwoodiwiss: Good catch, we have the same issue in DynamicMessagePublisher

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
hwoodiwiss: I think given Publishers tend to last the lifetime of the application, it should be okay.
To make sure unmanaged resources get cleaned up, SemaphoreSlim also (transiently) uses a SafeHandle, which will also cleanup on finalize.

samples/src/JustSaying.Sample.Restaurant.OrderingApi/Program.cs
hwoodiwiss: No

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
```diff
@@ -61,6 +61,7 @@ public async Task PublishAsync(Message message, PublishMetadata metadata, Cancel
         var config = _staticConfigBuilder(topicArn);
 
         _ = _publisherCache.TryAdd(topicArn, config.Publisher);
+        lockObj.Release(1);
```
martincostello, line 64: I think you need to do this in the batch version here and in the other file too.

hwoodiwiss: Good catch, I've added that now

hwoodiwiss:
I'll take a look at the flaky test in JustSaying.Extensions.DependencyInjection.StructureMap.Tests as a follow up

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
```diff
@@ -102,6 +102,7 @@ public async Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetada
                 var config = _staticConfigBuilder(topicArn);
 
                 var cachedPublisher = _batchPublisherCache.GetOrAdd(topicArn, config.BatchPublisher);
+                lockObj.Release(1);
```
slang25, line 105: Should this be done in a finally block, I worry we could lose count and result in a hanging app

hwoodiwiss: Yeah, this makes sense. I've added a little extension so we can do this IDisposable-y

src/JustSaying/Extensions/SemaphoreSlimExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        await semaphoreSlim.WaitAsync(cancellationToken).ConfigureAwait(false);
+        return new LockLifetime(semaphoreSlim);
+    }
+
+    private sealed class LockLifetime(SemaphoreSlim lockSemaphore) : IDisposable
+    {
+        public void Dispose()
+        {
+            lockSemaphore.Release();
```
hwoodiwiss, line 14: `.Release(1)` was redundant, we can just `.Release()`

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace JustSaying.Fluent;
+
+internal sealed class DynamicAddressMessagePublisher(
+    string topicArnTemplate,
+    Func<string, Message, string> topicAddressCustomizer,
+    Func<string, StaticAddressPublicationConfiguration> staticConfigBuilder,
+    ILoggerFactory loggerFactory) : IMessagePublisher, IMessageBatchPublisher
+{
+    private readonly string _topicArnTemplate = topicArnTemplate;
+    private readonly ConcurrentDictionary<string, IMessagePublisher> _publisherCache = new();
```
slang25, line 16: Can we make this `ConcurrentDictionary<string, Lazy<IMessagePublisher>>`, and do away with the lock collection (`_topicCreationLocks`)? It feels simpler and safer.

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
```diff
@@ -49,22 +50,24 @@ public async Task PublishAsync(Message message, PublishMetadata metadata, Cancel
         var lockObj = _topicCreationLocks.GetOrAdd(topicArn, _ => new SemaphoreSlim(1, 1));
 
         _logger.LogDebug("Publisher for topic {TopicArn} not found, waiting on setup lock", topicArn);
-        await lockObj.WaitAsync(cancellationToken).ConfigureAwait(false);
-        if (_publisherCache.TryGetValue(topicArn, out var thePublisher))
+        using(await lockObj.WaitScopedAsync(cancellationToken).ConfigureAwait(false))
```
martincostello, line 53:
```suggestion
        using (await lockObj.WaitScopedAsync(cancellationToken).ConfigureAwait(false))
```

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
```diff
@@ -81,31 +84,32 @@ public async Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetada
...
                 {
                     publisherTask.Add(publisher.PublishAsync(batch, metadata, cancellationToken));
                     continue;
                 }
 
                 var lockObj = _topicCreationLocks.GetOrAdd(topicArn, _ => new SemaphoreSlim(1, 1));
                 _logger.LogDebug("Publisher for topic {TopicArn} not found, waiting on creation lock", topicArn);
-                await lockObj.WaitAsync(cancellationToken).ConfigureAwait(false);
-                if (_batchPublisherCache.TryGetValue(topicArn, out publisher))
+                using(await lockObj.WaitScopedAsync(cancellationToken).ConfigureAwait(false))
```
martincostello, line 96:
```suggestion
                using (await lockObj.WaitScopedAsync(cancellationToken).ConfigureAwait(false))
```

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -47,24 +48,25 @@ public async Task PublishAsync(Message message, PublishMetadata metadata, Cancel
         var lockObj = _topicCreationLocks.GetOrAdd(topicName, _ => new SemaphoreSlim(1, 1));
 
         _logger.LogDebug("Publisher for topic {TopicName} not found, waiting on creation lock", topicName);
-        await lockObj.WaitAsync(cancellationToken).ConfigureAwait(false);
-        if (_publisherCache.TryGetValue(topicName, out var thePublisher))
+        using(await lockObj.WaitScopedAsync(cancellationToken).ConfigureAwait(false))
```
martincostello, line 51:
```suggestion
        using (await lockObj.WaitScopedAsync(cancellationToken).ConfigureAwait(false))
```

martincostello (approved):

src/JustSaying/Fluent/PublishConfig/DynamicAddressMessagePublisher.cs
hwoodiwiss: Yeah, I think that would express what we're trying to do in a nicer way, and make it less error prone to make changes to.

hwoodiwiss: In fact, we could go one further and have a `private readonly ConcurrentDictionary<string, Lazy<StaticAddressPublicationConfiguration>> _publisherCache = new();`, then we won't have to have a duplicate Dictionary for IMessagePublisher and IMessageBatchPublisher.

slang25: Yeah, I think that would be an improvement

hwoodiwiss: I thought it was worth looking at the changes to do this in isolation, so I've raised a PR in my fork into this branch:
https://github.com/hwoodiwiss/JustSaying/pull/1

It has a side-effect of changing the result of Interrogation, but I think that's okay? It just represents the fact that we're using the same publisher for Single vs Batch publishing now

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
```diff
@@ -83,37 +53,27 @@ public async Task PublishAsync(IEnumerable<Message> messages, PublishBatchMetada
...
         }
 
         await Task.WhenAll(publisherTask).ConfigureAwait(false);
     }
+
+    private Dictionary<string, InterrogationResult> GetInterrogationResultForTasks(Func<StaticPublicationConfiguration, InterrogationResult> interrogate) =>
+        _publisherConfigurationCache
+            .Where(w => w.Value.Value.IsCompleted)
+            .OrderBy(o => o.Key)
+            .ToDictionary(x => x.Key, x => interrogate(x.Value.Value.Result));
```
martincostello, line 69: This is a bit gnarly 😅

slang25: I like it tbh 😅

hwoodiwiss: I think I should change the where to `w.Value.Value.Status == TaskStatus .RanToCompletion`, otherwise .Result could throw if it's faulted or cancelled.

slang25 (approved):

src/JustSaying/Fluent/PublishConfig/DynamicMessagePublisher.cs
slang25: Maybe, alternatively we could just do away with the `Lazy<T>` part, `Task<T>` already has the right behaviour.

slang25: Ah, you're right, ignore me

PR #1724, opened 2025-07-08
Upgrade xunit and remove JustSaying.Tools

slang25:
Updated to xUnit v3.

Also noticed that `JustSaying.Tools` doesn't actually do anything, and hasn't for years, so have deleted it.


JustSaying.slnx
```diff
@@ -65,4 +64,4 @@
     <Project Path="tests/JustSaying.TestingFramework/JustSaying.TestingFramework.csproj" />
     <Project Path="tests/JustSaying.UnitTests/JustSaying.UnitTests.csproj" />
   </Folder>
-</Solution>
+</Solution>
```
hwoodiwiss, line 67:
```suggestion
</Solution>

```

tests/JustSaying.TestingFramework/Patiently.cs
```diff
@@ -1,8 +1,7 @@
 using System.Diagnostics;
 using NSubstitute.Exceptions;
 using Shouldly;
-using Xunit.Abstractions;
-using Xunit.Sdk;
+using Xunit; using Xunit.Sdk;
```
hwoodiwiss, line 4:
```suggestion
using Xunit;
using Xunit.Sdk;
```

PR #1783, opened 2025-11-06
Migrate documentation from GitBook to Docusaurus

slang25:
Moving documentation over to Docusaurus, and hosting on GH Pages.

Once completed I'll update the links and delete the docs branch.

I want to have more control over the docs, I have some improvements to how we render code (side project)  I want us to be able to take advantage of, and I think it's important to have the docs in the same branch for the sake of coding agents to take as context, and to update as part of making changes.


slang25:
@copilot I like this PR in general, but prefer the root of the Docusaurus site to be contained in `/docs` rather than `/website` (I know that will results in `/docs/docs/` for many of the files), can you please make this change and make the necessary changes to any config and workflows.

PR #2143, opened 2026-06-03
Feature: Optional queue existence check

ryanbell1997:
## Context
As it stands, when you configure queues with JustSaying using `ForQueueArn` the bus will start even if the queue doesn't exist. This results in heaps of errors being thrown as the application attempts to receive messages from the queue ARN. 

## Change Summary
This pull request introduces an **optional** existence check during start up when using `ForQueueArn` and `WithQueueArn`. If the queue ARN does not exist, an exception is thrown.


slang25 (dismissed):
I like this @ryanbell1997 

Just one change before we do a final review, could you rename the builder method to `WithQueueExistenceCheck`, it's wordier yes, but it's consistent and explicit.

slang25 (approved):
A nice addition

